PR #51, opened 2016-06-02
Add support for hiding the Toolbar on Android

MattDavies:
Uses the `toolBarHidden` style property.

### DIFF
--- a/android/app/src/main/java/com/reactnativenavigation/activities/BottomTabActivity.java
+++ b/android/app/src/main/java/com/reactnativenavigation/activities/BottomTabActivity.java
@@ -64,10 +64,10 @@ public class BottomTabActivity extends BaseReactActivity implements AHBottomNavi
 
     private void setupToolbar(ArrayList<Screen> screens) {
         Screen initialScreen = screens.get(0);
-        setNavigationStyle(initialScreen);
         mToolbar.setScreens(screens);
         mToolbar.setTitle(initialScreen.title);
         setSupportActionBar(mToolbar);
+        setNavigationStyle(initialScreen);
     }
 
     @Override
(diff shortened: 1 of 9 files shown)


android/app/src/main/java/com/reactnativenavigation/activities/BottomTabActivity.java
```diff
@@ -111,12 +111,12 @@ public void push(Screen screen) {
     public Screen pop(String navigatorId) {
         super.pop(navigatorId);
         Screen screen = mScreenStacks.get(mCurrentStackPosition).pop();
-        setNavigationStyle(screen);
+        setNavigationStyle(getCurrentScreen());
```
MattDavies: This was being called with the screen that was being popped rather than the screen being changed to. Not sure if that was an existing issue or if I've misinterpreted what was going on.

android/app/src/main/java/com/reactnativenavigation/modal/RnnModal.java
```diff
@@ -87,4 +96,17 @@ public void onBackPressed() {
...
     }
+
+    @Override
+    public void dismiss(){
+        super.dismiss();
+        if (mContext != null && !mContext.isFinishing()) {
+            mContext.runOnUiThread(new Runnable() {
+                @Override
+                public void run() {
+                    updateToolbar(mOriginalScreen);
```
MattDavies: Need to restore the original toolbar style before the modal was raised.

MattDavies:
Still need to test if the action bar context needs resetting after modal close.

guyca:
@MattDavies Thanks for the PR - looks promising!

MattDavies:
@guyca No problems! Moved to #52

PR #52, opened 2016-06-03
Add support for hiding the Toolbar on Android

MattDavies:
This PR will ideally also make it a lot easier to support multiple Toolbar styles within a single app.

Most of the edge cases I could find were around multiple modals, but worth checking tab/single screen activity push/pop shows/hides the Toolbar correctly as well.

Haven't touched native Android code before, so really open to feedback and suggestions.

### DIFF
--- a/android/app/src/main/java/com/reactnativenavigation/activities/BottomTabActivity.java
+++ b/android/app/src/main/java/com/reactnativenavigation/activities/BottomTabActivity.java
@@ -64,10 +64,9 @@ public class BottomTabActivity extends BaseReactActivity implements AHBottomNavi
 
     private void setupToolbar(ArrayList<Screen> screens) {
         Screen initialScreen = screens.get(0);
-        setNavigationStyle(initialScreen);
         mToolbar.setScreens(screens);
-        mToolbar.setTitle(initialScreen.title);
-        setSupportActionBar(mToolbar);
+        mToolbar.updateToolbar(initialScreen);
+        setNavigationStyle(initialScreen);
     }
 
     @Override
@@ -104,6 +103,7 @@ public class BottomTabActivity extends BaseReactActivity implements AHBottomNavi
     @Override
     public void push(Screen screen) {
         super.push(screen);
+        setNavigationStyle(screen);
         mScreenStacks.get(mCurrentStackPosition).push(screen);
     }
 
@@ -111,7 +111,7 @@ public class BottomTabActivity extends BaseReactActivity implements AHBottomNavi
     public Screen pop(String navigatorId) {
         super.pop(navigatorId);
         Screen screen = mScreenStacks.get(mCurrentStackPosition).pop();
-        setNavigationStyle(screen);
+        setNavigationStyle(getCurrentScreen());
         return screen;
     }
 
--- a/android/app/src/main/java/com/reactnativenavigation/core/objects/Screen.java
+++ b/android/app/src/main/java/com/reactnativenavigation/core/objects/Screen.java
@@ -34,6 +34,7 @@ public class Screen extends JsonObject implements Serializable {
     private static final String KEY_TOOL_BAR_STYLE = "navigatorStyle";
     private static final String KEY_STATUS_BAR_COLOR = "statusBarColor";
     private static final String KEY_TOOL_BAR_COLOR = "toolBarColor";
+    private static final String KEY_TOOL_BAR_HIDDEN = "navBarHidden";
     private static final String KEY_NAVIGATION_BAR_COLOR = "navigationBarColor";
     private static final String KEY_BUTTONS_TINT_COLOR = "buttonsTint";
     private static final String KEY_TITLE_COLOR = "titleColor";
@@ -54,6 +55,7 @@ public class Screen extends JsonObject implements Serializable {
 
     // Navigation styling
     @Nullable @ColorInt public Integer toolBarColor;
+    @Nullable public Boolean toolBarHidden;
     @Nullable @ColorInt public Integer statusBarColor;
     @Nullable @ColorInt public Integer navigationBarColor;
     @Nullable @ColorInt public Integer buttonsTintColor;
@@ -105,6 +107,7 @@ public class Screen extends JsonObject implements Serializable {
         ReadableMap style = getMap(screen, KEY_TOOL_BAR_STYLE);
         if (style != null) {
             toolBarColor = getColor(style, KEY_TOOL_BAR_COLOR);
+            toolBarHidden = getBoolean(style, KEY_TOOL_BAR_HIDDEN);
             statusBarColor = getColor(style, KEY_STATUS_BAR_COLOR);
             navigationBarColor = getColor(style, KEY_NAVIGATION_BAR_COLOR);
             buttonsTintColor = getColor(style, KEY_BUTTONS_TINT_COLOR);
--- a/android/app/src/main/java/com/reactnativenavigation/modal/RnnModal.java
+++ b/android/app/src/main/java/com/reactnativenavigation/modal/RnnModal.java
@@ -29,6 +29,7 @@ public class RnnModal extends Dialog implements DialogInterface.OnDismissListene
     private ScreenStack mScreenStack;
     private View mContentView;
     private Screen mScreen;
+    private RnnToolBar mToolBar;
 
     public RnnModal(BaseReactActivity context, Screen screen) {
         super(context, R.style.Modal);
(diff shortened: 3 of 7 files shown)


guyca:
@MattDavies Thanks a lot for the PR. I'm sure the community will appreciate it!

android/app/src/main/java/com/reactnativenavigation/activities/BottomTabActivity.java
```diff
@@ -64,10 +64,10 @@ private void setupPages(ArrayList<Screen> screens) {
 
     private void setupToolbar(ArrayList<Screen> screens) {
         Screen initialScreen = screens.get(0);
-        setNavigationStyle(initialScreen);
         mToolbar.setScreens(screens);
         mToolbar.setTitle(initialScreen.title);
         setSupportActionBar(mToolbar);
+        setNavigationStyle(initialScreen);
```
guyca: Out of curiosity, why did you move the call to `setNavigationStyle` to the bottom?

android/app/src/main/java/com/reactnativenavigation/core/objects/Screen.java
```diff
@@ -34,6 +34,7 @@
     private static final String KEY_TOOL_BAR_STYLE = "navigatorStyle";
     private static final String KEY_STATUS_BAR_COLOR = "statusBarColor";
     private static final String KEY_TOOL_BAR_COLOR = "toolBarColor";
+    private static final String KEY_TOOL_BAR_HIDDEN = "toolBarHidden";
```
guyca: Lets change `toolBarHidden` to `navBarHidden` which is the property we [already use](https://github.com/wix/react-native-navigation#style-object-format) in iOS

MattDavies: Sure, easy. There are a few other "toolBar" props we can update in a separate PR too.

android/app/src/main/java/com/reactnativenavigation/activities/BottomTabActivity.java
```diff
@@ -64,10 +64,10 @@ private void setupPages(ArrayList<Screen> screens) {
...
     }
```
MattDavies: This had to be at the bottom to ensure the action bar was set before applying styles (else you can't show/hide).

MattDavies:
@guyca Changed to `navBarHidden`

android/app/src/main/java/com/reactnativenavigation/modal/RnnModal.java
```diff
@@ -87,4 +105,17 @@ public void onBackPressed() {
...
     }
+
+    @Override
+    public void dismiss(){
+        super.dismiss();
+        if (mContext != null && !mContext.isFinishing()) {
+            mContext.runOnUiThread(new Runnable() {
+                @Override
+                public void run() {
+                    updateToolbar(mOriginalScreen, mOriginalToolbar);
```
guyca: If the Modal is dismissed, the previous toolbar already has the correct style and buttons. Do we call `updateToolbar` only for `mContext.setSupportActionBar(toolbar);` ?

MattDavies: Ah. I was finding the visibility of the previous toolbar was being altered and I needed to reset it, but that might have been before I was setting the action toolbar context correctly on the modal. Let me have a play and see if I can clean this up.

guyca: Great. What do you think about moving `updateToolbar` to `RnnToolbar` class? I didn't have time to check this, but I think it will be used from other places as well. 

Maybe something like this:

```
@UiThread
    public void updateToolbar(Screen screen) {
        setTitle(screen.title);
        setSupportActionBar();
        setStyle(screen);
        setupToolbarButtonsAsync(screen);
    }

    public void setSupportActionBar() {
        ((AppCompatActivity) getContext()).setSupportActionBar(this);
    }
```

Then, in `RnnModal` we would call `mToolBar.updateToolbar(mScreen);`
This also eliminates the need to keep a hard ref to `mOriginalToolbar`, a View that belongs to the previous Activity/Modal, which is a bad practice.

If we expose `setSupportActionBar` we also don't need to keep a ref to `mOriginalScreen` - seems like it simplifies the class a bit.

android/app/src/main/java/com/reactnativenavigation/activities/BottomTabActivity.java
```diff
@@ -104,14 +104,15 @@ public boolean onCreateOptionsMenu(Menu menu) {
...
+        setNavigationStyle(screen);
         mScreenStacks.get(mCurrentStackPosition).push(screen);
     }
 
     @Override
     public Screen pop(String navigatorId) {
         super.pop(navigatorId);
         Screen screen = mScreenStacks.get(mCurrentStackPosition).pop();
-        setNavigationStyle(screen);
+        setNavigationStyle(getCurrentScreen());
```
MattDavies: When popping it was applying the style of the screen that had just been popped rather than the new screen.

android/app/src/main/java/com/reactnativenavigation/modal/RnnModal.java
```diff
@@ -67,10 +75,17 @@ public void onDisplayed() {
...
 
     public Screen pop() {
-        return mScreenStack.pop();
+        if (mScreenStack.getStackSize() > 1) {
+            return mScreenStack.pop();
+        } else {
+            ModalController.getInstance().remove();
+            super.onBackPressed();
+            return null;
+        }
```
MattDavies: When pop was called on the only remaining screen on a modal it didn't appear to be clearing the modal out properly.

MattDavies:
Hey @guyca I managed to simplify the modal logic a fair bit and fixed another couple of issues which I was hitting. I've got a couple of changes in https://github.com/MattDavies/react-native-navigation/tree/test-toolbar-hide which you might find useful for testing as well.

android/app/src/main/java/com/reactnativenavigation/modal/RnnModal.java
```diff
@@ -66,11 +73,18 @@ public void onDisplayed() {
     }
 
     public void push(Screen screen) {
+        updateToolbar(screen, mToolBar);
```
MattDavies: It didn't look as though we were previously updating the toolbar style for a pushed screen within a modal.

guyca: Good catch! We also miss this when pushing a screen in Activities (`BottomTabActivity` and `SingleScreenActivity`). That's why I suggested we declare `updateToolbar` in `RnnToolbar`.

guyca:
@MattDavies I went over #60ab611a31a5b5681e236b4f7a8ca3f0a51f6622 and seems like you're still using `toolBarHidden` instead of `navBarHidden`. Or did I miss something?

If you wish I can move `updateToolbar` to `RnnToolBar` for you. Ping me once you update the PR and I'll merge it.

PS
I'll be out of the office all of next week, So if we don't merge today I'll ask one of the other guys here to pick this up.

MattDavies:
Updated to move `updateToolbar` to `RnnToolbar`. It definitely reads a lot clearer now - great suggestion!

I think I lost the `navBarHidden` change somewhere along the lines...added it back.

android/app/src/main/java/com/reactnativenavigation/activities/BottomTabActivity.java
```diff
@@ -64,10 +64,9 @@ private void setupPages(ArrayList<Screen> screens) {
 
     private void setupToolbar(ArrayList<Screen> screens) {
         Screen initialScreen = screens.get(0);
-        setNavigationStyle(initialScreen);
         mToolbar.setScreens(screens);
-        mToolbar.setTitle(initialScreen.title);
-        setSupportActionBar(mToolbar);
+        mToolbar.updateToolbar(initialScreen);
+        setNavigationStyle(initialScreen);
```
MattDavies: This still sets window styles so I left it there even though there's now a slight duplication. Lets look at that in another PR?

guyca: Sure, we'll address that in a separate PR. I'll merge soon. Thanks for the contribution Matt 👍

MattDavies: Welcome! Great library, it's making a huge difference to some of my apps compared to the official nav. I'll keep jumping in where I can to fill gaps.

PR #69, opened 2016-06-19
Added support for passing client prop to Provider wrapper component

heysailor:
Passing a client prop to the <Provider /> allows use of libraries such as Apollo Client, which depend on access to the root element.

More information: http://docs.apollostack.com/apollo-client/react.html

For example, the Apollo Client library has an enhanced ApolloProvider component to allow access to its client in the react tree. It depends on a client prop. This PR allows use as below:

app.js:
```
/* ...create redux store, etc  */

import ApolloClient from 'apollo-client';
import { ApolloProvider } from 'react-apollo';
import { registerScreens } from './screens'; // react-native-navigation

const client = new ApolloClient();

registerScreens(store, ApolloProvider, client);
```

registerScreens.js:
```
/* ... import screens ... */

export function registerScreens(store, Provider, client) {
  Navigation.registerComponent('cz.ScreenWithApolloDependencies', () => ScreenWithApolloDependencies, store, Provider, client);
}
```

Another viable alternative would a more generic way of passing props to the `<Provider />` component. This could be as simple as taking a 4th `providerProps` argument where I've put '`client`', which would  then be spread on to `<Provider />`.

### DIFF
--- a/src/Navigation.js
+++ b/src/Navigation.js
@@ -10,9 +10,9 @@ function registerScreen(screenID, generator) {
   AppRegistry.registerComponent(screenID, generator);
 }
 
-function registerComponent(screenID, generator, store = undefined, Provider = undefined) {
+function registerComponent(screenID, generator, store = undefined, Provider = undefined, client= undefined) {
   if (store && Provider) {
-    return _registerComponentRedux(screenID, generator, store, Provider);
+    return _registerComponentRedux(screenID, generator, store, Provider, client);
   } else {
     return _registerComponentNoRedux(screenID, generator);
   }
@@ -35,7 +35,7 @@ function _registerComponentNoRedux(screenID, generator) {
   return generatorWrapper;
 }
 
-function _registerComponentRedux(screenID, generator, store, Provider) {
+function _registerComponentRedux(screenID, generator, store, Provider, client) {
   const generatorWrapper = function() {
     const InternalComponent = generator();
     return class extends Screen {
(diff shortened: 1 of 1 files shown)


brianjd:
Thanks for adding this @heysailor!  When can we expect this to be merged?

heysailor:
I think we're waiting for some major library structure decisions to be made, from other emails I've received. If you look at the merge changes, it's trivial to implement if you want it in the meantime.






On Tue, Aug 23, 2016 at 5:34 AM +1000, "b_d" <notifications@github.com> wrote:












Thanks for adding this @heysailor!  When can we expect this to be merged?

smooJitter:
It is easy to customize your own registerComponent( ) function.  Another possibility is to use registerComponent  and yet pass in a wrapped component.  E.g,  Navigation.registerScreen('screenId', () => <ApolloProvider ...> screen </ApolloProvider>).   I haven't tried it but essentially, that's the only difference.

heysailor:
That's true, and I do use a custom screen registration function (so I can centralise settings and a few other things).
But adding a client in the way proposed by this PR means you don't have problems accessing the redux state.






On Tue, Aug 23, 2016 at 9:02 PM +1000, "smooJitter" <notifications@github.com> wrote:












It is easy to customize your own registerComponent( ) function.  Another possibility is to use registerComponent  and yet pass in a wrapped component.  E.g,  Navigation.registerScreen('screenId', () =>  screen ).   I haven't tried it but essentially, that's the only difference.

DanielZlotin:
Should be handled by #233

guns2410:
Please merge this...

jay1337:
.. and add this to the online documentation !

PR #82, opened 2016-06-24
Android drawer implementation

jskuby:
This is a native drawer implementation that uses the animated arrow menu icon by default.  Currently only a left menu is implemented for SingleScreenActivity and BottomTabActivity.  I've been using this in a client application with SingleScreenActivity for the past few weeks and it's been working well.

### DIFF
--- a/android/app/src/main/java/com/reactnativenavigation/activities/BaseReactActivity.java
+++ b/android/app/src/main/java/com/reactnativenavigation/activities/BaseReactActivity.java
@@ -7,6 +7,7 @@ import android.os.Bundle;
 import android.os.Handler;
 import android.provider.Settings;
 import android.support.annotation.CallSuper;
+import android.support.v4.widget.DrawerLayout;
 import android.support.v7.app.ActionBarDrawerToggle;
 import android.support.v7.app.AppCompatActivity;
 import android.util.Log;
@@ -14,6 +15,7 @@ import android.view.KeyEvent;
 import android.view.Menu;
 import android.view.MenuItem;
 import android.widget.EditText;
+import android.widget.FrameLayout;
 import android.widget.Toast;
 
 import com.facebook.common.logging.FLog;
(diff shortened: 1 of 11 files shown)


android/app/src/main/java/com/reactnativenavigation/activities/BottomTabActivity.java
```diff
@@ -65,6 +72,20 @@ public void run() {
         });
     }
 
+    protected void setupDrawer(Drawer drawer, Screen screen) {
```
guyca: can this function be defined in `BaseReactActivity` ?

jskuby: I don't see why not,  I was just keeping in line with the setup functions being defined in the derived classes.

guyca: Let's move it to the base class then I'll merge. Looks great! Thanks a lot.

jskuby: No problem at all, it was a blast to work with!  I'll get on that update later today.

android/app/src/main/java/com/reactnativenavigation/activities/BaseReactActivity.java
```diff
@@ -355,6 +389,20 @@ public void invokeDefaultOnBackPressed() {
         super.onBackPressed();
     }
 
+    protected void setupDrawer(Screen screen, Drawer drawer, int drawerFrameId, int drawerLayoutId) {
```
jskuby: `setupDrawer` is taking in the layout IDs here since I'm not sure there's a way to get those from the base activity.  I wasn't able to when initially refactoring.

jskuby: Nevermind me, it makes sense.  They're different layouts, so they will generate different layout IDs.

guyca: We also don't have to pass the id's as params to `setupDrawer` since in both layout files we use the same ids for the views. `findViewById` will search for these views in the layout which was passed to `setContentView`. No biggie, I'll merge and make that minor change later.
@jskuby Thanks a lot for your work!

PR #83, opened 2016-06-24
Android style fixes

jskuby:
This update ensures navigator styles are applied on screen changes.  It also merges the base styles with screen-specific styles as to allow default values that can be overridden.

### DIFF
--- a/src/platformSpecific.android.js
+++ b/src/platformSpecific.android.js
@@ -73,7 +73,7 @@ function navigatorPop(navigator, params) {
 function navigatorPopToRoot(navigator, params) {
   RctActivity.navigatorPopToRoot({
     navigatorID: navigator.navigatorID,
-    animated: !(params.animated !== false)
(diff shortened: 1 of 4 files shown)


src/platformSpecific.android.js
```diff
@@ -73,7 +73,7 @@ function navigatorPop(navigator, params) {
 function navigatorPopToRoot(navigator, params) {
   RctActivity.navigatorPopToRoot({
     navigatorID: navigator.navigatorID,
-    animated: !(params.animated !== false)
+    animated: !(params.animated === false)
```
guyca: Can this conditional be changed to `params.animated === true`?

jskuby:
That depends on what the default behavior should be. If we're defaulting to false then the === true makes sense, since any non-true value would return false. With my change the value defaults to true for any non-false (but still falsey) value.

PR #84, opened 2016-06-24
Set tab title on nav title change

jskuby:
Small fix to set the tab title when the screen title is updated in a BottomTabActivity.


guyca:
Hey @jskuby - I'm not sure this is the intended behaviour of `BottomNavigation` on Android. Wasn't able to find anything about this in the design guidelines
https://material.google.com/components/bottom-navigation.html#bottom-navigation-specs

jskuby:
When dynamically setting the title in the example app, the iOS version updates the tab title. This update is to make Android behave in a similar manner.

guyca:
@jskuby This sounds actually more like a bug then a desired behaviour and was reported in #113, Sorry about the confusion.

PR #85, opened 2016-06-24
Example updates

jskuby:
Upgrade both example projects to RN 0.27.2 and consolidate example app screens.  A `react-native upgrade` was also performed on the iOS example project.  Both example projects were running after the upgrade on my simulators/emulators.


DanielZlotin:
An upgrade to RN31 is scheduled for 2.0.0

PR #93, opened 2016-06-25
[Android] Fix starting menu icon color

jskuby:
The navBarButtonColor was only being applied to screens after the first.  The fix ensures the color is applied to the starting screen as well.

### DIFF
--- a/android/app/src/main/java/com/reactnativenavigation/views/RnnToolBar.java
+++ b/android/app/src/main/java/com/reactnativenavigation/views/RnnToolBar.java
@@ -56,6 +56,7 @@ public class RnnToolBar extends Toolbar {
     private DrawerLayout mDrawerLayout;
     private ActionBarDrawerToggle mDrawerToggle;
     private ArrayList<View> mMenuItems;
+    private int mButtonColor;
 
     public RnnToolBar(Context context) {
         super(context);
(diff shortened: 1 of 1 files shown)


android/app/src/main/java/com/reactnativenavigation/views/RnnToolBar.java
```diff
@@ -56,6 +56,7 @@
     private DrawerLayout mDrawerLayout;
     private ActionBarDrawerToggle mDrawerToggle;
     private ArrayList<View> mMenuItems;
+    private int mButtonColor;
```
guyca: @jskuby Lets not keep this as a member since we use it only in a single function

jskuby: @guyca When calling `setNavUpButton` without a screen it's implied we're navigating back.  Having no screen object also uses the default color when styling the icon, which may not have been the color of the previous screen's menu icon.  This may just be more of an argument for a previous comment of mine mentioning splitting up `setNavUpButton` into separate style and state methods.

guyca: hmm.. the next developer who would step into the code won't know that passing `null` to `setNavUpButton` implies it's called after navigating back.
Maybe if the screen is `null` - don't style the up button. Instead, we can style it in `setStyle` which is called after `StyleHelper.updateStyles` is called.
wdyt?

jskuby: I like it.  That also means we don't need to pass screen into `setNavUpButton`, it just needs to be able to set the correct state.

jskuby:
Pretty sure this will be resolved with the updates from PR #97.

Issue #94, opened 2016-06-25
[Android] Module reference lost after rotation


jskuby:
I have several native modules included with my application, but after I switched over to this navigation package I get an error in my localization module after rotation.  The red screen error is `undefined is not an object (evaluating 'RNI18n.locale')`.

My `MainActivity.java` is pretty bare, only having `getPackages` implemented as follows:

```
    @Override
    public List<ReactPackage> getPackages() {
        return Arrays.<ReactPackage>asList(
            new MainReactPackage(),
            new RnnPackage(),
            new ReactNativeI18n(),
            ...
        );
    }
```

I can't tell if other modules are ok after rotation since I can't get past this error.  I do see some of the screen rotate before it blows up, but I'm not sure if that helps.

jskuby:
This is fixed by the change in PR #97.

PR #97, opened 2016-06-25
[Android] Fix crash on device rotate

jskuby:
This PR fixes crashes when rotating, at least on the project I'm working on.  This is related to and fixes issue #94.


jskuby:
For more info, I found the fix here: http://stackoverflow.com/a/27937906

jskuby:
This isn't fully fixed since my rotate events aren't firing, but it doesn't crash either.  I'll work with it a bit more to see if I can get it working.

jskuby:
Last two commits fix the menu icon state when `onConfigurationChanged` fires.

android/app/src/main/java/com/reactnativenavigation/views/RnnToolBar.java
```diff
@@ -214,17 +214,16 @@ private void hideToolbar() {
     }
 
     public void setNavUpButton() {
-        setNavUpButton(null);
+        setNavUpButton(null, false);
     }
 
     @SuppressWarnings({"ConstantConditions"})
-    public void setNavUpButton(Screen screen) {
+    public void setNavUpButton(Screen screen, boolean isBack) {
```
guyca: @jskuby Why is this change necessary? Sending boolean params to functions that change what the behaviour of the function is quit an anti pattern.

jskuby: This was a quick update so it can be called from `onConfigurationChanged` since my menu icon was losing its state on rotation.

How should we go about handling this?  Separate functions for styling and setting the state?

guyca: @jskuby Can we move the `setNavUpButton` call to `StyleHelper.updateStyles`? If so, seems like we determine if the back button needs to be displayed or not buy calling `getScreenStackSize() ` which is already public. Am I right?

In `RnnToolbar` we can grab a the `BaseReactActivity` instance by calling `getContext()`.
wdyt?

jskuby: I took a look and that all seems good @guyca.  I probably won't have time to work on this until tomorrow, but I'll see what I can do to move this along.

jskuby:
@guyca After moving the `setNavUpButton` call into `StyleHelper.updateStyles` there is no need to call the super navigation methods.  Right now I've removed them and the annotation for the super call, but left the implementations.  How should those be handled?  We can make them abstract, but RootActivity will then need noop implementations.

guyca:
Thanks @jskuby! I'll check it out, hopefully later today

ronaldheft:
Any ETA on the merge?

ronaldheft:
@jskuby What issues were you having with rotation events not firing? I've pulled this change into my fork, and I'm trying to use react-native-orientation to lock my app in portrait. I'm not seeing react-native-orientation pick up the rotation events.

ronaldheft:
^ For anyone experiencing the issues I described above, I figured it out. ^

The issue is the hierarchy of class extensions. React Native's `MainActivity` extends `RootActivity` which extends `BaseReactActivity`. When React Native Navigation starts up, it change's the app's Activity from `MainActivity` to either `BottomTabActivity` or `SingleScreenActivity`. The issue is both of those classes extend `BaseReactActivity` directly, so any listeners you place in `RootActivity` will not be firing since that activity is no longer running.

I ended up adding the listener needed for the orientation module to my fork directly in `BaseReactActivity`. I also added `android:screenOrientation="portrait"` to React Native Navigation's AndroidManifest.xml for the various activities.

Thanks @jskuby for the original code that got me started on figuring out what in the hell is going on.

drorbiran:
Hi,
Unfortunately, we can't merge your PR.
We will address this issue in the near future. You can follow #258

dg92:
what is the solution ?

guyca:
@dg92 For now, I suggest you lock you app to portrait mode by adding `portraitOnlyMode=true` property when calling `startApp`

```js
    screen: {
      screen: 'core.LoginScreen',
      title: 'Login'
    },
    portraitOnlyMode: true
  })
```

dg92:
@guyca no its not working for me.
`case actionTypes.app.screens.login:
      screen: {
        screen: 'auth.Login',
        navigatorStyle: navBarStyle,
        passProps: action.payload.props
      },
      portraitOnlyMode: true
    });
    return {...state, currentScreen: 'login'};
`

guyca:
@dg92 `portraitOnlyMode` works only on Android for now and is available in the experimental branch. If you're using the stable 1.30 version you'll need to upgrade

dg92:
@guyca  using "react-native-navigation": "^2.0.0-experimental.105", and checking on android 4.4.4

guyca:
@dg92 `portraitOnlyMode` was added in experimental 125 if i'm not mistaken

vivekparekh8:
@ronaldheft My scenario : whole app is portrait using rnn and one of the screens is landscape. so I am using react-native-orientation. My code for landscape screen
```
componentWillMount(){
Orientation.lockToLandscape()
}
```
For me, whole app restarts on doing this

PR #121, opened 2016-07-11
Unmount ReactRootView when the JS bundle is being reloaded.

beast:
Garbage collect all JS memory.

### DIFF
--- a/android/app/src/main/java/com/reactnativenavigation/views/RctView.java
+++ b/android/app/src/main/java/com/reactnativenavigation/views/RctView.java
@@ -71,11 +71,10 @@ public class RctView extends FrameLayout {
     }
 
     /**
-     * Must be called before view is removed from screen inorder to ensure onDetachedFromScreen is properly
-     * executed and componentWillUnmount is called
+     * Must be called before view is removed from screen inorder to ensure ReactRootView is unmounted.
      */
     public void onRemoveFromScreen() {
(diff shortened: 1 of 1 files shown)


beast:
Never mind. This wouldn't work. I just realized pop is also using this. Will resubmit

beast:
Wrong branch. Re-Submitting PR

android/app/src/main/java/com/reactnativenavigation/views/RctView.java
```diff
@@ -71,11 +71,10 @@ public void onTemporallyRemovedFromScreen() {
     }
 
     /**
-     * Must be called before view is removed from screen inorder to ensure onDetachedFromScreen is properly
-     * executed and componentWillUnmount is called
+     * Must be called before view is removed from screen inorder to ensure ReactRootView is unmounted.
      */
     public void onRemoveFromScreen() {
-        ReflectionUtils.setField(mReactRootView, "mAttachScheduled", false);
+        ReflectionUtils.invoke(mReactRootView, "unmountReactApplication");
```
guyca: unmountReactApplication is public, no need to invoke it using reflection.
From what I can tell this method was introduced in RN 0.26. This will break the library for users that use older versions.
I'll look into this now

guyca:
@beast closing the issue. No need for a PR, we'll fix it in on our end.

Issue #100, opened 2016-06-28
Is it possible to pass a function to props when using navigator.push?


zocario:
I am playing with your great library with my iOS app and I have a weird issue.
I like the idea of passing functions through props in React components, as for InputText's onChangeText property.

```
<InputText onChangeText={ (text) => this.setState({ myText: text }) } />
```

So I tried to do the same between two views, the first is a ListView form containing a long Text (not editable). When I tap on the long text I push a new ListView using your navigator.push method to present a long InputText. I am passing in passProps a method like this:

```
this.props.navigator.push({
  ...
  passProps: {
    onChangeText: (text) => this.setState({ myLongText: text })
  }
  ...
})
```

..so I can update my state and retrieve the new entered text. But the only thing I receive in the view is an empty object. So is it possible to pass a function through passProps? If not is there an other way to retrieve the information?

<img width="905" alt="capture d ecran 2016-06-28 a 12 39 02" src="https://cloud.githubusercontent.com/assets/800617/16412595/55f21910-3d2d-11e6-8215-2308353d6a25.png">

webzepter:
No you can't pass a function, because you can serialise it for passing through native bridge.
You can without extending:
1) Pass function code as string and `eval` it 😬
2) Pass e.g. boolean prop and new screen will run necessary function
3) Extend this library and create module with function registry (in `push` you need register props-functions and pass them in new screen during initialisation. This approach allows you have closures (like in your example with `this.setState`)

zocario:
Thanks for the answer, it could be really nice to have this possibility in the future, as it is really a common pattern in React. For now I think I will try something like this as a temporary workaround: 
https://www.npmjs.com/package/react-native-simple-events

In my opinion options 1) and 2) would not work in this case as eval would be done in an other view, and passing a boolean doesn't allow me to have the returned value of the new screen isn't it? Am I missing something?

webzepter:
Yes, you are right

iosipeld:
As an alternative to @zocario solution, redux/flux action could be used to modify the state of parent component (parent component will listen to `componentWillReceiveProps` and then set a new state via `setState`).

sshymko:
An alternative to passing a callback function is to pass an `EventEmitter` object, for example:
```js
import EventEmitter from 'EventEmitter';

let events = new EventEmitter();
events.addListener('changeText', (text) => this.setState({myText: text}));
this.props.navigator.push({
  // ...
  passProps: {
    events: events,
  }
})
```
[...]

Issue #10, opened 2016-04-02
How to get Value from a ModalScreen without put it into redux state tree


iwater:
I want pick date from DatePicker using ModalScreen, the passProps only can transport serializable values, can't use a callback function, because it's a temp value, it's there any way to get value back from ModalScreen without put it into redux state tree when using react-native-navigation

talkol:
See this related issue:
https://github.com/wix/react-native-controllers/issues/13

Please go over the solutions proposed there and participate in the discussion of how this issue should be addressed

PR #136, opened 2016-07-19
Passing callbacks to modal/pushed views

dingbat:
Addresses: https://github.com/wix/react-native-controllers/issues/13 (sort of, since it's in a different repo), https://github.com/wix/react-native-navigation/issues/10, https://github.com/wix/react-native-navigation/issues/100

This is a simple & functioning prototype for passing callbacks to modal views for both iOS and Android. I was inspired by option 3 in https://github.com/wix/react-native-controllers/issues/13.

The callback is passed in as a parameter to `showModal`:

```js
this.props.navigator.showModal({
  screen: 'example.CameraView',
  callback: (imagePath) => this.setState({ imagePath })
})
```

And the modal view simply calls:

```js
this.props.navigator.callback(value)
```

This model only supports one callback per view, but IMO that's fine because you can differentiate events using params.

It's a very small change and avoids any bridging/platform-specific implementations by putting the callback function in a global registry when `showModal`/`push` are called and associating it with the new screen's ID. `navigator.callback(...)` then calls the function from the registry using the current screen's ID.

Thoughts?

### DIFF
--- a/src/Navigation.js
+++ b/src/Navigation.js
@@ -5,6 +5,8 @@ import Screen from './Screen';
 
 const registeredScreens = {};
 
+const registeredCallbacks = {};
+
 function registerScreen(screenID, generator) {
   registeredScreens[screenID] = generator;
   AppRegistry.registerComponent(screenID, generator);
@@ -63,6 +65,14 @@ function getRegisteredScreen(screenID) {
   return generator();
 }
 
+function registerCallback(screenInstanceID, callback) {
+  registeredCallbacks[screenInstanceID] = callback;
+}
+
+function getRegisteredCallback(screenInstanceID) {
+  return registeredCallbacks[screenInstanceID];
+}
+
 function showModal(params = {}) {
   return platformSpecific.showModal(params);
 }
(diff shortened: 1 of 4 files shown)


lprhodes:
Thanks for this, looks like you made the amends just as I was needing them! Bare in mind that Navigation.showModal can be used also but your amends don't support that.

lprhodes:
...in fact it breaks it.

dingbat:
@lprhodes thanks for bringing that up! Latest commit should fix that, and also makes the callback registration cleaner I think.

DanielZlotin:
Hey @dingbat.
Thanks for your contribution.
We are currently working hard on a `v2` version of this package which addresses a lot of issues, one of them is addressing the passProps issue in a more general way: similar to your solution, we will keep *all* passed props in JS side without serialising them over the bridge. This will obviously solve your issue.
We will update you once the new version is ready (there should not be breaking changes in the API, only in the implementation)

Again, thank you very much for your contributions!
We really appreciate your effort.

dingbat:
@DanielZlotin That's perfect, thanks for the update!

DanielZlotin:
Fixed in `2.0.0-experimental.48`.
keep in mind this is **experimental** release and contrary to previous statements, the API will likely break until the official `2.0.0` launch. This feature will stay there though :)

The reasons for breaking are described elsewhere: @drorbiran please link to the article.

PR #140, opened 2016-07-19
Support showing a SnackBar on Android

artald:


drorbiran:
A re-implementation of a SnackBar for Android will be available in V2.0.0

guyca:
Snackbar is now available on Android.

**Usage:**
```js
  this.props.navigator.showSnackbar({
      text: 'Counter: ' + this.props.counter.count,
      actionText: 'Undo',
      actionColor: '#ff0000',
      actionId: 'undo', 
      duration: 'indefinite' // optional. available values: 'short', 'long' and 'indefinite'. Default is 'short'
    });
  }
```

The event is handled in `onNavigatorEvent` as you would handle any other button event.

Issue #49, opened 2016-06-01
Set to default type=MMdrawer animationType=slide


folivi:
Hi all,
anyone has figured out how to get rid of this message when opening sidemenu?
Cheers

gran33:
Hi @folivi (and @madvas),
This warning msg is there just to let u know, that the default values has been set.
If u would like to disappear the waning, simple set values.

To keep the default values, set it to:
```
type="MMDrawer"
animationType="slide"
```

folivi:
thanks for the reply, but in case of a TabBasedApp where to set those values?
thx again

gran33:
@folivi I pushed a change, so u can now remove the warn and set (for the default drawer animation type):
inside your 
```
Navigation.startTabBasedApp({
...
  drawer: {
  type="MMDrawer"
  animationType="slide"
  ...
  }
})
```
U can have many animation types (took from [react-native-controllers](https://github.com/wix/react-native-controllers#drawercontrollerios)):
> if type=MMDrawer animationTypes=door/parallax/slide/slide-and-scale. If type=TheSideBar animationType=airbnb/facebook/luvocracy/wunder-list (Default is type=MMDrawer animationType=slide)

folivi:
Great! 
Thanks a lot.
```

  drawer: {
      type: 'MMDrawer',
      animationType: 'slide',
      left: {
          screen: 'gsb.MenuView'
      }
  }

```

gran33:
👍

SvetoslavKuzmanov:
Hi all,

I'm still getting this warning and I cannot change the neither animation type nor the drawer type. Here is how I'm using it: 
```
Navigation.startSingleScreenApp({
    screen: {
      screen: 'example.WelcomeScreen',
      title: 'Welcome',
      navigatorStyle: {},
      navigatorButtons: {
        leftButtons: [{
          icon: iconsMap['ios-menu'],
          id: 'menu'
        }]
      }
    },
    drawer: {
      left: {
        screen: 'example.SideMenu'
      },
      type: 'MMDrawer',
      animationType: 'slide',
      disableOpenGesture: true
    },
  });
```

Any ideas anybody?

jackread:
Doesn't look like this was implemented for startSingleScreenApp

Still getting a warning every time.

gazpachu:
I'm also still getting that warning with startSingleScreenApp all the time.

joshwhatk:
[...]

PR #154, opened 2016-07-26
type and animationType on startSingleScreenApp

jackread:
implements same features as startTabBasedApp


jackread:
should reference #49

DanielZlotin:
Close as duplicate of #66

PR #175, opened 2016-08-05
Ability to set the id of of the container when using startTabBasedApp

lprhodes:


DanielZlotin:
I'm not sure why would you do that? Please provide a use case.

lprhodes:
I use it for the following purpose:

```
  const tabBarController = Controllers.TabBarControllerIOS('main_tabs')

  tabBarController.setBadge({
    tabIndex: 3,
    badge: badgeCount
  })
```

DanielZlotin:
I'm still not quite sure why would you use it like that? This is an internal API, you should be using the `Navigation/Navigator` API:

```
this.props.navigator.setTabBadge({
  tabIndex: 0, // (optional) if missing, the badge will be added to this screen's tab
  badge: 17 // badge value, null to remove badge
});
```

And in case you have more than 1 `BottomTabs` controller, you should dispatch that action with some other way, for example, through a `redux` store, or EventBus, or Registry etc.

lprhodes:
- We start off with a single view application.
- After the user has entered their details when then use an auth class to authenticate the user.
- Once authenticated we start a Tab Bar application and fade it in, at the same time we call our badgeNotification class that observes a Firebase path for a change in value.
- We then update the badge when the value changes as per the code above.

DanielZlotin:
@lprhodes I see. 
so as I said, instead of using a private API inside controllers (which will change anyway), and furthermore to keep a parity between the platforms: pass the badge value through some registry / redux / emitter then update it from the props.

Hope this is sufficient. If not, let's try to think of a way using the current Navigation API, and not through controllers. Otherwise, we have a feature planned that should address accessing a specific screen.

PR #262, opened 2016-08-29
individually import for lodash components

alexesDev:
This saves ~200kb uncomporessed code on my project.


DanielZlotin:
Appreciate the PR, but I don't like the individual imports, that is - until we will have full test coverage. Too easy missing an import. 

Also, we consider using a JS dead code removal (pseudo compiler) on our production app.

PR #270, opened 2016-08-31
Slightly more backwards compatibility in android leftButtons

yedidyak:


DanielZlotin:
It's temp anyway but OK

PR #288, opened 2016-09-07
Remove react hacks

guyca:
This commit removes necessary hacks which were introduced due to
limitations of previous React Native apis.

Mount/unmount logic isn't coupled to view attached/detached
to window callbacks. Instead, `unmountReactApplication` was added.
We can safely call it when a screen is popped and when application is
reloaded in debug.


DanielZlotin:
Umm this is incorrect I think.. talk to me

Issue #284, opened 2016-09-06
switchToTab doesn't work after 2.0.0-experimental.96


agumack:
Hi, switchToTab doesn't work after 2.0.0-experimental.96
Log: could not find UITabBarController

guyca:
Seems to work fine in the example project

agumack:
sorry @guyca 

It work on Android only.
IOS not working. "could not find UITabBarController"

guyca:
No worries, thanks for reporting @agumack

agumack:
Detail log:

```
2016-09-07 13:54:06.621 [warn][tid:com.facebook.react.JavaScript] Possible Unhandled Promise Rejection (id: 6):
could not find UITabBarController
createErrorFromErrorData@http://192.168.1.101:8081/index.ios.bundle?platform=ios&dev=true:7460:20
http://192.168.1.101:8081/index.ios.bundle?platform=ios&dev=true:7411:35
__invokeCallback@http://192.168.1.101:8081/index.ios.bundle?platform=ios&dev=true:7274:15
http://192.168.1.101:8081/index.ios.bundle?platform=ios&dev=true:7145:24
guard@http://192.168.1.101:8081/index.ios.bundle?platform=ios&dev=true:7072:3
invokeCallbackAndReturnFlushedQueue@http://192.168.1.101:8081/index.ios.bundle?platform=ios&dev=true:7144:6
invokeCallbackAndReturnFlushedQueue@[native code]
```

edvinasbartkus:
I find that `navigatorID == undefined` when received in [navigatorSwitchToTab](https://github.com/wix/react-native-navigation/blob/master/src/deprecated/platformSpecificDeprecated.ios.js#L363). Therefor, it tries to find `undefined_tabs` component in iOS code which results in failure. However, I think the issue was introduced much earlier. Can't find which version exactly.

sogko:
I encountered the same `could not find UITabBarController` exception (on iOS) recently and fix #290 did not address it in my case. 

Took awhile to track it down; finally figured out what caused my tab bar component to be missing.
Hope this helps out someone else down the line.

-----

```
- "react-native-navigation": "^2.0.0-experimental.127"
- iOS only
```

In my case:

The source of the exception was from `navigator.setTabBadge()`. I had it hooked up to an event handler that updates the badge when a value changes.

```js
onValueChanged = (size) => {
  this.props.navigator.setTabBadge({
        tabIndex: 1,
        badge: size
[...]

PR #290, opened 2016-09-07
Generate controllerID before using it in tabs.map

edvinasbartkus:
Fixes https://github.com/wix/react-native-navigation/issues/284

`controllerID` was generated when used in `params.tabs.map(function(tab, index) {...`. Moving it before the `.map` fixes the issue as each screen has correct name reference for controllerID.


gran33:
@edvinasbartkus Good one, 10x!

PR #327, opened 2016-09-21
Create tabs Left to Right

yedidyak:


DanielZlotin:
This is a fix not for the creation of the tabs per se, but for the bug in react-native which creates the content in a LIFO matter, ie, when calling `startReactApplication` from multiple contentViews, the last one to call is the *first* one to be rendered. weird huh?

PR #357, opened 2016-10-05
[Android] Fix for RN33 onActivityResult interface change

beast:
Retrieve current activity from reactInstanceManager in order to listen to onActivityResult interface change in RN33 or above. This also ensure result is passed back to React instance so other modules are able to get onActivityResult called.

### DIFF
--- a/android/app/src/main/java/com/reactnativenavigation/react/NavigationReactGateway.java
+++ b/android/app/src/main/java/com/reactnativenavigation/react/NavigationReactGateway.java
@@ -74,7 +74,8 @@ public class NavigationReactGateway implements ReactGateway {
     }
 
     public void onActivityResult(int requestCode, int resultCode, Intent data) {
-        getReactInstanceManager().onActivityResult(requestCode, resultCode, data);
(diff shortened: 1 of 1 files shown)


DanielZlotin:
This is a breaking change, once we update we'll merge.

tylerclark:
Damn, wish I would have saw this when I made [this post](https://github.com/wix/react-native-navigation/issues/359) earlier!

alidawud:
👍🏼

andybbtt11:
Works for me on 0.35.0, thanks for the temp fix.

glesage:
Any updates on this? is there anything to be done before it can be merged?

guyca:
@glesage Before we can upgrade the supported RN version, we need all of our development teams to upgrade their modules/apps to latest version. Adopting the code base to new RN version takes some time, we can't afford to dedicate whole day every two weeks/month for this task. Another issue is that most of the team is on vacation since last week of September until the end of October because of the Jewish holidays. So upgrading during October was simply not possible for that reason alone.

I'm sorry this task is taking so long, hope we can resolve this by next week.

PR #371, opened 2016-10-10
Properly define header files in Navigation project

LeoNatan:
- Add Navigation project as target dependency of demo project; Change import statement to use correct header.


DanielZlotin:
Figure out why the ppl at facebook decided to stick with static libs and abandoned the use of frameworks...

Issue #394, opened 2016-10-23
You attempted to set the key `onPress` with the value...


WhatAKitty:
### Steps to Reproduce / Code Snippets / Screenshots
1. Before the component will mount, I set rightButtons;
2. Then I setOnNavigatorEvent to listen button event;
3. When I click the button what I add. The first time, it worked, but when I turned back and clicked it again. It broken.

![image](https://cloud.githubusercontent.com/assets/4178729/19624707/c3262a4e-9935-11e6-9673-76ee64c8dac2.png)
---
### Environment
- React Native Navigation version: 2.0.0-experimental.133
- React Native version: 0.35.0
- Platform(s) (iOS, Android, or both?): IOS
- Device info (Simulator/Device? OS version? Debug/Release?): Simulator IOS 10.0 Debug

WhatAKitty:
I run your example code right now. And I found the same issue.
![10 -23-2016 16-17-51](https://cloud.githubusercontent.com/assets/4178729/19624923/523e64c0-993c-11e6-9595-01c819fa8a28.gif)

sexykodo:
Same problem here, after updating to RN 0.35

JuneDomingo:
@sexykodo @WhatAKitty RN 0.35 is not yet fully supported, but this might help you #357

WhatAKitty:
@JuneDomingo It seems worked on Android. How about IOS?

beast:
@WhatAKitty I am the author for https://github.com/wix/react-native-navigation/pull/357. It only affects Android. It's not related here. However, I believe somewhere along the way React Native fixed the way they handle listview datasource (Making it immutable). I suggest you have a look at the accepted answer on http://stackoverflow.com/questions/38270445/you-attempted-to-set-the-key-on-an-object-that-is-meant-to-be-immutable-and-has

beast:
@guyca I am reviewing the source code for this, to be honest I am a backend developer that just got into frontend and mobile world. So bear with me if I got it wrong. Error is thrown at 199 [platformSpecificDeprecated.ios.js](https://github.com/wix/react-native-navigation/blob/master/src/deprecated/platformSpecificDeprecated.ios.js)

```javascript
 const navigatorButtons = Object.assign({}, screenClass.navigatorButtons);

  if (navigatorButtons.rightButtons) {
    for (let i = 0; i < navigatorButtons.rightButtons.length; i++) {
      navigatorButtons.rightButtons[i].onPress = navigatorEventID;
[...]

PR #397, opened 2016-10-25
Fix immutable error after updating rn@0.35

WhatAKitty:
using deep clone. #394

### DIFF
--- a/src/deprecated/platformSpecificDeprecated.ios.js
+++ b/src/deprecated/platformSpecificDeprecated.ios.js
@@ -184,10 +184,10 @@ function _mergeScreenSpecificSettings(screenID, screenInstanceID, params) {
     Object.assign(navigatorStyle, params.navigatorStyle);
   }
 
-  const navigatorEventID = screenInstanceID + '_events';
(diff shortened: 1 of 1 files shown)


indivisable:
This fixed some issues I had with 0.31 as well (having to do with opening Models) - Reference: https://github.com/wix/react-native-navigation/issues/394

rawrmaan:
Just a heads up for MobX users, this will cause you to lose referential integrity and observability when passing props to a new view or modal via `passProps`.

DanielZlotin:
I'm accepting this as a temporary fix, will be removed before 2.0.0

stonecold123:
@rawrmaan 
I'm using the @next version. 
I tried passing the mobx store via `passProps` to another screen and update a new value in the store, 
I'm getting 
````
You attempted to set the key K with the value X on an object that is meant to be immutable and has been frozen
````
What's your workaround for this problem?

stonecold123:
The issue with passing mobx via passProps not being observable still remains. Is there any solutions?

PR #426, opened 2016-11-09
iOS TabBar : Image gets center if there is no title

Crash--:
If there is no label, then the icon will be vertically centered

```jsx
Navigation.startTabBasedApp({
      tabs: [
        {
          screen: 'InTeach.CoursesScreen',
          icon: icons['graduation'],
          //label: 'My Label',
          selectedIcon: icons['graduation'],
          title: language('navCourses'),
          navigatorStyle: {
            navBarHidden: true,
            tabBarButtonColor: 'rgba(255,255,255,0.5)',
            tabBarBackgroundColor: colors.mainColor,
            tabBarSelectedButtonColor: colors.white,
            statusBarTextColorScheme: 'light'
          }
        }
```

I need to do the same on Android,if someone has an idea on how to do it... ;)

### DIFF
--- a/ios/RCCTabBarController.m
+++ b/ios/RCCTabBarController.m
@@ -94,6 +94,12 @@
     viewController.tabBarItem = [[UITabBarItem alloc] initWithTitle:title image:iconImage tag:0];
     viewController.tabBarItem.accessibilityIdentifier = tabItemLayout[@"props"][@"testID"];
     viewController.tabBarItem.selectedImage = iconImageSelected;
+    //if there is no title, then the image will be centered
+    if(!title){
+      int offset = 7;
+      UIEdgeInsets imageInset = UIEdgeInsetsMake(offset, 0, -offset, 0);
+      viewController.tabBarItem.imageInsets = imageInset;
+    }
     
     if (buttonColor)
     {


gran33:
@Crash-- Thanks for your contribution.

2 Things:
1. Can u avoid from magic number? (I looked a bit, and it seems that the simplest solution is magic number, maybe u can do some calculations and avoid from adding them).
2. U better change your `if` statement to be like this:
```if(!title || (title && [title length] == 0))```
in order to solve existing `label` but without text:
```label: ''```

Plz check that the changes I offered r ok with u, and if so, submit a correction.

10x!

File: ios/RCCTabBarController.m [hunk 1]
```diff
@@ -94,6 +94,12 @@ - (instancetype)initWithProps:(NSDictionary *)props children:(NSArray *)children
     viewController.tabBarItem = [[UITabBarItem alloc] initWithTitle:title image:iconImage tag:0];
     viewController.tabBarItem.accessibilityIdentifier = tabItemLayout[@"props"][@"testID"];
     viewController.tabBarItem.selectedImage = iconImageSelected;
+    //if there is no title, then the image will be centered
+    if(!title){
```
gran33: U better check the case when `title` exist but empty:
```if(!title || (title && [title length] == 0))```

File: ios/RCCTabBarController.m [hunk 2]
```diff
@@ -94,6 +94,12 @@ - (instancetype)initWithProps:(NSDictionary *)props children:(NSArray *)children
...
+      int offset = 7;
```
gran33: Is there a way to avoid from magic number?

Crash--: I'll try to find a new way

Crash--: I didn't find something that works well without this magic number :/

Crash--:
The only way to achieve the same thing for Android is to upgrade `ahbottomnavigation` to v2 because `setTitleState(AHBottomNavigation.TitleState.ALWAYS_HIDE);` was added 

Is it a problem?

guyca:
@Crash-- not a problem at all

grundmanise:
@gran33 Hey, may you check & merge this pull? :)

gran33:
@grundmanise I'll try to do it ASAP, sorry for the late delay, short in time a bit.

grundmanise:
@Crash--  , @gran33 any updates on this? Would be very happy if this merge.

EdwardNelson:
Any idea if this will be merged soon?

guyca:
@EdwardNelson 
On iOS you can set [iconInsets](https://wix.github.io/react-native-navigation/#/top-level-api?id=starttabbasedappparams) to center the icons.
On Android icon should be centered if title isn't provided.

chrise86:
@guyca in what version is `iconInsets` supported? It doesn't seem to work in v1 :-/

guyca:
@chrise86 Indeed it's not implemented in v2. It's currently supported in 1.1.x

PR #433, opened 2016-11-13
When starting the app, pass through extras included in the launch intent

MattDavies:
It's great that we now have the DeepLinkHandler to pass through deep links!

It would be really handy to also pass through extras from the intent - there's a variety of uses, but the main one is making sure we don't lose push notification data (which gets set in the intent extras) when starting NavigationApplication.


guyca:
cc @amitdwix

Issue #358, opened 2016-10-05
White flash on screen push.


sexykodo:
### Issue Description
In IOS when i push a new screen i see a white background on the rootview before the first react native render, since i have a dark styled app it's quite annoying.

Changing the color on rootview in AppDelegate.m to black,
or setting a screenBackgroundColor doesn't seem to to anything 

### Steps to Reproduce / Code Snippets / Screenshots

This is the snippet from AppDeletage.m
`self.window = [[UIWindow alloc] initWithFrame:[UIScreen mainScreen].bounds];
  self.window.backgroundColor = [UIColor blackColor];`

This is the snipper from the screenComponent

`class EventScreen extends Component {
    static navigatorStyle = {
        screenBackgroundColor: SituaColors.darkBg
    };
   ....
`
Thanks
---
### Environment
* React Native Navigation version: 2.0.0-experimental.105
* React Native version: 0.34.0
* Platform(s) (iOS, Android, or both?): IOS
* Device info (Simulator/Device? OS version? Debug/Release?): Both simulator and device, both debug and release

alexcurtis:
Would be great to get some options for this. I also have a dark themed app and the white flashes between scenes are frustrating.

DavoCg:
Do we know why this kind of flashing occurs ? 
I had not them before and since few changes in my code I had one flash only between two specific screens.

ryanscottaudio:
same issue here; happens on push between two screens with `navBarHidden: true`

savelichalex:
I add this as temp solution for my project under `setStyleOnAppearForViewController` in `RCCViewController`
```objective-c
NSString *screenColor = self.navigatorStyle[@"screenColor"];
    if (screenColor)
    {
      UIColor *color = screenColor != (id)[NSNull null] ? [RCTConvert UIColor:screenColor] : nil;
      viewController.view.backgroundColor = color;
    }
```

sexykodo:
@savelichalex thanks! works like a charm!

DavoCg:
Thanks @savelichalex it works well but does somebody a way to avoid this flashes ? maybe with preloading or idk ?

savelichalex:
Maybe @guyca help with it, but I don't think so. As I understand how it works, when `push` happened VC create new native view(1), then they send event to js that render react view, and then rendered new react view added to native view(2). Between this two phases we see default native view with default background color.

guyca:
[...]

PR #437, opened 2016-11-14
Ability to set a screenBackgroundColor on iOS

MattDavies:
By default, the iOS view colour seems to be white. Without this option, I still get a white screen briefly while the first component loads.

Only really relevant to iOS as you can set the Android view background in the manifest. If I'm just doing something wrong and this isn't necessary, I'm all ears!


lakhman:
Related to https://github.com/wix/react-native-navigation/issues/358

MattDavies:
Ah. @guyca is this useful or do you want me to close it in favour of what you're working on already as per https://github.com/wix/react-native-navigation/issues/358 ?

PR #511, opened 2016-11-29
softInputMode activity param (Android only)

adamski:
To allow TextInputs in ListViews that would be obscured by the keyboard (i.e. at the bottom of the screen) to move so that they are visible, it is necessary to set `softInputmode` to `adjustPan`.  
If this is not set, the keyboard appears and then disappears immediately. 
See e.g. https://github.com/facebook/react-native/issues/8180

This PR allows the `softInputMode` to be set from JS via `startApp`. It expects a numerical string, e.g.
```
softInputMode: '32' 
```
(I tried sending as an Int but I got a ClassCastException as it seems to be transmitted as a Double?)

It would be more user friendly to also be able to send a string representation like you would in the manifest e.g. 
```
softInputMode: 'adjustPan'
```
This could be achieved with a map representing the values listed at https://developer.android.com/reference/android/view/WindowManager.LayoutParams.html#softInputMode. I'm happy to implement this but wanted to get feedback first.

### DIFF
--- a/android/app/src/main/java/com/reactnativenavigation/controllers/NavigationActivity.java
+++ b/android/app/src/main/java/com/reactnativenavigation/controllers/NavigationActivity.java
@@ -3,7 +3,9 @@ package com.reactnativenavigation.controllers;
 import android.content.Intent;
 import android.os.Bundle;
 import android.support.v7.app.AppCompatActivity;
+import android.util.Log;
 import android.view.KeyEvent;
+import android.view.WindowManager;
 
 import com.facebook.react.bridge.Callback;
 import com.facebook.react.modules.core.DefaultHardwareBackBtnHandler;
@@ -53,6 +55,8 @@ public class NavigationActivity extends AppCompatActivity implements DefaultHard
         }
 
         activityParams = NavigationCommandsHandler.parseActivityParams(getIntent());
+
+        getWindow().setSoftInputMode(activityParams.softInputMode);
 
         disableActivityShowAnimationIfNeeded();
         createLayout();
--- a/android/app/src/main/java/com/reactnativenavigation/controllers/NavigationActivity.java
+++ b/android/app/src/main/java/com/reactnativenavigation/controllers/NavigationActivity.java
@@ -3,9 +3,7 @@ package com.reactnativenavigation.controllers;
 import android.content.Intent;
 import android.os.Bundle;
 import android.support.v7.app.AppCompatActivity;
-import android.util.Log;
 import android.view.KeyEvent;
-import android.view.WindowManager;
 
 import com.facebook.react.bridge.Callback;
 import com.facebook.react.modules.core.DefaultHardwareBackBtnHandler;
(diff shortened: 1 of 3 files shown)


DanielZlotin:
Thanks for the contribution, but this is not the way I would handle the problem.
In general if your layout is defined correctly (not rigid, no hardcoded width/height, can shrink and grow to accommodate all screen sizes and variations) you will 90% of the time have no problems.

When you do encounter issues with the layout when the keyboard is shown, you basically have 2 options: 
1. You should use the `resize` mode and make you layout scrollable -this way the input field will lift into the correct location and no other elements be hidden, the user can scroll to show the rest of the layout, in general the UI looks better.
2. You can place the input field aligned to the bottom of the layout (in android this basically means `Align parent bottom`, see the input field in Whatsapp chat) - this basically makes the bottom input field "float" above the rest of the layout.

`pan` mode is less predictable and can be harder to control, so I would recommend not to use it, and this is why we hardcoded `resize` in navigation. 

If you really do want to use `pan` mode, your solution is not enough in any case. You can actually attach an Activity Lifecycle listener through your application object and force soft input mode on each `onCreate` of each activity.

adamski:
Thanks for the info. The solution works for my use case. I was trying to generalise it. Please can you point me to documentation on "attach an Activity Lifecycle listener through your application object and force soft input mode on each onCreate of each activity". 

I have a ListView with a TextInput on each line. Without `pan` mode the keyboard appears and immediately disappears on the lower TextInputs. It's almost as if it starts to move but not fast enough and then triggers a `blur` and keyboard disappears. My layout is scrollable - its basically one big ListView. If you have another solution I'd be happy know.

DanielZlotin:
https://developer.android.com/reference/android/app/Application.html#registerActivityLifecycleCallbacks%28android.app.Application.ActivityLifecycleCallbacks%29

DanielZlotin:
But this sounds more than a layout-keyboard problem.. something causes your keyboard to disappear, which means either some hacky code calls `adjust soft input` or something steals the focus..

DanielZlotin:
Just to clarify, the TextInputs are below the list?

adamski:
No the TextInputs are *in* the list. Each item in the list is two TextInputs.

adamski:
It seems to be a known issue, see the RN issue I linked to above.

DanielZlotin:
Oh wow. I see. So as I said this seems like something is messing with the focusability. So until they fix this issue you can use the `ActivityLifecycleCallbacks` or you can actually check [this out](https://github.com/wix/react-native-navigation/blob/master/android/app/src/main/java/com/reactnativenavigation/NavigationApplication.java#L57)

adamski:
Thanks for the tip! ActivityLifecycleCallbacks actually didn't work, but extending ActivityCallbacks, overriding `onActivityCreated` and passing it to `setActivityCallbacks` did the trick. Thanks for that. 

Maybe worth adding this to the docs somewhere, if its a known RN issue? Due to the way RNN creates its own Activities the usual methods may not work.

adamivancza:
If you still looking for something similar solution, try this package: https://github.com/jreziga/react-native-android-keyboard-adjust

PR #525, opened 2016-12-02
Added "AppTheme"

brianjd:
- The RNN container should use the standard RN theme


guyca:
Hey @brianjd 
Thanks for the PR.
We previously had the `theme` property defined but users ran into manifest merge issues so we removed it.

PR #578, opened 2016-12-19
Pass the style parameter to DrawerControllerIOS

megahertz:
Now the style parameter is used inside RCCDrawerController but it's not passed through startTabBasedApp/startSingleScreenApp. Passing the style parameter allows to disable sidebar shadow trough style.drawerShadow.


megahertz:
It seems that there is a problem with travis. Trying to close and reopen to run it again.

PR #587, opened 2016-12-21
Support for using a custom UIBarButtonItemStyle for iOS navigation bar buttons

JosephDuffy:
This adds support for a new key: style. When the style key’s value is "done", UIBarButtonItemStyleDone will be used.

To accommodate existing apps and to stick to iOS standards, UIBarButtonItemStylePlain remains the default.

If this PR is approved I will create another PR for changes to the wiki.


benkraus:
Any possibility of getting this in? Building modals, with non-bolded weights, and it doesn't look to pretty. This would help immensely to make it look better.

PR #593, opened 2016-12-22
Adding a custom button to the tab bar

DanielMaly:
I have added the option to have a custom middle button on the iOS tab bar which sends an event to JS when it gets clicked. Here is what it looks like:

screenshot 2016-12-22 16 47 48

You can define it in `startTabBasedApp` like this:

```
Navigation.startTabBasedApp({
	tabs,
	title: 'My Application',
	animationType: 'slide-down',
	middleButton: {
		height: 65,
		width: 150,
		icon: require('resources/img/icon_add.png'),
		onclicked: () => {console.log('Middle tab bar button clicked')}
	}
})
```

The button will only appear if there are 2 or 4 other tabs defined. The TabBarViewController creates a dummy ViewController under the middle tab and it is the responsibility of the calling code to make sure the button is big enough to obscure the tab completely.

I also added an EventEmitter module that doesn't use the now deprecated `bridge.eventDispatcher`.

### DIFF
--- a/ios/RCCEventEmitter.h
+++ b/ios/RCCEventEmitter.h
@@ -1,11 +1,3 @@
-//
-//  RCCEventEmitter.h
-//  ReactNativeNavigation
-//
-//  Created by Daniel Maly on 22/12/2016.
-//  Copyright © 2016 artal. All rights reserved.
-//
-
 #import "RCTEventEmitter.h"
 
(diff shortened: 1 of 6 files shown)


carlosroberto555:
This pull was abandoned?
Is there another way to make custom button?

sheva7x7:
Is this in any of the milestones?

PR #600, opened 2016-12-26
Send bottomTabSelected event

guyca:
The event is sent to the top most screen displayed in the current screenStack.


messense:
Will iOS version support this?

guyca:
Yes of course, hopefully this week.

messense:
Seems that on Android this event will only be sent when switch between different tabs?
Meanwhile I find that on iOS the event is sent every time I tap on tab bar regardless of selecting same or different tabs.

guyca:
The event should be sent only when the active bottom tab changed.  @messense Can you please open an issue?

PR #605, opened 2016-12-27
added tab selected event

gran33:


rawrmaan:
I haven't been able to get this to work on iOS. In one of tabs' components, I'm listening for `onNavigatorEvent`, which I have properly registered in the constructor, but no event of any kind is received when its tab or any other tab is selected.

adamski:
Looks like I missed this - so I should now be able to detect which screen is currently selected via bottom tabs on both iOS and Android?

guyca:
@adamski yes

Issue #616, opened 2016-12-28
How to get a left button event in Android?


SudoPlz:
Hey there @guyca,

I've been manually keeping track of wether the side menu is open or not in iOS by adding a custom left button as an object in my `leftButtons`, toggling the drawer and keeping a `boolean` in my redux state to know if it's open or not. 

But when I tried to make the same thing in Android, I found that I **cannot overwrite** the `leftButton` with a custom one, and thus I don't know when it's clicked, and thus I'm not able to tell if the side menu is open or not..

How come don't you dispatch an event when a button is clicked?

Any ideas about this?

Thank you..!

guyca:
Thanks for the PR, I wrote one comment, think there's one small mistake but other then that looks good 👍

SudoPlz:
Hmm, can't really find that comment, any idea where it is?

guyca:
merged. Thanks @SudoPlz

SudoPlz:
Thank you!!

PR #617, opened 2016-12-28
Sends a nav event whenever the sideMenu button is clicked. (Android)

SudoPlz:
Just an effort to get around this: https://github.com/wix/react-native-navigation/issues/616


SudoPlz:
Thanks!!

PR #636, opened 2017-01-05
Ability to pass options to Redux Provider

guns2410:
With this merge request, we will have the ability to supply additional options to the Provider.

This is commonly used to libraries like `react-apollo` where client is to be passed to the provider.

Usage
```javascript
Navigation.registerComponent(screenName, () => screen, store, ApolloProvider, { client: GraphQLClient });
```


godlevel:
I would love to see this merged.. Any plans?

PR #655, opened 2017-01-10
Fix popToRoot when keyboard is visible

varungupta85:
Before this, if we have multiple screens on the stack and the keyboard is visible and we call `popToRoot`, then the screens don't pop as expected. This is because of the way the screen is popped if the keyboard is visible and when the keyboard is not visible. When the keyboard is visible, we attach a keyboard close handler in which we pop the screen and close the keyboard which would eventually invoke the keyboard close handler and pop the screen. If they keyboard is not visible, we directly pop the screen.

If we call `popToRoo`t, since we call pop in a while loop, in the first call to pop, we detect that the keyboard is visible, hence attach the keyboard close handler and close the keyboard which would eventually invoke the keyboard close handler and we will pop the screen. But since the event handling would be done asynchronously, we end up calling the next pop in the while loop before the close keyboard handler is invoked. In the second pop, since according to the code, the keyboard is not visible, we directly pop the screen. And then keyboard close handler is finally invoked but at that time, parent for the current screen has already been unmounted, so it doesn't pop the screen and we continue to see the topmost screen

The fix is to start popping screens after the keyboard has been closed and to emit the `ScreenChangedEvent` after the last screen has been popped. I have applied the fix to the `pop` API as well because it may be affected as well due to the asynchronous nature of the keyboard  close handler.


varungupta85:
@guyca Removed the comments and tried to work on the DRY comment for the keyboard visibility stuff. This is the best I could come up with where we separately call keyboard visibility code for `pop` and `popToRoot`. It is still repeated but it is not called twice for `popToRoot`. I couldn't find an easy way of having the keyboard visibility stuff just once. I checked if I can the function to call inside the keyboard close listener as a parameter but it looks like we can only pass around functions with same signatures. Other thing I could think of was having an interface and have two classes which implement the interface - one for `pop` and the other for `popToRoot` but I thought that it would make it too complicated for no obvious returns.

If you have a better way of handling it, please let me know.

PR #669, opened 2017-01-17
[android] Add option for scrollable top tabs

rusel1989:
Small changes to enable scrollable top tabs in android


guyca:
Thanks @rusel1989 👍

PR #675, opened 2017-01-20
Allow the user to use his own navigatorButtons when calling showModal, push etc..

SudoPlz:
Android: If the user has specified navigatorButtons on the showModal params use those instead of the default navigatorButtons of the registered screen.

If I don't add that if, the existing code erases the navigatorButtons I pass on my params object.


SudoPlz:
Awesome thanks!

PR #687, opened 2017-01-25
dismissInAppNotification support on android.

Royce:
Add .dismissInAppNotification on android.

- Separate hide and show animations.
- Implement a little bit of throttling around successive calls to .showInAppNotification

### DIFF
--- a/android/app/src/main/java/com/reactnativenavigation/animation/PeekingAnimator.java
+++ b/android/app/src/main/java/com/reactnativenavigation/animation/PeekingAnimator.java
@@ -1,7 +1,6 @@
 package com.reactnativenavigation.animation;
 
 import android.animation.Animator;
-import android.animation.AnimatorSet;
 import android.animation.ObjectAnimator;
 import android.view.View;
 import android.view.animation.OvershootInterpolator;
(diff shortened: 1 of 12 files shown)


android/app/src/main/java/com/reactnativenavigation/animation/PeekingAnimator.java
```diff
@@ -28,33 +26,26 @@ public void animate() {
         animator.start();
     }
 
-    private Animator createAnimator(View view) {
+    private Animator createAnimator(View view, boolean show) {
```
guyca: This is just nitpicking, but we can avoid the boolean and simply have two public methods:
`createShowAnimator(View view)`
`createHideAnimator(View view)`

Royce:
@guyca That method (createAnimator) wasn't really required, so I removed it (and rebased), but I still pass the "show" boolean to the PeekingAnimator ctor.

Alternatively, I could make it more like https://github.com/wix/react-native-navigation/blob/master/android/app/src/main/java/com/reactnativenavigation/animation/VisibilityAnimator.java#L43

varungupta85:
@Royce Thanks for the PR. Would the behavior for `showInAppNotification` be changed a little bit for Android after this PR? Before this PR, the in-app notification would appear and disappear shortly. There was no control to keep it there and dismiss it when I want or dismiss it after a set interval. With the addition of `dismissInAppNotification` on Android, would I be able to keep the in-app notification in the window till I explicitly call the dismiss method?

Royce:
I'll be doing a few more PRs soon to address things like that.

PR #696, opened 2017-01-29
Notification auto dismiss timer

Royce:
[Android] Control the duration of the auto dismiss timer with `autoDismissTimerSec` like on iOS
[Android] Default `autoDismissTimerSec` to 5 like on iOS
[All] Add `autoDismiss` param to enable/disable auto dismiss behaviour. Defaults to `true`

### DIFF
--- a/android/app/src/main/java/com/reactnativenavigation/params/SlidingOverlayParams.java
+++ b/android/app/src/main/java/com/reactnativenavigation/params/SlidingOverlayParams.java
@@ -17,5 +17,6 @@ public class SlidingOverlayParams {
 
     public String screenInstanceId;
     public NavigationParams navigationParams;
+    public Integer autoDismissTimerSec;
     public Position position;
 }
--- a/android/app/src/main/java/com/reactnativenavigation/params/parsers/SlidingOverlayParamsParser.java
+++ b/android/app/src/main/java/com/reactnativenavigation/params/parsers/SlidingOverlayParamsParser.java
@@ -11,6 +11,9 @@ public class SlidingOverlayParamsParser extends Parser {
         final SlidingOverlayParams result = new SlidingOverlayParams();
         result.screenInstanceId = bundle.getString("screen");
         result.navigationParams = new NavigationParams(bundle.getBundle("navigationParams"));
+        result.autoDismissTimerSec = bundle.containsKey("autoDismissTimerSec")
+                ? bundle.getInt("autoDismissTimerSec")
+                : null;
         result.position = SlidingOverlayParams.Position.fromString(bundle.getString("position", "top"));
         return result;
     }
--- a/android/app/src/main/java/com/reactnativenavigation/views/slidingOverlay/SlidingOverlaysQueue.java
+++ b/android/app/src/main/java/com/reactnativenavigation/views/slidingOverlay/SlidingOverlaysQueue.java
@@ -9,7 +9,6 @@ import java.util.TimerTask;
 
 public class SlidingOverlaysQueue implements SlidingOverlay.SlidingListener{
 
-    private static final int SUSTAIN_DURATION = 3000;
     private static final int SHORT_SUSTAIN_DURATION = 500;
 
     protected Timer autoDismissTimer = null;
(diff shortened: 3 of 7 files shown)


android/app/src/main/java/com/reactnativenavigation/params/SlidingOverlayParams.java
```diff
@@ -17,5 +17,6 @@ public static Position fromString(String string) {
 
     public String screenInstanceId;
     public NavigationParams navigationParams;
+    public Integer autoDismissTimerSec;
```
guyca: Shouldn't the dismiss time be supplied as milliseconds?

android/app/src/main/java/com/reactnativenavigation/views/slidingOverlay/SlidingOverlaysQueue.java
```diff
@@ -60,23 +59,27 @@ else if (currentOverlay.isVisible()) {
...
-                        queue.peek().hide();
-                    }
-                });
-            }
-        }, autoDismissDuration);
+        if (autoDismissTimerSec != null || pendingHide || queue.size() > 1) {
+            int autoDismissDuration = autoDismissTimerSec != null
+                    ? autoDismissTimerSec * 1000
+                    : SHORT_SUSTAIN_DURATION;
+            pendingHide = false;
```
guyca: Not sure I fully understand what `pendingHide` means

android/app/src/main/java/com/reactnativenavigation/params/parsers/SlidingOverlayParamsParser.java
```diff
@@ -11,6 +11,9 @@ public SlidingOverlayParams parse(Bundle bundle) {
         final SlidingOverlayParams result = new SlidingOverlayParams();
         result.screenInstanceId = bundle.getString("screen");
         result.navigationParams = new NavigationParams(bundle.getBundle("navigationParams"));
+        result.autoDismissTimerSec = bundle.containsKey("autoDismissTimerSec")
+                ? bundle.getInt("autoDismissTimerSec")
+                : null;
```
guyca: Aren't we always passing `autoDismissTimerSec` from js?
`params.autoDismissTimerSec || 5;`
In that case no need to check if bundle contains key etc and we can avoid null checks.

android/app/src/main/java/com/reactnativenavigation/params/SlidingOverlayParams.java
Royce: Yes, milliseconds would be superior, though it probably won't matter. X seconds is probably fine grained enough. Also, I also think matching the API that iOS uses is more important.

But I could add `autoDismissTimerMilli` as well as `...Sec`

android/app/src/main/java/com/reactnativenavigation/params/parsers/SlidingOverlayParamsParser.java
Royce: Not anymore because I need some way of disabling the auto dismiss. Also, this matches the iOS implementations. Doing this check in the native code gives us more flexibility in the javascript API.

android/app/src/main/java/com/reactnativenavigation/views/slidingOverlay/SlidingOverlaysQueue.java
Royce: If there is a call to hide/dismiss whilst the overlay is animating in, then I flip this flag to remember that we need to hide the overlay as soon as the overlay has finished appearing.

Alternative. Start the out animation straight away.

Royce: But, I'm not sure interrupting the show animation is always feasible/aesthetically pleasing (depending on the type of animation. e.g. Flip)

android/app/src/main/java/com/reactnativenavigation/params/SlidingOverlayParams.java
guyca: Fair enough. So lets leave it as is. We can always add milliseconds support later

android/app/src/main/java/com/reactnativenavigation/views/slidingOverlay/SlidingOverlaysQueue.java
guyca: Gotcha. Lets keep it as is and change/improve if needed later on

Issue #485, opened 2016-11-22
Cannot read property 'navigatorStyle' of undefined


florianbepunkt:
I tried to incorporate the library into my project, got the error in the title when running in simulator. my app initialization is this:

```js
import { Navigation } from 'react-native-navigation';

import { StartPage } from './app/startPage.js';
import { FirebaseTestPage } from './app/firebaseTest.js';

function registerScreens() {
  Navigation.registerComponent('roomary.StartPage', () => StartPage);
  Navigation.registerComponent('roomary.FirebaseTestPage', () => FirebaseTestPage);
}

registerScreens();

// start the app
Navigation.startTabBasedApp({
  tabs: [
    {
      label: 'One',
      screen: 'roomary.StartPage', // this is a registered name for a screen
      icon: require('./assets/img/one.png'),
      selectedIcon: require('./assets/img/one_selected.png'), // iOS only
      title: 'Screen One'
    },
    {
      label: 'Two',
      screen: 'roomary.FirebaseTestPage',
      icon: require('./assets/img/two.png'),
      selectedIcon: require('./assets/img/two_selected.png'), // iOS only
      title: 'Screen Two'
    }
  ]
});
```

---
### Environment
* React Native Navigation version: react-native-navigation@2.0.0-experimental.142 
* React Native version: 0.37.0
* Platform(s) (iOS, Android, or both?): iOS
* Device info (Simulator/Device? OS version? Debug/Release?): Simulator iPhone 6 iOS10.1

guyca:
Hey @florianbepunkt Did you manage to make progress? `navigatorStyle` is supposed to be optional.

florianbepunkt:
I was using this library in conjunction with firestack library… I guess somewhere in the installation process something got mixed up, but can't say for sure. If I install this library first and other libraries after this, it works fine. But not the other way around.

wpapper:
I'm running into the same issue after following the examples in the documentation. I tried overriding the navigatorStyle with custom styling, but it didn't fix the problem.

Is there a known solution?

wimgoeman:
I had the same issue. It took me some digging around in the code of this lib, but the reason is you are incorrectly registering the component. The function you use at regsitration time probably returns undefined. I am making a pull request to properly show this as an error.

guyca:
Thanks a lot @wimgoeman

bjacog:
[...]

PR #708, opened 2017-02-01
Add error when trying to load a component which was badly registered

wimgoeman:
I struggeled quite a while debugging #485 "Cannot read property 'navigatorStyle' of undefined"  which originates from incorrectly registering the component. In order to save time for other users I suggest this small change to give a more verbose error when this happens.


lashakv:
@wimgoeman can you please specify, why this change was only made in ### _registerComponentNoRedux ###. Won't it be useful to have the same check in ###_registerComponentRedux ### ?

PR #735, opened 2017-02-07
Additional snackbar style params

guyca:
Implement backgroundColor and textColor style params


moeinrahimi:
what about fontfamily is it supported ?

PR #738, opened 2017-02-07
fix (imports) switch to #import <React/….h> and remove references to …

pentateu:
Fix iOS compilation issues when trying to use with react 0.41 and above.

This is very preliminary. I still doing more tests to see if there are other issues or side effects.

I'm also not 100% sure that just removing RCTReloadNotification and the handlers onRNReload is ok.
It looks like onRNReload methods are just doing clean-up, which is not required anymore.


guyca:
Support for RN 0.41.2 was added in 503eef5 cc94e02 fc357ba
Thanks @pentateu

PR #771, opened 2017-02-16
Fix possible IllegalStateException crash 

acerbetti:
Crash description here
http://stackoverflow.com/questions/7575921/illegalstateexception-can-not-perform-this-action-after-onsaveinstancestate-wit


guyca:
Thanks!

PR #848, opened 2017-03-12
add app style

gran33:
Add app style in order to provide defaults style


artald:
@gran33  RCCAppStyler is nice! but it seems that it's not used anywhere to get the styles that are passed from JS?

Can you please check?

PR #850, opened 2017-03-13
Removed not existing link from readme

Gongreg:
There was no link for migration to v2 and changelog. It looks bad when dead links are on main readme, so I removed the link.


DanielZlotin:
heh

Issue #845, opened 2017-03-10
Feature |  Support background image


dzunglht:
Please help support screen background image so that when we push a screen, user can see a splash of background image rather than a background colour.
Tons of thanks

PR #861, opened 2017-03-15
Add ability to set background image when transitioning between screens.

urbancvek:
So I had a problem that app I'm working on needs to have a gradient background. The default color when transitioning between screens is white. But even though you can set it to a custom color it still doesn't look as it should.

Here's an issue #845 that this PR would solve. 

So the solution is to set `screenBackgroundImageName` which should be a `string`. And then in Images.xcassets add the image with the same name. It then picks it up and works as one would expect.

Will add android version too in the following days.


lakhman:
related:
- https://github.com/wix/react-native-navigation/pull/437
- https://github.com/wix/react-native-navigation/issues/358

guyca:
Thanks for the contribution @urbancvek.

grundmanise:
@urbancvek Hi, why didn't you go by this solution? http://stackoverflow.com/a/23074591/4541800

urbancvek:
@grundmanise Hmm yeah it definetely could also be implemented in this way too. But then you would also have to use some sort of gradient library for background of the views on JS side. Do you think there is a big performance hit using it this way?

grundmanise:
I use [this gradient lib](https://github.com/react-native-community/react-native-linear-gradient) on the JS side. For me it seems easier and more 'correct' approach than to create bg images just for the gradient.. Don't know about a performance diff, but maybe we can add a new param: `screenBackgroundColorGradient` for the gradient, and use your solution (param) for anything else (true background images)? What do you think?

urbancvek:
Ok will try and implement later today. Have to ship now! 😅
How would you think the api would look like?
```js
{
  navigatorStyle: {
    screenBackgroundColorGradient: [#startColor, #endColor]
  }
}
```

Also how would we define direction of the gradient (top-bottom or left-right).

grundmanise:
Super! :)))
API looks good to me.
To define a directions we can get a hint from [here](https://github.com/react-native-community/react-native-linear-gradient/blob/master/BVLinearGradient/BVLinearGradient.m) & [here](https://github.com/react-native-community/react-native-linear-gradient#additional-props)

urbancvek:
Are you including `@2x`and `@3x` sizes in Xcode? Like this: 
screen shot 2017-04-11 at 16 38 36

For me the sizes of the images are: 
- @1x - 375x667
- @2x - 750x1334
- @3x - 1125x2001

eeynard:
I just found the problem, the background image wasn't stretch so I created this PR #1032.

I am not an iOS expert so let me know if i did something wrong.

Issue #645, opened 2017-01-07
Swiping / pop gesture on root Screen breaks navigation push / touches.


Royce:
### Issue Description

In the example app, swiping (a pop gesture) at the edge of the root Screen causes touches to the ContentView etc. have no effect. Eg. tapping the “Push Plain Screen” link engages the TouchableHighlight animation but does not cause the navigation to push.

After further swiping the screen, eventually, the ignored touches/navigation start having an effect. When this happens the phone behaves as if mid-way through a pan gesture / pushing a Screen

Note: If the app has a SideMenu, The ContentView inside the SideMenu does not stop responding to touches.

### Steps to Reproduce / Code Snippets / Screenshots

Build and run the example app (preferably disable the drawer/SideMenu) as it makes the issue clearer.

To replicate, whilst on the root Screen, swipe from the edge of the physical screen. Now tap on a link to push a Screen. Nothing happens. :(

After this point the Screen stops responding to all touches.

Swiping at the left of the screen again makes the problem go away…

See this video: https://drive.google.com/file/d/0B6eyHwHL_Zq2OFZEdkt4anF1dzg/view?usp=sharing

---
### Environment
* React Native Navigation version: 2.0.0-experimental.165 (also .141)
* React Native version: 0.38.0
* Platform(s): iOS
* Device info: 4S; 5S; 6 Plus; 6S

Royce:
Can someone explain what this code does in RCCViewController.m as commenting it out will make the problem disappear, yet I can still use a swipe gesturing to pop the second Screen...

```objective-c
//Bug fix: in case there is a interactivePopGestureRecognizer, it prevents react-native from getting touch events on the left screen area that the gesture handles
//overriding the delegate of the gesture prevents this from happening while keeping the gesture intact (another option was to disable it completely by demand)
self.originalInteractivePopGestureDelegate = nil;
if (self.navigationController != nil && self.navigationController.interactivePopGestureRecognizer != nil)
{
  id <UIGestureRecognizerDelegate> interactivePopGestureRecognizer = self.navigationController.interactivePopGestureRecognizer.delegate;
  if (interactivePopGestureRecognizer != nil)
  {
    self.originalInteractivePopGestureDelegate = interactivePopGestureRecognizer;
    self.navigationController.interactivePopGestureRecognizer.delegate = self;
  }
}
```

guyca:
cc @artald

gilra2000:
1+

lakhman:
[...]

Issue #428, opened 2016-11-10
iOS pushed screen never shows up and freeze the app


psyzz:
Hello,

I encountered a bug on iOS when pushing a new screen.
Basically i'm clicking on an item, calling this.props.navigator.push( {screen: 'app.myscreen'} )
and sometimes the screen never open.
To make the app "clickable" again i need to swipe back (like if i wanted to close the pushed screen) and the screen appears.

---
### Environment
* React Native Navigation version: ^2.0.0-experimental.136
* React Native version: 0.35.0
* Platform(s): iOS
* Device info: Simulator/Device - Debug/Release

arshavinho:
By testing another of my app made with RNN I just realized that this bug is also present and appears quite often

guyca:
@arshavinho We weren't able to reproduce this bug and no other user reported a bug with pushing screens. If you upload a project where this bug reproduces to GitHub, I'm sure someone will be glad to take a look at it. It's probably an issue in your app's logic. I doubt it's related to RNN. 👍

arshavinho:
ok i will try to create a project to reproduces the bug

arshavinho:
After a discussion with @guyca i was able to find this maybe related issue:
https://github.com/wix/react-native-navigation/issues/245

i tried to add the piece of code from littlesome post but the problem was still there and i couldnt swipe back to close the non-open screen. Here a new gif that shows a better view of the issue:
![rnn-bug-opening-screen-plateculture-gif](https://cloud.githubusercontent.com/assets/1016441/20278239/32ec318c-aaa4-11e6-8dc4-217e28b8f960.gif)

To solve my issue i had to comment in RCCViewControler.m this:
`//Bug fix: in case there is a interactivePopGestureRecognizer, it prevents react-native from getting touch events on the left screen area that the gesture handles
    //overriding the delegate of the gesture prevents this from happening while keeping the gesture intact (another option was to disable it completely by demand)
    self.originalInteractivePopGestureDelegate = nil;
    if (self.navigationController != nil && self.navigationController.interactivePopGestureRecognizer != nil)
    {
      id <UIGestureRecognizerDelegate> interactivePopGestureRecognizer = self.navigationController.interactivePopGestureRecognizer.delegate;
      if (interactivePopGestureRecognizer != nil)
      {
        self.originalInteractivePopGestureDelegate = interactivePopGestureRecognizer;
        self.navigationController.interactivePopGestureRecognizer.delegate = self;
      }
    }
`

[...]

PR #872, opened 2017-03-18
iOS - add disabledBackGesture

gran33:
This new style will NOT pass between push screens.

### DIFF
--- a/ios/RCCNavigationController.m
+++ b/ios/RCCNavigationController.m
@@ -88,6 +88,7 @@ NSString const *CALLBACK_ASSOCIATED_ID = @"RCCNavigationController.CALLBACK_ASSO
       [mergedStyle removeObjectForKey:@"statusBarHideWithNavBar"];
       [mergedStyle removeObjectForKey:@"autoAdjustScrollViewInsets"];
       [mergedStyle removeObjectForKey:@"statusBarTextColorSchemeSingleScreen"];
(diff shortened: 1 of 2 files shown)


grundmanise:
So we possibly can use this new param to disable back gesture on all root screens and prevent https://github.com/wix/react-native-navigation/issues/428 ?

gran33:
As far as I see, the real problem in #428 is the fact that the cell isn't push any more and the swipe is just weird thing that happened. So, `disabledBackGesture` will just prevent the swipe gesture, no more then this. If it solves any other bugs, use it.

grundmanise:
This in not the issue with the cell, it happens when you swipe from left to right on the root screen. https://github.com/wix/react-native-navigation/issues/645 <- same issue. Thanks for the commit, it is a useful feature.

grundmanise:
@gran33 only thing is that we should document this new param.

gran33:
@grundmanise thanks for the reminder.
https://github.com/wix/react-native-navigation/wiki/Styling-the-navigator#style-object-format

kl:
I don't think you should edit the wiki until the patch is merged. Now the ```disabledBackGesture``` property is documented but it has no effect, which is very confusing. Any chance of getting this merged into master?

nbonamy:
@kl agreed. was struggling with it until I found your comment!

PR #910, opened 2017-03-22
Feature/button items

simonmitchell:
This adds support (iOS only) for:
- Styling navigation items independently for each side of the navigation
- Showing both a title and image on iOS in navigation items
- Styling navigation items like any button (Border radius, border width, background color e.t.c.)


simonmitchell:
IGNORE THIS, realised it's got a lot of un-needed stuff in

PR #937, opened 2017-03-26
Add ability to set attributes of navigation bar buttons separately.

urbancvek:
This will enable to set style attributes to navigation bar buttons separately. It enables you to do stuff like this.

screen shot 2017-03-27 at 01 04 30

So you can set anything you could before (color, font size, font weight) this just add the ability to set them separately if you need to. It also doesn't break previous API.

So how would you use it? Before you could do something like this.
```js
class Screen {
  static navigatorStyle = {
    navBarButtonFontSize: 17,
    navBarButtonColor: 'red',
    navBarButtonFontWeight: '400',
  };
}
```

Now you can just write styles like this.
```js
class Screen {
  static navigatorStyle = {
    navBarLeftButtonFontSize: 17,
    navBarLeftButtonColor: 'red',
    navBarLeftButtonFontWeight: '400',

    navBarRightButtonFontSize: 17,
    navBarRightButtonFontWeight: '600',
    navBarRightButtonColor: 'blue',
  };
}
```

Also where could I update the documentation? Because even for the API that's used right now I couldn't find the docs anywhere and stumbled on it in the code when I was trying to implement it. Would like to help and add some if you would point me in the direction of the docs.


urbancvek:
I've now also added a few lines of code into the `setButtons` method which allows you to change styles of a button when calling it. I've set the prefix to be only "button" so it's shorter.

chuckhacker:
Has this been merged? Would love to see this fix in here.

chadharrington:
+1. We would really benefit from this fix. Thanks for your work on this great project.

guyca:
Hey @urbancvek 
Thanks for the PR. What properties are supported by this pr? Would love to see these wonderful additions on Android as well. Do you have plans for Android support?

urbancvek:
@guyca Every property that was supported before for setting the buttons. So that would mean every property that is extracted by:
```
NSMutableDictionary *leftNavButtonTextAttributes = [RCTHelpers textAttributesFromDictionary:self.navigatorStyle withPrefix:@"navBarLeftButton"];
```
You can see it on line 277 in `ios/RCCViewController.m`. I've just reused it so it also looks for `navBarLeftButton` and `navBarRightButton` prefixes.

In the `setButtons` function I've just added the ability to also set button styles, again by just reusing the `NSMutableDictionary........` stuff.

As goes for android support. Are you able to style buttons now just not separately? (Haven't worked on android part of the app and don't know the android API for react-native-navigation) If yes this should be very easy to implement.

guyca:
@urbancvek Let's merge this pr asap and add Android support later. TBH I don't think this feature will be needed that much on Android since it's really not common to have different style buttons on Android.

So, the only thing missing is properly documenting this great addition [here](https://wix.github.io/react-native-navigation/#/adding-buttons-to-the-navigator). Learning from past mistakes, not documenting features right away causes great confusion and frustration later. It will also help us carry this change to the next major version which is in the works now.
This is the file that needs to be updated: https://github.com/wix/react-native-navigation/blob/master/docs/adding-buttons-to-the-navigator.md

urbancvek:
@guyca Just added the needed docs. I've also added fontSize and fontWeight to global settings since it's supported just not documented. Thanks for pointing me in the direction of the docs. This is a recent addition right? Because I remember trying to do it last month and couldn't find them.

vtsatskin:
@urbancvek Thanks for making this PR! I noticed that setting the `buttonFontWeight` attribute causes the font size to become smaller than the default on iOS. 

# With buttonFontWeight Set

screen shot 2017-05-30 at 5 02 30 pm

```
class MyComponent extends Component {
  static navigatorButtons = {
    leftButtons: [
      {
        title: 'Cancel',
        id: 'cancel',
      },
    ],
    rightButtons: [
      {
        title: 'Save',
        id: 'save',
        buttonFontWeight: '600',
      },
    ]
  }

  // ...

}
```

# Without buttonFontWeight Set
screen shot 2017-05-30 at 5 04 03 pm

```
class MyComponent extends Component {
  static navigatorButtons = {
    leftButtons: [
      {
        title: 'Cancel',
        id: 'cancel',
      },
    ],
    rightButtons: [
      {
        title: 'Save',
        id: 'save',
      },
    ]
  }

  // ...

}
```

urbancvek:
Hey @vtsatskin this is happening because if you don't set any styles for buttons, default style is created by `react-native-navigation`. If you do provide some styles they are used and the defaults are just ignored. Might be a good thing if defaults styles were just merged with provided ones in the way that default styles get overwritten by provided styles.

Right now to get around this just also set `fontSize` whenever you provide `fontWeight` or anything else. I think the default is 15 or 16 just try both and see which one works. Hope this helps.

vtsatskin:
@urbancvek Thanks for the quick reply! I've used the workaround you've suggested and it worked for me. The current behaviour isn't entirely intuitive, so I agree that merging would be an understandable behaviour. Moving forward, would it make more sense to document the current behaviour or have the merging behaviour implemented?

PR #955, opened 2017-03-30
Added 'bottomTabPressed' navigation event

AidenMontgomery:
Added an additional navigation event, 'bottomTabPressed', for the press of an already selected tab.


guyca:
Thanks for the PR @AidenMontgomery 
Perhaps rename the event name to `BottomTabReselected` ? It might be a bit unclear what the difference between `tabPress` and `tabSelected` is.

PR #981, opened 2017-04-05
Easier to understand lede

aarongreenwald:
I saw some comments online in discussions about RNN that readers didn't understand what RNN was about. This change makes it clearer and removes the technical note about react-native-controllers that isn't really interesting to most people.

P.S. _lede_ is a [word](https://www.merriam-webster.com/dictionary/lede), not a spelling mistake. In case you were wondering.


guyca:
I ❤ U

PR #983, opened 2017-04-05
fix push while transitioning

gran33:


henrikra:
More details what does this fix?

gran33:
Sure @henrikra, we had a crash when the JS perform `push` one after another without waiting to the first one to complete the push. iOS UIKit allow that, but if u try to push another screen while the first push is happening the app might crash.

So, my fix is queuing the push if the push is happening and another push is perform. E.i, I add to an array the push that cannot be perform, and when the relevant `UINavigationController` is ready to perform the push, in the next `runloop` I perform the push.

If u have any other questions, feel free to ask here.

PR #999, opened 2017-04-07
Add tabBarTranslucent style option in tabsStyle

kiddo13:
When we change our splash screen background color from white to other color (e.g red), the tabBar background color also become red. We expect the tabBar background color to be white.

We try to set the tabBarBackgroundColor to white, but the tabBar has translucent as default. So, we could not get the expected color.

Using this option, we can turn the translucent on or off from the tabsStyle option.


guyca:
@kiddo13 Thanks for the pr. Can you please add it to the [docs](https://wix.github.io/react-native-navigation/#/styling-the-tab-bar) as well?

kiddo13:
sure. will do it soon.

guyca:
@pie6k It's in the [docs](https://wix.github.io/react-native-navigation/#/styling-the-tab-bar?id=style-object-format)
Indeed some properties are missing, if you can add missing props that would be really appreciated, Thanks 👍

Issue #571, opened 2016-12-15
StatusBar translucent causing TopBar cut off on Android


messense:
### Issue Description

Using `StatusBar` component to set `translucent` causing TopBar(I don't know if it is called `TopBar`) cut off on Android

### Steps to Reproduce / Code Snippets / Screenshots

![image](https://cloud.githubusercontent.com/assets/1556054/21211525/3e7ba2b6-c2bf-11e6-83bd-4168ea05f7b0.png)

---
### Environment
* React Native Navigation version: 2.0
* React Native version: 0.39
* Platform(s) (iOS, Android, or both?): Android
* Device info (Simulator/Device? OS version? Debug/Release?): Simulator Android 7.0 Debug

guyca:
Thanks for reporting. If you only want to control the color of the status bar - you can set `statusBarColor: 'red'` in `navigatorStyle or `AppStyle`

bluetec:
I have same issue with Experimental-300.
```navBarBackgroundColor: '#00A387',
        navBarTextColor: '#FFFFFF',
        navBarButtonColor: '#FFFFFF',
        navBarTranslucent: false,
        topBarElevationShadowEnabled: false,
        navBarTextFontFamily: 'IRANYekan',
        navBarTitleTextCentered: true,
        statusBarColor: '#00997d',```

navBarTranslucent is disabled, but TopBar is same as @messense screenshot.
How to fix this?

guyca:
@bluetec Are you using [immersive mode](https://developer.android.com/training/system-ui/immersive.html)? rnn renders only below the status bar, please check if you have one of the immersive mode flags defined.

bluetec:
@guyca thanks, a @Shoutem/Ui StatusBar component cause the issue, That's my fault. 👍

guyca:
RNN is compatible with the status bar component, You can use `statusBarColor` instead. Hope that help.

dev-two:
Actually, it would be great if you add support for translucent statusbar. It's actually very important part of material design, for example. Most of apps are using translucent status bar.

jurajkrivda:
@dev-two you can use StatusBar from RN, `<StatusBar barStyle="light-content" backgroundColor="rgba(0, 0, 0, 0.60)" translucent />`

xvonabur:
[...]

PR #1004, opened 2017-04-07
Add statusBarColor to navigator props docs

Ehesp:
Reading the docs I couldn't see a way to simply change the color of the status bar on Android. Only via https://github.com/wix/react-native-navigation/issues/571#issuecomment-267255350 I saw there was a prop to do this.

(Assuming this is Android only since iOS only allows for light or dark).


guyca:
Thanks 👍

PR #1030, opened 2017-04-11
Support custom font in the Menu items.

ruiaraujo:


android/app/src/main/java/com/reactnativenavigation/views/TitleBarButton.java
```diff
@@ -38,6 +40,16 @@ MenuItem addToMenu(int index) {
         return item;
     }
 
+    private MenuItem createMenuItem(int index) {
+        if (buttonParams.font == null) {
```
ruiaraujo: I was keeping the style of the line 85 without thinking much. I will change it.

Issue #828, opened 2017-03-07
Navbar white flash when pushing


danieljvdm:
I'm trying to push a view with the following navbar style:
```
  static navigatorStyle = {
    navBarBackgroundColor: '#05668D',
    navBarButtonColor: 'white',
    navBarTransparent: true,
    drawUnderNavBar: true,
    navBarNoBorder: true,
    statusBarHidden: true,
    screenBackgroundColor: '#05668D'
  }
```

The result is the following. The primary view has a `navBarHidden: true` setting, and if I don't hide it the entire navbar flashes white during the push. Bare in mind this only shows for about 1s then comes right.

![white-navbar](https://cloud.githubusercontent.com/assets/3450486/23649319/5d892cc4-031e-11e7-9196-d2416b944bf0.png)

---
### Environment
* React Native Navigation version: next (236)
* React Native version: 0.42
* Platform(s) (iOS, Android, or both?): iOS
* Device info (Simulator/Device? OS version? Debug/Release?): Simulator 10.2

pqkluan:
Hi ,

```

Issue #757, opened 2017-02-13
White flash after splash and before app render


tvansteelandt:
### Issue Description

I have a white flash when booting the app, after the splash screen hides and the RNN single screen app shows.
I've read a lot on the RN repo, and here on the RNN repo, but I'm not sure what's causing this.
Is it react native itself, or is it a problem with react native navigation?

I've stumbled upon this:
https://peterminarik.tumblr.com/post/153039209421/how-to-fix-the-initial-white-flash-in-your-react
But I don't see a way to implement this using the bootstrap from RNN?

### Steps to Reproduce / Code Snippets / Screenshots

Use standard splash screen image and boot singleApp or tabBasedApp.

---
### Environment
* React Native Navigation version: 2.0.0-experimental.187
* React Native version: 0.37
* Platform(s) (iOS, Android, or both?): iOS
* Device info (Simulator/Device? OS version? Debug/Release?): iPhone5S device

JuneDomingo:
Hi @tvansteelandt, when using actual device, try to change `jsCodeLocation` value from ios/AppDelegate.m to
`jsCodeLocation = [[RCTBundleURLProvider sharedSettings] jsBundleURLForBundleRoot:@"index.ios" fallbackResource:nil];`

guyca:
@tvansteelandt https://github.com/wix/react-native-navigation/pull/437 might help

PR #1046, opened 2017-04-14
Fix flash issue, Added ability to set secondary badge

shahen94:
Closes out #358 #828 #757 

I've debounce the push method to make sure that the react views will be rendered until they will be visible

Also added ability to add secondary badge for every tab (iOS only)


juancaacuna:
We need this! Thanks!!

guyca:
@shahen94 @gevorg94 As discussed on discord i'm closing the pr.
The white flash on iOS should be addressed but adding a timeout is a bit problematic, support for custom BottomTabs will probably be included in 2.x. Thanks a lot guys 👍

juancaacuna:
@guyca Yes, the white flash on iOS is urgent. Thank you!

Mark2xv:
Excuse me for the necro.
Might I ask why this was closed? Not everyone is involved on the discord channel. At the very least, could you state the reasons why this was rejected?

The white flashes are quite an issue as far as I'm concerned. And if I'm completely honest, a  workaround with a solid background image or color isn't really a solution either. Why not just wait until there has been a render on the screen? 

This issue has been closed for 9 months now stating the solution would come in v2. v2 is not done yet, as far as I'm aware. Thus you have rejected a potential solution to a big issue while not coming up with something better for three quarters of a year. Why?

Excuse me if this sounds hostile, that is not my intent. I'm just trying to figure out *why*.

guptarohit:
This is still a issue.

guyca:
Hey @Mark2xv 
To avoid the while flash, this pr introduced a hardcoded delay when pushing screen. While this will effectively eliminate the white flash, it's not an acceptable solution.

This issue is resolved for both platforms on v2. The reason we're not resolving this in v1 is simply because we don't want to invest any resources in v1 so we can transition as soon as possible to v2.

If someone would like to PR this for v1, the propper way is to use [RCTContentDidAppearNotification](https://github.com/facebook/react-native/blob/26684cf3adf4094eb6c405d345a75bf8c7c0bf88/React/Base/RCTRootView.m#L35)

PR #1054, opened 2017-04-17
Deep Link doco correction

arggrande:
Raising this PR to move the contents from `deep-links.md` inline with `screen-api.md`, as the hyperlink used in `handleDeepLink` was attempting to link to the Deep Link information on the same page.

I wasnt entirely sure how to change the hyperlink to reference `deep-links.md` directly, so I decided to inline it instead.

Happy for feedback if theres a better/alternative way of doing it.


guyca:
Hey @arggrande 
Thanks for the pr. The docs were neglected for a while an it shows, but we would like to make sections more visible and actually break some of the long pages into smaller ones - so I'm not sure merging both pages is a good idea.
Where is the problematic link in screen-api.md?

arggrande:
Hi @guyca yeah completely understandable :)

The issue is in the section: `handleDeepLink(params = {})` of `screen-api.md`.

Would you like me to rework it to try and use the correct file (though, as I previously mentioned, I'm not entirely sure how to go about this within markdown and referencing external files)?

Happy to help in any way i can.

PR #1128, opened 2017-04-25
Add possibility to pass payload through handleDeepLink

eeynard:


DanielZlotin:
@guyca your call. deeplinks (which is bad naming. this is a simple broadcast mechanism) will be removed in the next major version. This has nothing to do with navigation, we can export this to a different library if we really want

eeynard:
To clarify my need, I need to push a screen from a modal that had been open from the drawer. So my first try was something like that :
```javascript
// In the modal
Navigator.dismissModal()
this.props.navigator.toggleDrawer({ side: 'left', animated: false, to: 'closed' });
this.props.navigator.push(screen);
```
But it appeared that this code was not valid. So I read about "deep links" and I finally achieve to do what I wanted to do :
```javascript
// In the modal
Navigator.dismissModal()
this.props.modal.handleDeepLink({ link: 'myscreen' });

// In my home screen
onNavigatorEvent = (event) => {
  if (event.link === 'myscreen') {
    this.props.modal.toggleDrawer({ side: 'left', animated: false, to: 'closed' });
    this.props.navigator.push(screen);
  }
}
```
The problem is that I need to pass specific props from the modal to the **push** instruction. Is there any other clean way to achieve my goal ?

Thanks !

guyca:
@DanielZlotin Can you clarify how this use case is handled in v2? Thanks 👍

DanielZlotin:
easy, with the new architecture all commands are screen-specific:

```javascript
Navigation.dismissModal(modal-id);  /* or */
 Navigation.on(modal-id).dismissModal();
Navigation.on(sideMenuScreenId).hideSideMenu({animated: false}) //
temporary naming
Navigation.on(screenA-id).push(screenBWithPassProps);
```

The naming is temporary and subject to change, but in any case the goal is
to be able to send imperative commands to Navigation by screen id (string
which will be injected as a prop to root components) from anywhere.

I'm also playing with the idea to dynamically pass props from anywhere by
screen id, something like:

```javascript
Navigation.on(screenId).passProps({...});
```
But not sure this is a good idea

DanielZlotin:
easy, with the new architecture all commands are screen-specific:

```javascript
Navigation.dismissModal(modal-id);  /* or */ Navigation.on(modal-id).dismissModal();
Navigation.on(sideMenuScreenId).hideSideMenu({animated: false}) //temporary naming
Navigation.on(screenA-id).push(screenBWithPassProps);
```

The naming is temporary and subject to change, but in any case the goal is
to be able to send imperative commands to Navigation by screen id (string
which will be injected as a prop to root components) from anywhere.

I'm also playing with the idea to dynamically pass props from anywhere by
screen id, something like:

```javascript
Navigation.on(screenId).passProps({...});
```
But not sure this is a good idea

guyca:
Thanks for the pr @eeynard 👍 Merged it for the current scope but, as @DanielZlotin mentioned, the deepLinks feature it likely to be removed in the next major version.
A high order component can be used to help keep track of screen visibility, and registering listeners for specific events.

PR #1130, opened 2017-04-25
SET fixes

Kelt:
fixes for SET

### DIFF
--- a/android/app/src/main/java/com/reactnativenavigation/utils/ViewUtils.java
+++ b/android/app/src/main/java/com/reactnativenavigation/utils/ViewUtils.java
@@ -185,7 +185,7 @@ public class ViewUtils {
     }
 
     public static ForegroundColorSpan[] getForegroundColorSpans(TextView view) {
-        SpannedString text = (SpannedString) view.getText();
+        SpannedString text = new SpannedString(view.getText());
         return text.getSpans(0, text.length(), ForegroundColorSpan.class);
     }
 
--- a/android/app/src/main/java/com/reactnativenavigation/views/sharedElementTransition/SharedElementTransition.java
+++ b/android/app/src/main/java/com/reactnativenavigation/views/sharedElementTransition/SharedElementTransition.java
@@ -80,7 +80,7 @@ public class SharedElementTransition extends FrameLayout {
             @Override
             public void run() {
                 if (spannableText == null) {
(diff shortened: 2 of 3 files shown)


Kelt:
@guyca @enterline

android/app/src/main/java/com/reactnativenavigation/utils/ViewUtils.java
```diff
@@ -185,7 +185,7 @@ private static int getStatusBarHeight() {
     }
 
     public static ForegroundColorSpan[] getForegroundColorSpans(TextView view) {
-        SpannedString text = (SpannedString) view.getText();
+        SpannedString text = new SpannedString(view.getText());
```
guyca: What's the motivation behind this change?

android/app/src/main/java/com/reactnativenavigation/views/sharedElementTransition/SharedElementTransition.java
```diff
@@ -80,7 +80,7 @@ private void saveTextViewSpannedText(final TextView view) {
             @Override
             public void run() {
                 if (spannableText == null) {
-                    spannedText = (SpannedString) view.getText();
+                    spannedText = new SpannedString(view.getText());
```
guyca: Why create a new SpannedString? Is there a performance hit from creating a new instance of SpannedString?

Kelt: finally was able to repro ... so, if i wrap TextInput in SharedElementTransition, i get 

java.lang.ClassCastException: android.text.SpannableStringBuilder cannot be cast to android.text.SpannedString at com.reactnativenavigation.views.sharedElementTransition.SharedElementTransition$2.run(SharedElementTransition.java:80)

since view is com.facebook.react.views.textinput.ReactEditText and view.getText() returns SpannableStringBuilder

This however doesnt happen if I wrap TextInput in transparent View

guyca: The exception isn't thrown when you wrap the Text in a transparent view since in that case We regard the shared element as a simple view - which doesn't have a text color property so this entire logic isn't executed.

I wonder under which condition `getText()` returns a builder and when does it return the concrete SpannedString. Can you confirm this has no performance hit?

PR #1159, opened 2017-04-27
Render children and pass props to children

guyca:


src/views/sharedElementTransition.ios.js
```diff
@@ -1,11 +1,18 @@
...
 
 export default class SharedElementTransition extends Component {
+  static propTypes = {
+    children: PropTypes.object
+  };
+
   render() {
-    return <View />;
+    return (
+      <View {...this.props}>
```
shahen94: Would be good to omit `children` from props

PR #1163, opened 2017-04-28
Update sharedElementTransition.ios.js

tolu360:
Fix missing import of PropTypes throwing an error on iOS

### DIFF
--- a/src/views/sharedElementTransition.ios.js
+++ b/src/views/sharedElementTransition.ios.js
@@ -1,4 +1,4 @@
-import React, {Component} from 'react';
+import React, {Component, PropTypes} from 'react';
 import {
   View
 } from 'react-native';


guyca:
Thanks @tolu360

zwenza:
this will not work with react 16.0 because the prop-types package was moved out of react

tolu360:
Hi @zwenza, was the move done after `16.0.0-alpha.6`?

zwenza:
i don't exactly know if the alpha6 already removed it but the final 16.0 will remove it as far as i know

tolu360:
Still works in alpha 6, moving forward, what would be the right way to use PropTypes?

zwenza:
okay then it's fine. The right way would be to add the new prop-types package (https://www.npmjs.com/package/prop-types) and use it with `import PropTypes from 'prop-types'` and then use it just like before.

tolu360:
Alright, noted.

PR #1164, opened 2017-04-28
changed centering logic of title on android

Kelt:
when screen uses slide in animation (so screen starts 'off screen'), getLocationOnScreen is not helpful to center title, ends up being of the screen ... Using absolute values , screen width solves this problem ... @guyca


guyca:
Nice catch!

PR #1165, opened 2017-04-28
Support launch storyboards as splash screens on iOS

iotize:
[Apple's docs](https://developer.apple.com/ios/human-interface-guidelines/graphics/launch-screen/) recommend the use of storyboards as launch screens:

> Because device screen sizes vary, launch screen sizes vary too. To accommodate this, you can provide a launch screen as an Xcode storyboard or as a set of static images for the devices your app supports. Using an Xcode storyboard is the recommended approach, as storyboards are flexible and adaptable. You can use a single storyboard to manage all of your launch screens. To learn about implementing adaptable interfaces, see Auto Layout Guide.

This PR changes `-[RCCManager showSplashScreen]` so that it checks first for a launch storyboard, then for a launch xib, and then for launch images. If none of these are found, it defaults to showing a view controller with a white background.

* The code used for looking for a launch image is unchanged.
* The xib code is made slightly safer by checking that the class of the xib first object is a `UIView`.
* Class methods are used for all of these things as none of them need to refer to `self`.
* Grouped Splash Screen related methods under a `#pragma mark` and moved `getBridge` and `getAppWindow` to the top of the file.

In my tests, this works beautifully. I would be interested to know if it still works correctly for images and xibs.

Any feedback much appreciated.


tmaly1980:
Does this have anything to do with why I'm getting a blank white screen with my fatal errors? Seems like something to do with storyboard files.

https://stackoverflow.com/questions/46206139/error-screen-with-white-background

yogevbd:
Closes #2168

PR #1167, opened 2017-04-29
navigator.setSubTitle() support for iOS

kristoff-it:
# Problem
`navigator.setSubTitle()` is implemented for Android but not for iOS. 
`navigator.setTitle({subtitle})` allows to change the subtitle of a navigator but omitting the title property will result in it being deleted. This PR brings feature parity between iOS and Android.

# Resolution
As I have **never** done any native work for iOS, I started by replicating every piece of code that involved `setTitle` to get a better understanding of the data flow. This approach worked but it felt like I added too much redundancy. I opted instead for piggybacking the internals of `setSubTitle` on `setTitle` by injecting `isSetSubtitle` into the props passed by the user to differentiate wether the title should be kept as-is or not.
I don't know if it's a good idea to keep both methods so close together but on the other hand I have no idea where a good split might be and the two functions are already kind of intertwined (`setTitle({subtitle})`, for example). Additionally I both don't know wether any of this code will survive to v2 and, as a user, I feel like the `setTitle`/`setSubTitle` API should be reworked a little to make it more affordable, especially in regard of keeping a property as-is vs deleting it. 

As an example of how I think it could be done:
```
navigator.setTitle({title: '0 New'}) // changes the title
navigator.setTitle({subtitle: 'Connecting'}) // changes the subtitle and leaves the title as it was before
navigator.setTitle({title: '10 New', subtitle: null}) // changes the title AND deletes the subtitle
// same with all the other props that can be given to `setTitle()`: omitting leaves as-is, specifying to null deletes it
```

It doesn't seem too hard to implement in iOS, I think, but I have no idea about Android.
In any case feature parity between the two platforms seems a good thing.

PS
This is also my first PR to a project of this size, please be gentile 🍒  :)


khatv911:
hi @guyca : what about style for subtitle?

guyca:
@khatv911 You can set `titleBarSubtitleColor` in `navigatorStyle`
I think iOS supports a few more styling options, though I'm not sure if they are documented 😓

PR #1189, opened 2017-05-03
this.props.navigator is undefined

dabit1:


guyca:
Hey @dabit1 
Can you please explain the reason for this change? the navigator is injected to props, shouldn't be undefined.

dabit1:
Yes, sorry. The undefined is `this.props` because you are in constructor() that can't access to `props` via `this`. It has to access via `props` without `this`.

DanielZlotin:
probably not calling `super(props)` in the constructor

guyca:
@dabit1 props are injected in the constructor, if you're calling `super(props` they shouldn't be null.
I'm closing for now, but let me know if I misunderstood the issue or if the docs are not clear enough on this matter.

dabit1:
You are right.. sorry I was wrong.

Issue #1179, opened 2017-05-01
Custom buttons on Android does not work


dabit1:
Hi!

I'm changing buttons to custom buttons. I changed 'sideMenu' button with an icon from react-native-vector-icons and it is working well. But when I try to change de 'back' button does not work, always show the default left arrow.

This is the code:

```js
import MaterialIcon from 'react-native-vector-icons/MaterialIcons'

export default class Activity extends Component {

  static navigatorStyle = {
    navBarTextFontFamily: 'Raleway-Bold',
    navBarButtonColor: COLORS.overFirst
  }

  constructor(props) {
    super(props)

    MaterialIcon.getImageSource('arrow-back', 35, COLORS.overFirst).then(icon => {
      props.navigator.setButtons({
        leftButtons: [{
          icon,
          id: 'back'
        }]
      })
    })
  }

  //...
}
```

React native navigation version: 1.1.45
React native version: 0.43.3
Platfrom: Android
Device: Simulator

EDIT: I've noticed that also occurs with 'sideMenu', basically it seems that the problem is when I try to change button icon from a no-root screen.

a289459798:
i same

guyca:
Hey there,
`back` is reserved on Android for the default back button - [read the docs for more information](https://wix.github.io/react-native-navigation/#/adding-buttons-to-the-navigator?id=android-left-button). You can use any other id if you'd like to use your own back button.

dabit1:
If I use any other id simply occurs that no button appears. I thought that on Android only I'm able to use 4 types of left buttons, I'm wrong? If I'm right then the issue should be open.

guyca:
@dabit1 Can you please confirm that you're not able to use a custom icon only when calling `setButtons` ? Does using a custom icon work with a static `navigatorButtons` object?

dabit1:
I just tested it and I can confirm that I'm not able to use a custom icon using `setButtons` neither `navigatorButtons`. I've noticed a rare behavior, if I use both `setButtons` and `navigatorButtons` then the button setted in `setButtons` appears correctly:

**This doesn't work:**
```js
import MaterialIcon from 'react-native-vector-icons/MaterialIcons'

export default class Activity extends Component {

  static navigatorStyle = {
    navBarTextFontFamily: 'Raleway-Bold',
    navBarButtonColor: COLORS.overFirst
  }

  constructor(props) {
    super(props)

[...]

PR #1203, opened 2017-05-05
Custom left button on android

dabit1:
It fix the problem about setButtons function that doesn't set a custom icon on Android. However, if the ID is not none of the default 4 ids doesn't work. It solves the issue that I created #1179


guyca:
Hey @dabit1 Thanks for the pr, I've addressed this and a few more issues I found in #1205

dabit1:
Thanks to you man!

PR #1205, opened 2017-05-06
Custom left button fixes

guyca:
* navBarButtonColor was not applied to custom initial left button
* Dynamic custom button didn’t work
* Setting default icon (like menu or back button) after a custom icon was used didn’t work


Timson020:
custom left button on android did't work
how to do ?

PR #1268, opened 2017-05-17
[iOS] Support for statusBarColor

jondanao:
This is the effect. Without `statusBarColor` on the left, with `statusBarColor` on the right.

artboard

The idea was to create a UIView with the same size as the status bar and add that view to the navigationBar.


Naoto-Ida:
I just wanted to mention that this does not work as intended when `navBarHidden: true` due to it being added as a subview of the `navigationBar`.

Issue #606, opened 2016-12-27
Custom navbar buttons


berdof:
By default I can place image button to navbar. Is there any way to use custom button? For example drawer icon + notifications counter http://take.ms/eMrtm

Issue #524, opened 2016-12-01
How to have text next to an icon on the nav bar.


SudoPlz:
Hey there.

I'm trying to have some text on the nav bar (as part of one of the icons) like so:
<img width="441" alt="screen shot 2016-12-01 at 18 00 35" src="https://cloud.githubusercontent.com/assets/3070948/20801007/1fd3753e-b7f0-11e6-9bcd-433c8ae7c76d.png">

This would totally be possible if I was able to use an icon font on the nav bar, but to my knowledge that's not possible.

Please tell me there's a way to do this.

Thank you.

Jaafar-abusair:
You can use a button with title only with out icon property, in this case you will have text button and set the disabled property to true, so user will not be able to press on it

SudoPlz:
That's a nice hack @Jaafar-abusair , thanks. 
The only problem is, I can't really control it's position.

Also @guyca suggested to use a transparent Navigation Bar, and render a regular `<Text>` beneath it, but I would like to avoid that solution as well.

I might do what you suggest until we're able to render custom elements on the nav bar.

perrosnk:
I am also interested in this! I would like to have an icon next to some text.

Jaafar-abusair:
For me I hid the bar and i built my own custom top menu

guyca:
@Jaafar-abusair solution is definitely preferable. You can also try adding the `showAsAction: 'withText'` property to that button but that will work only on Android.

SudoPlz:
I have implemented a Custom Nav bar in 1.* by using the `navBarCustomView` (Check the [docs](https://wix.github.io/react-native-navigation/#/styling-the-navigator?id=style-object-format))

PR #1274, opened 2017-05-19
Custom UIBarButtonItem View in NavigationBar

hnryjms:
See #606 #524 

Allowing custom views in UIBarButtonItems will allow much functionality with text/image combinations, badges with symbols and other React-based functionality embedded inside a single button of the navigation bar.

I built an alternative to `class Navigator()` for retaining button information in order to pass an `onPress` prop down to the internal component of a screen when it is being rendered, which allows buttons in the nav bar to propagate outward, and then down to their respective `navigator` view rendering the page. I'm open to some suggestions on how to improve this.

Thanks to @artald and @gran33 for the earlier commits (95e0958eeee4251464fe1ec49421e9e68dd49014  and 1320de5d3e2e33471938620bb29e4889dd0d6626) I referenced when building this.

### Remaining Tasks

- [ ] Add documentation on new properties
- [ ] TBD: Split `<Screen` and `<NavButtonScreen` compnoents https://github.com/wix/react-native-navigation/pull/1274#discussion_r125650407
- [ ] Merge `master` and resolve conflicts

### DIFF
--- a/ios/RCCNavigationController.m
+++ b/ios/RCCNavigationController.m
@@ -3,10 +3,12 @@
 #import "RCCManager.h"
 #import <React/RCTEventDispatcher.h>
 #import <React/RCTConvert.h>
+#import <React/RCTRootView.h>
 #import <objc/runtime.h>
 #import "RCCTitleViewHelper.h"
 #import "UIViewController+Rotation.h"
 #import "RCTHelpers.h"
+#import "RCCCustomButtonView.h"
 
 @implementation RCCNavigationController
 {
@@ -272,10 +274,27 @@ NSString const *CALLBACK_ASSOCIATED_ID = @"RCCNavigationController.CALLBACK_ASSO
     NSString *title = button[@"title"];
     UIImage *iconImage = nil;
     id icon = button[@"icon"];
+    NSString *customView = button[@"customView"];
     if (icon) iconImage = [RCTConvert UIImage:icon];
-    
+    NSString *buttonId = button[@"id"];
+    NSString *callbackId = button[@"onPress"];
+
     UIBarButtonItem *barButtonItem;
-    if (iconImage)
+    if (customView) {
+      RCTBridge *bridge = ((RCTRootView*)viewController.view).bridge;
+
+      NSNumber *viewWidth = button[@"customViewWidth"] ?: @60;
+      NSMutableDictionary *initialProps = [NSMutableDictionary dictionaryWithDictionary:button[@"customViewInitialProps"]];
+      initialProps[@"buttonID"] = buttonId;
+      initialProps[@"callbackID"] = callbackId;
+      RCTRootView *reactView = [[RCTRootView alloc] initWithBridge:bridge moduleName:customView initialProperties:initialProps];
+      CGRect buttonFrame = CGRectMake(0, 0, viewWidth.floatValue, self.navigationBar.bounds.size.height);
+      RCCCustomButtonView *buttonView = [[RCCCustomButtonView alloc] initWithFrame:buttonFrame subView:reactView];
+      reactView.backgroundColor = [UIColor clearColor];
+
+      barButtonItem = [[UIBarButtonItem alloc] initWithCustomView:buttonView];
+    }
+    else if (iconImage)
     {
       barButtonItem = [[UIBarButtonItem alloc] initWithImage:iconImage style:UIBarButtonItemStylePlain target:self action:@selector(onButtonPress:)];
     }
@@ -289,10 +308,9 @@ NSString const *CALLBACK_ASSOCIATED_ID = @"RCCNavigationController.CALLBACK_ASSO
       }
     }
     else continue;
-    objc_setAssociatedObject(barButtonItem, &CALLBACK_ASSOCIATED_KEY, button[@"onPress"], OBJC_ASSOCIATION_RETAIN_NONATOMIC);
+    objc_setAssociatedObject(barButtonItem, &CALLBACK_ASSOCIATED_KEY, callbackId, OBJC_ASSOCIATION_RETAIN_NONATOMIC);
     [barButtonItems addObject:barButtonItem];
-    
-    NSString *buttonId = button[@"id"];
+
     if (buttonId)
     {
       objc_setAssociatedObject(barButtonItem, &CALLBACK_ASSOCIATED_ID, buttonId, OBJC_ASSOCIATION_RETAIN_NONATOMIC);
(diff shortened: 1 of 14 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "react-native-navigation",
-  "version": "1.1.0",
+  "version": "1.1.76",
```
hnryjms: Didn't mean for this commit to get into my master branch.......

SudoPlz:
This looks sleek, is it going to be merged soon @DanielZlotin ?

ios/RCCNavigationController.m
```diff
@@ -272,10 +274,27 @@ -(void)setButtons:(NSArray*)buttons viewController:(UIViewController*)viewContro
...
-    
+    NSString *buttonId = button[@"id"];
+    NSString *callbackId = button[@"onPress"];
+
     UIBarButtonItem *barButtonItem;
-    if (iconImage)
+    if (customView) {
+      RCTBridge *bridge = ((RCTRootView*)viewController.view).bridge;
+
+      NSNumber *viewWidth = button[@"customViewWidth"] ?: @60;
```
eliperkins: What's the rationale around this hardcoded 60 width?

ios/RCCNavigationController.m
```diff
@@ -272,10 +274,27 @@ -(void)setButtons:(NSArray*)buttons viewController:(UIViewController*)viewContro
...
+      NSMutableDictionary *initialProps = [NSMutableDictionary dictionaryWithDictionary:button[@"customViewInitialProps"]];
+      initialProps[@"buttonID"] = buttonId;
+      initialProps[@"callbackID"] = callbackId;
+      RCTRootView *reactView = [[RCTRootView alloc] initWithBridge:bridge moduleName:customView initialProperties:initialProps];
+      CGRect buttonFrame = CGRectMake(0, 0, viewWidth.floatValue, self.navigationBar.bounds.size.height);
```
eliperkins: Not every button needs to be as tall as the nav bar. Should we be relying on the view to set its own size and layout the RCCCustomButtonView relative to that?

src/Navigation.js
```diff
@@ -48,7 +48,7 @@ function _registerComponentNoRedux(screenID, generator) {
 
       render() {
         return (
-          <InternalComponent testID={screenID} navigator={this.navigator} {...this.state.internalProps} />
+          <InternalComponent testID={screenID} onPress={this.button ? this.button.onPress.bind(this.button) : null} navigator={this.navigator} {...this.state.internalProps} />
```
eliperkins: So this is passing an onPress prop to every component? Should this only be happening for nav bar button components?

eliperkins: Additionally, binding in props passed on render is generally frowned upon, since it happens every render. What do you think about lifting this out to a method on the subclass of `Screen` that is returned?

eliperkins:
@DanielZlotin @guyca Is this functionality you would like to get into the core of the lib? From my experience, setting custom views in bar button items is something that is common enough place to get in. Additionally, this would allow consumers to set custom bar button items (like those using an initializer like `[UIBarButtonItem -initWithBarButtonSystemItem:target:action:]` to allow consumers to use iOS's default icons.

eliperkins:
Also, this would enable using things like `react-native-vector-icons` without workarounds to extract out the icons as bitmaps!

ios/RCCNavigationController.m
hnryjms: Not completely positive, but I think as I was building this, I found `60px` as a comfortable distance when using a single icon as the button, giving a good touchable area size.

ios/RCCNavigationController.m
hnryjms: I think `UIBarButtonItem` is always going to adjust itself to be `100%` of the bar, so the flexibility here would be using padding within your custom view directly instead of reading more props through the JS/ObjC barrier to determine that padding...this way it functions a little more closely with the system customView buttons.

src/Navigation.js
hnryjms: I like the subclass idea...but then it would expose a `registerComponent()` and `registerNavigationButtonComponent()` top-level function as external API. This still seems desirable to me, but I want to confirm before I head down that road.

ios/RCCNavigationController.m
eliperkins: I'm pretty 👎 on giving a specific view size here, rather than self sizing. I think I'd prefer to see custom views have their own width and height as derived by RN/Yoga.

ios/RCCNavigationController.m
eliperkins: Coming with iOS 11, navigation bars will change height depending on the `prefersLargeTitles`, meaning that we'd have to also listen for changes there and act accordingly as well. I'd prefer to see views have some arbitrary height as defined by the React component's styling and have Yoga size it and feed out a view of that size, then allow for iOS to place it in the navigation bar as needed. This will allow for consumers to create components of arbitrary heights rather than forcing one on them.

src/Screen.js
```diff
@@ -163,6 +163,16 @@ class Navigator {
   }
 }
 
+class Button {
```
eliperkins: The more I think about this, the more I wonder if this can be accomplished using `Touchable`s, instead of creating a view that gets its own `onPress` defined.

hnryjms:
I don't have a ton of time to continue maintaining this PR, so if anyone is interested in finishing this up and getting it merged into `master`, please let me know!

tmaly1980:
I personally tend to use react-native-vector-icons (with Font Awesome) to do most of my icons, rather than graphics (PNG). Not being able to have non-bitmap icons in the navbar is the one big feature I think RNN is missing. Sometimes I want to have a 'settings' etc icon and it looks much nicer as a graphic than text, and as a programmer, I'd rather use a font symbol (Font awesome) than mess with a graphic design tool.

eliperkins:
I can pick this up this week, as it's a blocker for my company!

I'll take over this PR.

eliperkins:
I ended up taking a different approach here, so I didn't want to push to this same PR and branch (thanks for adding me to the repo though, @hnryjms!).

I opened a new PR, #1641, that uses more built in classes from UIKit and React Native to achieve a similar experience.

guyca:
@hnryjms Thanks a lot for the PR, we truly appreciate the effort. We went on with #1641 so Im closing this one.
Once again, thanks for the contribution 👍

hnryjms:
Hurray, I'm glad the functionality made it in!

eliperkins:
Thanks for your work on getting this going! ❤️

PR #1306, opened 2017-05-30
Update example AppDelegate to pass launchOptions to the bridge

jbolter:
I ran into an issue trying to get push notifications to work on iOS when the user taps on a notification to launch the app from a "cold" state using react-native-navigation with react-native-notifications.  Without passing launchOptions to the bridge it is not possible to detect the push notification that launched the app.  The example AppDelegate.m and associated docs don't mention this scenario so this should help other users since I don't see a downside to passing launchOptions by default.

This fixes https://github.com/wix/react-native-notifications/issues/35


varungupta85:
Thanks a ton @jbolter. I have been struggling with this as well and amazed to see that such a simple change fixed it.

PR #1312, opened 2017-06-01
iOS: Added native-like back button in RCCNavigationController

schumi1331:
I've added a native-like looking back button for iOS in the RCCNavigationController. It looks nearly the same as the real back button, but allows user-defined handling, e.g. navigating in a WebView or updating its title.

To use it, simply set it in the navigatorButtons with the id `back`:

```
static navigatorButtons = {
  leftButtons: [
    {
      id: "back",
      title: "Back"
    }
};
```

How it's looking like in the Simulator:
simulator screen shot 01 06 2017 10 02 14

### DIFF
--- a/ios/RCCNavigationController.m
+++ b/ios/RCCNavigationController.m
@@ -264,20 +264,62 @@ NSString const *CALLBACK_ASSOCIATED_ID = @"RCCNavigationController.CALLBACK_ASSO
    }];
 }
 
+-(void)onBackButtonPress:(UIButton*)backButton
+{
+  NSString *callbackId = objc_getAssociatedObject(backButton, &CALLBACK_ASSOCIATED_KEY);
+  if (!callbackId) return;
+  [[[RCCManager sharedInstance] getBridge].eventDispatcher sendAppEventWithName:callbackId body:@
+   {
+     @"type": @"NavBarButtonPress",
+     @"id": @"back"
+   }];
+}
+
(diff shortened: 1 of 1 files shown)


schumi1331:
May I ask when there is a chance that this pull request is going to be added or checked from someone? I mean that functionality is also available in the Android part of the package, so it's not a big deal...

eliperkins:
I would recommend against this approach, as it goes against [Apple's recommended UI guidelines](https://developer.apple.com/ios/human-interface-guidelines/ui-bars/navigation-bars/) as to what the back button should be doing.

> Don’t include multisegment breadcrumb paths. The back button always performs a single action—returning to the previous screen. If you think people might get lost without the full path to the current screen, consider flattening your app’s hierarchy.

Additionally, by encoding the back button image into the framework, we now need to replicate system functionality, rather than embrace it.

I would recommend creating a custom component to handle this type of logic, similar to how Safari and `SFSafariViewController` place their navigation controls into a toolbar. This is Apple's recommended approach.

> TIP: Use a toolbar when navigation isn’t needed or you want multiple controls for managing content.

guyca:
Thanks @eliperkins.
@schumi1331 I appreciate the PR but, as @eliperkins noted, this is bad practice so I'm closing.

schumi1331:
Actually there is a situation where this makes sense which is when using a WebView which shall look like a native view. In that situation where the user shall think that he is working in a native view, that back button is to go back to the previous page...

PR #1313, opened 2017-06-01
[docs] Fix projectDir for android/settings.gradle

thedrumsknight:
The react-native-navigation package's source files are installed under node_modules/lib folder and give an error while building with the currently suggested docs for adding projectDir in android/settings.gradle


guyca:
Hey @thedrumsknight
The path in the docs is correct. Perhaps your project is setup differently than the common setup.

PR #1314, opened 2017-06-01
Change "tabs" to "bottomTabs"

Ehesp:


Ehesp:
@DanielZlotin Not sure what the failing build here is? Something up with Travis? Took 30 mins to get that far haha

PR #1324, opened 2017-06-04
Add screen events for side menus on Android

MattDavies:
The existing screen events don't appear to encompass the left and right side menus. This change adds those events in.


guyca:
Thanks @MattDavies !

PR #1326, opened 2017-06-04
Add a way to dynamically enable/disable a side menu for Android

### LINKED ISSUE #326: How disable drawer on current window?
### Issue Description

How disable drawer on current window?

---
### Environment
- React Native Navigation version: 2.0.0-experimental.120
- React Native version: 0.32.0
- Platform(s) (iOS, Android, or both?): both

MattDavies:
Fixes #326 for Android only - I looked at the props approach but I needed a bit more control than I could get with that.

Sidenote: I'll need to do similar for iOS, but that looks a bit tricker given the multiple different drawer implementations (and the first implementation I looked at didn't have a clean way to disable without removing the view completely).

### DIFF
--- a/android/app/src/main/java/com/reactnativenavigation/bridge/NavigationReactModule.java
+++ b/android/app/src/main/java/com/reactnativenavigation/bridge/NavigationReactModule.java
@@ -152,6 +152,11 @@ public class NavigationReactModule extends ReactContextBaseJavaModule {
     }
 
     @ReactMethod
+    public void setSideMenuEnabled(boolean enabled, String side) {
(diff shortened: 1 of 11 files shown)


guyca:
@MattDavies looks good. Can we document this method here https://wix.github.io/react-native-navigation/#/screen-api ? Otherwise this will get left out of the next version

MattDavies:
@guyca Done. Hadn't seen the new docs, they are looking amazing. Nice work!

guyca:
Merged, thanks!

Fenwil:
Thanks for the work @MattDavies and @guyca !! 

I will leave the link here: http://wix.github.io/react-native-navigation/#/screen-api?id=setdrawerenabledparams-android-only

PR #1334, opened 2017-06-06
Added screen prop viewPagerEnableScroll [Enable/Disable swipe between viewPager screens]

Ibrahim-GHub:
I added a screen parameter viewPagerEnableScroll that would take a boolean and allow me to choose whether viewPager's scroll should be enabled or disabled this would be very useful in Top Tab Bar as sometimes we just need to press on tabs and disable switching between tabs swiping the screen especially if we got a horizontal scroller inside these tabs.


guyca:
Hey @Ibrahim-GHub 
Thanks for the pr. I have two requests
1. rename to property to `TopTabsScrollEnabled`
2. Add it to the [documentation](https://wix.github.io/react-native-navigation/#/styling-the-navigator)

Ibrahim-GHub:
Done

guyca:
@Ibrahim-GHub Is `topTabsScrollEnabled` true by default? It should be (since thats the default naive behavior), and we should also mention that in the docs.

Issue #1302, opened 2017-05-29
How to access drawer?


dominicsavb:
I want to disable drawer on some screen? how to do it?

How to change drawer from screen?

Thank you

Fenwil:
Hi @dominicsavb you should try this on the screen you want to disable the drawer: [http://wix.github.io/react-native-navigation/#/screen-api?id=setdrawerenabledparams-android-only](http://wix.github.io/react-native-navigation/#/screen-api?id=setdrawerenabledparams-android-only)

Fenwil:
@guyca you should close this one

Issue #922, opened 2017-03-23
[Question] use disableOpenGesture selectively


tolu360:
### Issue Description

This is more of a question than an issue. Please, is it possible to disable open gesture for the `drawer` on selected screens and not app-wide using the `disableOpenGesture` flag? If possible, can I get a quick guide on how? Thanks

---
### Environment
* React Native Navigation version: 2.0.0-experimental.240
* React Native version: 0.42.0
* Platform(s) (iOS, Android, or both?): Both
* Device info (Simulator/Device? OS version? Debug/Release?): Simulator

enterline:
+1 on this. I would definitely like to know if this is possible.

kis25791:
@tolu360 @enterline  Currently I think we have to do some workaround. In my side, I have a workflow like SplashScreen => Login => Main App. If I registered sidemenu at beginning then in my splash screen and Login screen, I was able to open SideMenu by gesture. So I decided to **startSingleScreenApp** without **Sidemenu** at first and after done with Login, I **startSingleScreenApp** again and add **Sidemenu** as drawer.
This workaround will prevent Open Gesture in Splash screen and Login Screen, but you will lost page animation transition from Login => Main app. But I think it was acceptable till now.

tolu360:
Thanks @kis25791, I am doing the exact flow you described but I have a slightly different issue here - it seems to be peculiar to IOS, for my main app, I **startTabBasedApp** with a **Drawer** and all is fine. My issue is just the views in my app which have **Swipeable ListViews**, the drawer seem to interfere with those lists and they are un-swipeable unless I **disableOpenGesture** altogether, after which I can now swipe open my **Swipeable ListViews**.

kis25791:
@tolu360 I think if you enable Swipeable ListView and Drawer, it will cause a conflict gestures in app. Did you try with different directions for both? Such as Drawer will move from left to right and your Swipeable List View will move from right to left, etc.

tolu360:
@kis25791 yes, I did. I have the drawer on the left (swiping right to open) and the listview swipes to the left to open, yet I cannot get the swipeable listview to work properly with the drawer on IOS.

Fenwil:
@guyca do you have any thoughts on this? thanks!

dabit1:
Hi! Any news about it? In most cases, the app has login system and we need to disable drawer depends on the screen. Also there are cases where some screens should not open the drawer.

Thanks!

kristikristo:
[...]

Issue #631, opened 2017-01-02
How to disable pan gesture to open drawer menu?


hossamghareeb:
### Issue Description

I have integrated drawer menu successfully but I want to disable the pan gesture to open it. How can I do that?

### Steps to Reproduce / Code Snippets / Screenshots
---

Navigation.startSingleScreenApp({
 screen: {
   screen: 'screen.Splash',
   title: 'First Screen',
   navigatorStyle: {
     navBarBackgroundColor: '#3C274D',
     navBarTextColor: '#ffffff',
     navBarNoBorder: true,
     navBarSubtitleTextColor: '#ff0000',
     navBarButtonColor: '#ffffff',
     statusBarTextColorScheme: 'light'
   }
 }
 ,
 drawer: {
   left: {
     screen: 'screen.SideMenu'
   }
 }
});

### Environment
* React Native Navigation version: 0.37.0
* React Native version: 2.0.0
* Platform(s) (iOS, Android, or both?): Both
* Device info (Simulator/Device? OS version? Debug/Release?): iPhone 7 Simulator

guyca:
set `disableOpenGesture: 'false'` in your drawer declaration. See the docs for me info 👍

hossamghareeb:
Thanks, but is it possible to disable it only for specific screens? Do you have a link for library docs which has the whole APIs

kyle-ssg:
+1

terrysahaidak:
Hey @guyca, @hossamghareeb and @kyle-ssg, any solution for disabling drawer open gesture on specific screens?

thiennhank9:
I have a same question of @terrysahaidak . How can I disable drawer open gesture on specific screens.
I find to use Drawer under Stack like this

```
export const draw = DrawerNavigator(
    {
        ChatView: { screen: ChatView },
        ListViewDone: { screen: ListViewDone },
    },
    {
        headerMode: 'screen',
        drawerPosition: 'right',
        contentComponent: Side
    }
)

export const stack = StackNavigator(
    {
        Login: { screen: Login },
        draw: { screen: draw }, //Insert drawer navigation under stack here
    },
    {
        headerMode: 'screen',
    }
)
```

But when I open drawer on ChatView scene, it now show header on top, although I set header be null  in all components, even component I use to be rendered in Drawer like this
```

static navigationOptions = {
        header: null
    };

```
Hope any ideas! Tks u in advance guys!

thiennhank9:
[...]

Issue #540, opened 2016-12-06
Is it possible to have API for setting disableOpenGesture for each view


Jaafar-abusair:
### Issue Description

Is it possible to have API for setting disableOpenGesture for each view
for example, for one view I want to set the value is false since I have many Gesture events, for the others I would like to keep it true

---
### Environment
* React Native Navigation version: 2.exp.151
* React Native version: 0.38.1
* Platform(s) (iOS, Android, or both?): Both, iOS at least
* Device info (Simulator/Device? OS version? Debug/Release?): all

guyca:
Hey @Jaafar-abusair 
What do you mean by "many gesture events"? If, for instance, you have a feed screen with cards that you can swipe to delete, swiping the card will delete the card. But if you start the swipe gesture from the edge of the screen the drawer is opened. Is this not the behavior you're getting?

annelorraineuy:
@guyca I have a screen wherein I need to disable the gesture to open the drawer. It's a game screen that has its own game menu when swiped from the left. This is a use case wherein it would be useful to be able to disable drawer gesture on specific screen.

Jaafar-abusair:
Hi @annelorraineuy ,

For the last version ^1.1.63, side menu is set on each screen by the following code
`static navigatorButtons = {
    leftButtons: [
      {
        id: 'sideMenu'
      }
    ]
  };`

So just don't insert it on the disabled screen

annelorraineuy:
Hi @Jaafar-abusair I did not set any buttons on the disabled screen. I set the drawer left button on the home screen. Also, I totally hid the navbar on the disabled screen.

Jaafar-abusair:
Which version do you use?

annelorraineuy:
@Jaafar-abusair I am using version 1.1.79.

On the disabled screen, I only put this:
```
static navigatorStyle = {
        navBarHidden: true
    };
```

My side menu button is declared in home screen (disabled screen is a child of home):
```
RooIcon.getImageSource('hamburger', 24, Colors.orange).then(drawerIcon => {
            props.navigator.setButtons({
                leftButtons: [{
                    icon: drawerIcon,
                    id: 'sideMenu'
                }]
            })
        })
```

Is there anything else I should do?

Jaafar-abusair:
what you mean disabled screen is a child of home? do you mean they are components? how do you display them? are you use posh/showModal?

annelorraineuy:
[...]

PR #1347, opened 2017-06-11
Feature/set drawer enabled ios

kyle-ssg:
Missing iOS implementation for https://github.com/wix/react-native-navigation/pull/1326. 

Relevant to: 
https://github.com/wix/react-native-navigation/issues/631
https://github.com/wix/react-native-navigation/issues/540
https://github.com/wix/react-native-navigation/issues/922
https://github.com/wix/react-native-navigation/issues/326
https://github.com/wix/react-native-navigation/issues/1302

### DIFF
--- a/ios/RCCDrawerController/RCCDrawerController.m
+++ b/ios/RCCDrawerController/RCCDrawerController.m
@@ -24,17 +24,17 @@ UIViewController *rightViewController = nil;
 {
 
     self.drawerStyle = props[@"style"];
-
+    
     // center
     if ([children count] < 1) return nil;
     UIViewController *centerViewController = [RCCViewController controllerWithLayout:children[0] globalProps:globalProps bridge:bridge];
-
+    
     // left
     UIViewController *leftViewController = nil;
     NSString *componentLeft = props[@"componentLeft"];
     NSDictionary *passPropsLeft = props[@"passPropsLeft"];
     if (componentLeft) leftViewController = [[RCCViewController alloc] initWithComponent:componentLeft passProps:passPropsLeft navigatorStyle:nil globalProps:globalProps bridge:bridge];
-
+    
     // right
     UIViewController *rightViewController = nil;
     NSString *componentRight = props[@"componentRight"];
--- a/ios/RCCDrawerController/RCCDrawerController.m
+++ b/ios/RCCDrawerController/RCCDrawerController.m
@@ -13,25 +13,28 @@
 
 @synthesize drawerStyle = _drawerStyle;
 
+UIViewController *leftViewController = nil;
+UIViewController *rightViewController = nil;
+
 -(UIInterfaceOrientationMask)supportedInterfaceOrientations {
     return [self supportedControllerOrientations];
 }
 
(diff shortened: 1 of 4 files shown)


kyle-ssg:
A quick note on this, @MattDavies was indeed correct about the drawer implementation, there didn't seem to be a way to disable without removing the view controller completely and then re-adding it when enabling the drawer.

This means it doesn't cover the usecase where you'd want to disable gesture dynamically but keep a drawer button clickable. I don't think this is really a scenario worth worrying about - especially seeing as this solves a very common need to disable the drawer on a per screen basis.

guyca:
Thanks @kyle-ssg !

MattDavies:
Nice work @kyle-ssg!

@guyca If I understand this correctly TheSidebarController is a little out of sync with the MMDrawerController features now - what's the purpose of having both of them in there? Is that something you want to maintain going forward or does it make sense to remove one of them now?

guyca:
@DanielZlotin You're familiar with the iOS aspect more then I am. `MMDrawerController` was removed in v2, right?

ios/RCCDrawerController/RCCDrawerController.m
```diff
@@ -124,7 +127,18 @@ - (void)performAction:(NSString*)performAction actionParams:(NSDictionary*)actio
...
         return;
     }
-    
+
+    // setDrawerEnabled
+    if ([performAction isEqualToString:@"setDrawerEnabled"])
+    {
+        bool enabled = [actionParams[@"enabled"] boolValue];
+        if ([actionParams[@"side"] isEqualToString:@"left"]) {
+            [super setLeftDrawerViewController: enabled ? rightViewController : nil];
```
MattDavies: Typo here - should be leftViewController

ios/RCCDrawerController/RCCDrawerController.m
```diff
@@ -13,25 +13,28 @@ @implementation RCCDrawerController
...
     self.drawerStyle = props[@"style"];
-    
+
     // center
     if ([children count] < 1) return nil;
     UIViewController *centerViewController = [RCCViewController controllerWithLayout:children[0] globalProps:globalProps bridge:bridge];
-    
+
     // left
     UIViewController *leftViewController = nil;
```
MattDavies: Because leftViewController and rightViewController are defined locally here, the class variables never get set. When you go to actually set the status of one of the drawers, the controller is therefore always nil (that caused me quite some confusion, as MMDrawerController doesn't handle that situation nicely and permanently disables interaction on the UI).

ios/RCCDrawerController/RCCDrawerController.m
```diff
@@ -13,6 +13,9 @@ @implementation RCCDrawerController
 
 @synthesize drawerStyle = _drawerStyle;
 
+UIViewController *leftViewController = nil;
```
MattDavies: Because leftViewController and rightViewController are defined locally in initWithProps as well as at a class level here, the class variables never get set. When you go to actually set the status of one of the drawers, the controller is therefore always nil (that caused me quite some confusion, as MMDrawerController doesn't handle that situation nicely and permanently disables interaction on the UI).

PR #1367, opened 2017-06-15
[iOS] Added fade transition animation between screens

xandyn:
Fade transition animation for:

- push
- pop
- popToRoot
- resetTo

for example
```javascript
this.props.navigator.push({ screen: 'example.ExampleScreen', animationType: 'fade' });
```
if you not specifying `animationType`, animation will be default.


guyca:
Very nice! Can you please update the [docs](https://github.com/wix/react-native-navigation/blob/master/docs/screen-api.md) as well? Thanks

guyca:
cc @JackThomson2

Issue #930, opened 2017-03-25
IOS back gesture not working


maurovisitin:
After i updated RNN to 2.0.0-experimental.256, swipe gesture to pop screen on IOS is not working anymore. I tried to add "disabledBackGesture: false,"  (wich should be default to false anyway) but no luck. Any suggestion?

---
### Environment
* React Native Navigation version: 2.0.0-experimental.256
* React Native version: 0.43.0-rc.4
* Platform: IOS (both simulator/device and debug/release)

Kureev:
Seems it's related to https://github.com/wix/react-native-navigation/commit/03d380f8a9d2fba5d4baf6815201fcebdcfa18c9. If you add it back, it "works".

Although, if you push multiple times and then pop to the very first VC, you can't push anymore.

gran33:
@sexykodo Thanks for reporting, can u plz share your `static navigatorStyle = {...}` and your 
```
this.props.navigator.push({
...
navigatorStyle = {...}
...
})

sexykodo:
@gran33 
static navigatorStyle = {
    drawUnderNavBar: true,
    navBarTransparent: true,
    navBarTranslucent: true,
    navBarHidden: true,
    tabBarHidden: true,
    drawUnderTabBar: true,
    statusBarTextColorScheme: 'light',
    screenBackgroundColor: SituaColors.darkBg,
    navBarButtonColor: SituaColors.action,
  };

Kureev:
In my case I used defaults

guyca:
Hey guys,
I wasn't able to reproduce this in the example app. Tried clean install (RN 0.43.3, `latest` rnn), running on simulator. I disabled the `drawer` component and pushed a screen, pop gesture worked as expected.  Perhaps someone can upload an example where the issue reproduces?

Kureev:
Sure! I'll make an example this week and let you know! Thanks for keeping an eye on this, really appreciate it

Kureev:
Ok, I tried `304` and it worked like a charm! 
@sexykodo, can you try to upgrade to the latest version? Looks like it is fixed now 🙌

guyca:
@Kureev Switch to `latest` in package.json, .304 is a pretty old release :)

Kureev:
@guyca For some reasons, `latest` & `next` installs a `1.1.x` version 🤔

guyca:
In order not to break everyone, `next` points to `latest` for now, but it will be deprecated soon. 👍

[...]

PR #1391, opened 2017-06-20
[iOS] Enable back swipe gesture when a navBar is hidden. Resolves this issue: https://github.com/wix/react-native-navigation/issues/930

grundmanise:
added https://github.com/wix/react-native-navigation/commit/03d380f8a9d2fba5d4baf6815201fcebdcfa18c9 back and also applied this fix https://github.com/wix/react-native-navigation/issues/428#issuecomment-278268928 on top of it


messense:
Would be great to have this merged.

Issue #1357, opened 2017-06-13
toggleTabs to 'shown' places BottomTabs at top of screen


jstheoriginal:
### Issue Description

Calling 

```
this.props.navigator.toggleTabs({
    to: 'shown',
    animated: false,
});
```

after previously hiding the `BottomTabs` causes the `BottomTabs` bar to appear at the top of the screen instead of bottom.

### Steps to Reproduce / Code Snippets / Screenshots

This was not an issue using `v1.1.89`, but it is an issue using `v1.1.96`.

Steps:

1. In `componentWillMount`, add Keyboard event listeners for `keyboardDidShow` and `keyboardDidHide`.
1. When `keyboardDidShow` event occurs, `toggleTabs` to `hidden`.
1. When `keyboardDidHide` event occurs, `toggleTabs` to `shown`.

Note that the `BottomTabs` is at the top of the screen instead of where it should be on the bottom, as it was before hiding.

---
### Environment
* React Native Navigation version: v1.1.96
* React Native version: 0.45
* Platform(s) (iOS, Android, or both?): Android
* Device info (Simulator/Device? OS version? Debug/Release?): Simulator, Google Nexus 6, 5.0.0 API 21

guyca:
Hey,
I'm not able to reproduce this in the example app (although I didn't try with a keyboard event listener, I simply tried toggling tabs)
The example app has an `action` screen (second tab) with a `toggleTab` action - can you reproduce it there? perhaps add a text input and relevant boiler plate code and post the screen here. cheers

jstheoriginal:
Sounds good I'll give it a go later today.

jstheoriginal:
@guyca Just trying to run the example app in this repo to configure it like we have it to reproduce, but at launch on Android I get this:

```
E/ReactNativeJS( 2538): Something went wrong initializing the native ReactLocalization module.
E/ReactNativeJS( 2538): Please check your configuration.
E/ReactNativeJS( 2538): Did you run 'react-native link'?

E/ReactNativeJS( 2538): undefined is not an object (evaluating 'localization.language')
```

And running `react-native link` says `rnpm-install info Android module react-native-navigation is already linked`

I cloned, ran `yarn`, ran `react-native link`. Is there another step to get the example app to run?

Thanks!

guyca:
1. clone rnn
2. cd examaple
3. yarn
4. react-native run-android (Or better yet, run from Android Studio)

no need to link, everything is already configured.

[...]

PR #1403, opened 2017-06-22
Fixed bottom tab bar animation setting the wrong value

BrendonSled:
Closes issue #1357 

setY should not be set on a view inside a RelativeLayout, this was setting the tab bar to the Y position of 0


jstheoriginal:
This should be the fix for https://github.com/wix/react-native-navigation/issues/1357

guyca:
Thanks @BrendonSled

PR #1404, opened 2017-06-22
Fixed tab bar hidden flag not being passed from parent to child screen pushes

### LINKED ISSUE #1402: TabBarHidden not remembered across pushes in Android
### Issue Description

According to the [docs](https://github.com/wix/react-native-navigation/blob/4c85cfde25494df09641d2d4ad58c41653a91e03/docs/styling-the-navigator.md)
`tabBarHidden: false, // make the screen content hide the tab bar (remembered across pushes)`
tabBarHidden should be remembered across pushes, this is the case for iOS but not Android

push screen with `tabBarHidden: true` and then push another screen and the tab bar will be shown again

---
### Environment
* React Native Navigation version: 0.45.1
* React Native version: 1.1.112
* Platform(s) (iOS, Android, or both?): Android
* Device info (Simulator/Device? OS version? Debug/Release?): Galaxy S7, 7.0.0, Debug

BrendonSled:
Closes #1402

### DIFF
--- a/android/app/src/main/java/com/reactnativenavigation/animation/VisibilityAnimator.java
+++ b/android/app/src/main/java/com/reactnativenavigation/animation/VisibilityAnimator.java
@@ -1,5 +1,7 @@
 package com.reactnativenavigation.animation;
 
+import android.animation.Animator;
+import android.animation.AnimatorListenerAdapter;
 import android.animation.ObjectAnimator;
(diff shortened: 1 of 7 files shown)


guyca:
Styles are not supposed to be inherited. It was a bad design decision which will be fixed in v2,

BrendonSled:
Thanks @guyca

PR #1416, opened 2017-06-26
Added tapBackgroundToDismiss to LightBox screens

BrendonSled:
Added param tapBackgroundToDismiss to LightBox screens to allow tapping on background views to dismiss the LightBox screen

### DIFF
--- a/android/app/src/main/java/com/reactnativenavigation/views/LightBox.java
+++ b/android/app/src/main/java/com/reactnativenavigation/views/LightBox.java
@@ -54,6 +54,15 @@ public class LightBox extends Dialog implements DialogInterface.OnDismissListene
         lightBox.setBackgroundColor(params.backgroundColor.getColor());
         lightBox.addView(content, lp);
 
+        if (params.tapBackgroundToDismiss) {
+            lightBox.setOnClickListener(new View.OnClickListener() {
+                @Override
+                public void onClick(View v) {
+                    hide();
+                }
+            });
+        }
+
         content.setOnDisplayListener(new Screen.OnDisplayListener() {
             @Override
             public void onDisplay() {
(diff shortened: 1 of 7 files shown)


android/app/src/main/java/com/reactnativenavigation/views/LightBox.java
```diff
@@ -54,6 +54,15 @@ private void createContent(final Context context, LightBoxParams params) {
         lightBox.setBackgroundColor(params.backgroundColor.getColor());
         lightBox.addView(content, lp);
 
+        if (params.tapBackgroundToDismiss) {
+            lightBox.setOnClickListener(new View.OnClickListener() {
+                @Override
+                public void onClick(View v) {
+                    hide();
```
guyca: You better call `destroy()` which dismisses the lightBos (Dialog) and unmounts the react view.

guyca:
Amazing! Please see me comment regarding `hide()`

BrendonSled:
Destroy should be called, hide calls animatedHide and when the animation end destroy is called.

android/app/src/main/java/com/reactnativenavigation/views/LightBox.java
BrendonSled: `destroy()` is called after the animation finishes

guyca: I don't think so. Once the animation ends, the Dialog is dismissed and the onDismiss listener is called. How is `destroy` getting called?

BrendonSled: Both BottomTabsLayout/SingleScreenLayout both use the `.hide()` method.

If I use the `.destroy()` the dialog will close without an animation

guyca: You're right, `destroy` should be called in `onAnimationEnd`.

BrendonSled: ok I've made that change @guyca

guyca: Great!

PR #1421, opened 2017-06-27
Fixed BottomTabs height being calculated twice

BrendonSled:
BottomTabs height is currently being calculated twice in Android making the hide animation not work in all cases


guyca:
Amazing catch, thanks!

PR #1424, opened 2017-06-28
Fixed horizontal transition android

JackThomson2:
Also using material motion guidelines for duration and curves as
referenced here:
https://material.io/guidelines/motion/duration-easing.html#duration-easi
ng-natural-easing-curves


alzalabany:
is this ready yet?

JackThomson2:
That error is very odd as that isn't code I've added.

MrPeproni:
+1

PR #1427, opened 2017-06-28
Fixed TabBar from not adopting screen background color

BrendonSled:
[http://imgur.com/a/JTvwm](http://imgur.com/a/JTvwm)

### DIFF
--- a/android/app/src/main/java/com/reactnativenavigation/layouts/BottomTabsLayout.java
+++ b/android/app/src/main/java/com/reactnativenavigation/layouts/BottomTabsLayout.java
@@ -8,7 +8,6 @@ import android.support.v4.widget.DrawerLayout;
 import android.support.v7.app.AppCompatActivity;
 import android.view.View;
 import android.widget.RelativeLayout;
-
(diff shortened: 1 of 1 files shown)


guyca:
Hey @BrendonSled 
If I understand correctly, this commit sets the entire Layout color to that of the current screen if tabs are hidden.
I'm not sure this is needed, user should define `drawUnderTabBar: true` if he intends to hide the `BottomTabs` or even better, RNN should stretch the `Screen` to fit the entire layout when tabs are hidden and `drawUnderTabBar === false`

BrendonSled:
That should be the case @guyca but on screen transitions from `tabBarHidden: false` to a new screen with `tabBarHidden: true` there is a block under the tabs that is color mismatched. It's very easy to reproduce

guyca:
Fair point @BrendonSled 👍

spikeisryu:
I have the same issue. 
image

How can I resolve this issue?

react-native-navigation version is v1.1.152
react-native v0.45.1
and testing iOS

jackmew:
The same issue In iOS.  I need to set tabBarTranslucent: true to solve this problem.

PR #1439, opened 2017-06-30
[iOS] Prefer use of view controller hierarchy to display lightbox

eliperkins:
| Before | After |
| ----- | ----- |
|lightbox-actionsheetios | lightbox-actionsheetios-fixed |

This fixes issues with other React Native views which rely on the view controller hierarchy.


eliperkins:
I also have a sample project here that shows the issue: https://github.com/eliperkins/react-native-navigation-lightbox-issue

guyca:
Thanks for the pr @eliperkins !

PR #1440, opened 2017-06-30
[iOS] Allow lightbox tap to dismiss without background color

eliperkins:
This allows for tapping the background without setting a color here, meaning this feature can be enabled with one less view in the view hierarchy. Additionally, this allows for taps on any background view, visual effect view, color tint view, etc, but not on the content.


guyca:
Thanks @eliperkins

PR #1441, opened 2017-07-01
[iOS] KYDrawerController support

egorkhmelev:
This PR adds support of `KYDrawerController` to current drawer implementation. This library is similar to default Android implementation (I need the exact iOS layout in my project). The only trade-off is that it supports left or right drawer pane but not both and this behaviour controlled by direction. If developer defines the right one it will switch to opposite direction, thats it. 

Also this PR adds missing `Navigator.setDrawerStyle`.

### DIFF
--- a/ios/RCCDrawerController/KYDrawerController/KYDrawerController.h
+++ b/ios/RCCDrawerController/KYDrawerController/KYDrawerController.h
@@ -1,0 +1,64 @@
+//
+//  KYDrawerController.h
+//  KYDrawerController
+//
+//  Created by Yifei Zhou on 1/8/16.
+//  Copyright © 2016 Yifei Zhou. All rights reserved.
+//
+
+#import <UIKit/UIKit.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@class KYDrawerController;
+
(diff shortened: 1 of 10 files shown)


eliperkins:
I'm not sure what value bringing in `KYDrawerController` does here that `MMDrawerController` can't handle. I don't think having two drawer libraries in the project is a good idea.

>  (I need the exact iOS layout in my project)

If you only need this styling on iOS, I would recommend [switching on the value of `Platform.OS`](https://facebook.github.io/react-native/docs/platform-specific-code.html) to handle this.

egorkhmelev:
Ok, I see. But I cannot figure out how to do a drawer with MMDrawer to overlay center controller and not shift it (see Google Inbox)? Also I see no option to implement it inside my project as navigator is not really extensible and Platform.OS don't really help here. Any recommendation on this?

guyca:
I'm not sure adding another drawer controller to the project is a good idea. If I'm not mistaken we already have two third party drawer libraries, adding another just adds more complexity with questionable value.
The next versions will focus on making RNN more extendable and flexible. RNN should provide some interface which will allow users to create plugins and add their own custom native components like drawers and bottom tabs.

kyle-ssg:
Damn, that's a bit of a shame. Overlaying the center controller would be great.

Issue #1419, opened 2017-06-27
startTabBasedApp not passing new props in passProps when used multiple times


richardmillen94:
### Issue Description

I have an iOS app which launches a login interface using startSingleScreenApp and subsequently the main app interface using startTabBasedApp. The first time startTabBasedApp is called the props are passed as expected but I've included a logout flow which relaunches the login interface using startSingleScreenApp and again relaunches the main interface with startTabBasedApp after login. After the second call to startTabBasedApp the app launches okay but the props passed in passProps are incorrect and they remain the same as the first launch.

### Steps to Reproduce / Code Snippets / Screenshots

```javascript
login(credentials, (success, username, version) => {
	if (success) {
		console.log(Tag, `logged in with username: ${username}`)
		Navigation.startTabBasedApp({
			tabs: [...],
			tabsStyle: tabStyle,
			passProps: { username, version }
		})
	}
})
```

The 'username' variable logs correctly here on second login but in the tab screen constructors it logs as the 'username' from the initial launch.

---
### Environment
* React Native Navigation version: 1.1.79
* React Native version: 0.43.4
* Platform(s) (iOS, Android, or both?): iOS
* Device info (Simulator/Device? OS version? Debug/Release?): iOS 10.3.1 Simulator, debug build.

PR #1460, opened 2017-07-04
Fixed props failing to pass to tabs after startTabBasedApp... iOS

BrendonSled:
Fixed props failing to pass to tabs after startTabBasedApp has been called more than once.

Closes Issue #1419

### DIFF
--- a/src/deprecated/platformSpecificDeprecated.ios.js
+++ b/src/deprecated/platformSpecificDeprecated.ios.js
@@ -600,9 +600,7 @@ function savePassProps(params) {
 
   if (params.tabs) {
     _.forEach(params.tabs, (tab) => {
-      if (!tab.passProps) {
-        tab.passProps = params.passProps;
-      }
+      tab.passProps = params.passProps;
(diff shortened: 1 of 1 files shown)


eliperkins:
This change looks like it was added in intentionally in 88e8c90955969ec37a8defa90a5305a6bea2319f by @artald.

I'm wondering if this was done intentionally?

artald:
@eliperkins it's a pretty old commit so I don't remember exactly what was it was all about.

Notice that the current piece of code is setting the `passProps` of a tab with the `passProps` of the params only if the tab doesn't already have its own `passProps`, so if the tab has props they remain intact. We don't want to set the same props to all the tabs all the time, right?

At the moment we don't have any issues with passing props to tabs, so I wonder if there's something else going on there. If we do come to a conclusion that this fix is the actual issue, we'll need to first test it thoroughly before we can accept this.

BrendonSled:
It's pretty easy to reproduce with the code sample in #1419

PR #1461, opened 2017-07-04
Fixed LightBox not updating overlay width/height on rotate

### LINKED ISSUE #1270: [iOS, Android] Orientation change in LightBox
LightBox screen dimensions do not update when orientation is changed. I tried ScrollView with flex 1 as a root element. Display LightBox in portrait and rotate the device.
OnLayout is not fired and there is no way how to handle orientation change.

Am I missing something?

---
### Environment
* React Native Navigation version: 1.1.77
* React Native version: 0.40.0
* Platform(s) (iOS, Android, or both?): both
* Device info (Simulator/Device? OS version? Debug/Release?): iOS simulator, iPhone SE (10.3.2), OnePlus 2 (6.0.1)

BrendonSled:
Closes #1270

(This was an iOS only issue)


ios/RCCLightBox.m
```diff
@@ -48,10 +48,27 @@ -(instancetype)initWithFrame:(CGRect)frame params:(NSDictionary*)params
                 UIColor *backgroundColor = [RCTConvert UIColor:style[@"backgroundColor"]];
                 if (backgroundColor != nil)
                 {
-                    self.overlayColorView = [[UIView alloc] initWithFrame:CGRectMake(0, 0, frame.size.width, frame.size.height)];
+                    self.overlayColorView = [[UIView alloc] init];
                     self.overlayColorView.backgroundColor = backgroundColor;
                     self.overlayColorView.alpha = 0;
+                    [self.overlayColorView setTranslatesAutoresizingMaskIntoConstraints:NO];
                     [self addSubview:self.overlayColorView];
+
```
eliperkins: While this will work, what do you think about just relaying out `self.overlayColorView` in `layoutSubviews` instead? This class already overrides it and most of RN uses manual view layout (because of using Yoga) instead of opting in to use autolayout.

eliperkins: I think `self.overlayColorView.frame = self.bounds` in `layoutSubviews` should do the trick too.

BrendonSled: thanks, I've updated

ios/RCCLightBox.m
```diff
@@ -48,9 +48,10 @@ -(instancetype)initWithFrame:(CGRect)frame params:(NSDictionary*)params
                 UIColor *backgroundColor = [RCTConvert UIColor:style[@"backgroundColor"]];
                 if (backgroundColor != nil)
                 {
-                    self.overlayColorView = [[UIView alloc] initWithFrame:CGRectMake(0, 0, frame.size.width, frame.size.height)];
+                    self.overlayColorView = [[UIView alloc] init];
                     self.overlayColorView.backgroundColor = backgroundColor;
                     self.overlayColorView.alpha = 0;
+                    [self.overlayColorView setTranslatesAutoresizingMaskIntoConstraints:NO];
```
eliperkins: Do we still need this? It seems that since we're manually setting the bounds anyway, we don't need to toggle anything with the autoresizing mask.

guyca: @BrendonSled Ping. I'd like to continue with this PR. Can you check if this is still required?

BrendonSled: @guyca yea sorry about that, I've removed it

PR #1469, opened 2017-07-06
Prevent back button flicker when calling resetTo

kyle-ssg:
The following occurs on iOS when calling resetTo

# Before

# After


kyle-ssg:
Apologies, realised the after gif was the same on first submission. This basically prevents the back button quickly flickering regardless of if either screen had one.

guyca:
Thanks @kyle-ssg
I'll get to this once I return from chain-react conf 👍

ios/RCCNavigationController.m
```diff
@@ -211,6 +211,8 @@ - (void)performAction:(NSString*)performAction actionParams:(NSDictionary*)actio
     
     RCCViewController *viewController = [[RCCViewController alloc] initWithComponent:component passProps:passProps navigatorStyle:navigatorStyle globalProps:nil bridge:bridge];
     
+    viewController.navigationItem.hidesBackButton = YES;
```
eliperkins: Hm, so just so I'm understanding, by doing this, we make sure that the back button is hidden during the reset, and then once the JS context for the next view controller kicks in, it'll set the correct properties?

It's weird to me that the back button even shows in the first place when calling `setViewControllers`, since it should theoretically wipe out the whole nav stack...

kyle-ssg: I don't think it's about setting the correct JS properties for the next controller - the back button flickers regardless of the back button props coming through the bridge. I was under the assumption that it was normal to have to specify hiding the back button when performing any routing.  

I do agree with the last statement, however I don't have the ObjC knowledge / the time to diagnose issues with the routing flow - I just know this works.

eliperkins: > I was under the assumption that it was normal on iOS side to have to specify hiding the back button when performing any routing.

I don't think that's necessarily true. I think navigation stacks should always be constructed in a way that is useful for the user to navigate using the back button. I would say hiding the back button is almost a bit unnatural, in terms of iOS interface guidelines.

From the [iOS Human Interface Guidelines](https://developer.apple.com/ios/human-interface-guidelines/interaction/navigation/):
> *Use a navigation bar to traverse a hierarchy of data*. The navigation bar’s title can show the current position in the hierarchy, and the back button makes it easy to return to the previous location.

kyle-ssg:
any update on this ?

guyca:
@kyle-ssg Hey sorry for not picking this up.
From what I understand the current behavior is the correct one. We can still go forward with the PR but we should make your change opt-in. wdyt?

kyle-ssg:
Hey @guyca, no I don't think the current behaviour is correct - it might be easier for me to make a repo to show you.

Navigating from the initial route with resetTo flickers the back button for around 200ms before hiding

guyca:
@kyle-ssg I misunderstood the issue description, missed the fact this is happening when calling `resetTo`. merging 👍

kyle-ssg:
No problem, thanks for merging!

PR #1474, opened 2017-07-07
Add tabBarBlur for native blur mask

birkir:
Added tabBarBlur to `tabsStyle` for better iOS native look and feel.

Uses `UIBarStyle`(https://developer.apple.com/documentation/uikit/uibarstyle?language=objc) for setting the type of blur style.

Value can be `true || false || 'dark'`


birkir:
Also it could be nice to add option to draw under the tabBar like navBar has the option to do.

Currently we need to have marginBottom negative offset of the tabBar's height.

```js
render() {
  const tabBarHeight = Platform.select({ android: 54, ios: 64 });
  return (
    <View style={{ marginBottom: -tabBarHeight }}>
      ...
    </View>
  );
}
```

eliperkins:
Can't this be achieved by setting `tabBarBackgroundColor` to a dark color and `tabBarTranslucent` to true?

I'm pretty against adding an arbitrary view to the tab bar.

PR #1490, opened 2017-07-11
[iOS] Full screen back gesture

gevorg94:
Added panGestureRecognizer to allow swiping back over the entire screen.
Just need to pass `enabledBackGestureFullScreen: true` in `navigatorStyle`.

### DIFF
--- a/docs/styling-the-navigator.md
+++ b/docs/styling-the-navigator.md
@@ -62,6 +62,7 @@ this.props.navigator.setStyle({
   statusBarHidden: false, // make the status bar hidden regardless of nav bar state
   
   disabledBackGesture: false, // default: false. Disable the back gesture (swipe gesture) in order to pop the top screen. 
+  enabledBackGestureFullScreen: true, // default false. Enable swipe back gesture over the entire screen.
   screenBackgroundImageName: '<name of image in Images.xcassets>', // Optional. default screen background image.
   rootBackgroundImageName: '<name of image in Images.xcassets>', // Static while you transition between screens. Works best with screenBackgroundColor: 'transparent'
 
--- a/ios/RCCAnimator.h
+++ b/ios/RCCAnimator.h
@@ -1,0 +1,16 @@
+#import <UIKit/UIKit.h>
+
+// Undocumented animation curve used for the navigation controller's transition.
+FOUNDATION_EXPORT UIViewAnimationOptions const RCCNavigationTransitionCurve;
+
+@class RCCAnimator;
+
+@protocol RCCAnimatorDelegate <NSObject>
+
+@end
+
+@interface RCCAnimator : NSObject <UIViewControllerAnimatedTransitioning>
+
+@property (nonatomic, weak) id<RCCAnimatorDelegate> delegate;
+
+@end
--- a/ios/RCCAnimator.m
+++ b/ios/RCCAnimator.m
@@ -1,0 +1,89 @@
+#import "RCCAnimator.h"
+
+UIViewAnimationOptions const RCCNavigationTransitionCurve = 7 << 16;
+
+
+@implementation UIView (TransitionShadow)
+- (void)addLeftSideShadowWithFading
+{
+    CGFloat shadowWidth = 4.0f;
+    CGFloat shadowVerticalPadding = -20.0f; // negative padding, so the shadow isn't rounded near the top and the bottom
+    CGFloat shadowHeight = CGRectGetHeight(self.frame) - 2 * shadowVerticalPadding;
+    CGRect shadowRect = CGRectMake(-shadowWidth, shadowVerticalPadding, shadowWidth, shadowHeight);
+    UIBezierPath *shadowPath = [UIBezierPath bezierPathWithRect:shadowRect];
+    self.layer.shadowPath = [shadowPath CGPath];
+    self.layer.shadowOpacity = 0.2f;
+    
+    // fade shadow during transition
+    CGFloat toValue = 0.0f;
+    CABasicAnimation *animation = [CABasicAnimation animationWithKeyPath:@"shadowOpacity"];
+    animation.fromValue = @(self.layer.shadowOpacity);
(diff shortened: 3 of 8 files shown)


eliperkins:
I'm not sure if this is something that belongs in the react-native-navigation core, since it's not a feature of UINavigationController... I wonder if there's a way that we could accomplish this by allowing for custom navigation transitions using `UIViewControllerAnimatedTransitioning`, such that we could support this, as well as things like Shared Element Transitions, like Android does...

joshuaflores:
@gevorg94 Would it be possible to adjust the code to accept a decimal value? Although personally I would just pass the full screen width, configuration makes the prop much more useful. Perhaps you could call it`backGestureWidth`.

@eliperkins I think you're on to something. An abstraction over the view transition animation would give RN developers a freedom to explore UX/UI design patterns currently only available to native platform developers. Any ideas how to something like this could be implemented?

eliperkins:
@joshuaflores: I would start with this blog post detailing how to work with the API: https://www.objc.io/issues/5-ios7/view-controller-transitions/

It won't be a small amount of work to expose this as a nice JavaScript API, but I think it's achievable 😄

I'd be happy to discuss with anyone who wants to pick up working on this using `UIViewControllerAnimatedTransitioning`!

joshuaflores:
@eliperkins I know some swift, no obj-c, but I'll take look. I'm going to start by trying to add the functionality requested in #1557 and see what a js abstraction would look like.

Just for an added incentive to anyone who know's objc and can build this api, these are some examples that could be replicated in react-native.

eofx7
[stackoverflow](https://stackoverflow.com/a/36863581)

frontflip-slow
[raywenderlich](https://www.raywenderlich.com/110536/custom-uiviewcontroller-transitions)

joshuaflores:
@eliperkins what are you're thoughts on using [lottie](https://github.com/airbnb/lottie-ios) to implement transitions, they support custom view controller animations? 

Initially the ideas of integrating lottie feels wrong, but give it some time and it might make sense. Sure RNN will now have a dependency, but it'll gain a wonderful transition abstraction that will eventually be cross-platform. Even better since the transitions are just json files they can easily be shared with the community. 

My reservations are that these animation are not interactive, they don't respond to gestures. For that we would have to implement UIViewControllerInteractiveTransitioning. I know, it's starting to look like a rabbit's hole. 

To give you an idea of what a possible api with lottie would look like:

```
import Animations from '../animations'
....
this.props.navigator.push({
   screen: 'example.TestScreen',
   ...,
   animation: Animations.flipX
})
```

If you think this could work let me so we can turn this into an issue and get more input on how this should be implemented.

eliperkins:
@joshuaflores Hmmm, bringing in lottie seems like a pretty heavy dependency for this lib... That'd mean that we'd need users of this lib to link against not only react-native-navigation's static library, but also with lottie. Additionally, users would need to have/know After Effects to make their own transitions (including us, if we want to ship pre-defined ones!).

I think the API you outlined might work out alright, but keep in mind that we'll wanna be linked into `UINavigationControllerDelegate`, meaning we'll need a reference to the `UINavigationController` object itself.

I would say that anyone wanting a custom `UIViewControllerAnimatedTransitioning` style (i.e., the ones you posted), they'll need to be dropping into UIKit, regardless. That makes me not as concerned about supporting this cross-platform right out of the gates. Keep in mind that the manifesto of this project is to support each platform's core navigation paradigms rather than reimplementing them in JS-land.

I think with the API you've outlined, you can at least check those values when doing the push right now, and then assign a custom `UIViewControllerAnimatedTransitioning` object that's written in Objective-C. I would start by just using the simple transition examples before handling gestures.

joshuaflores:
@eliperkins So my idea was to bring lottie in as a static library to avoid this issue or add it as a cocoapod (more work but easier to update). Concerning After Effects, I think it's less of an issue if you consider that the alternative is writing UIKit code, and I'm confident the community will create a platform to share the json files. Also, lottie can run natively it's wont be a js hack.

That being said I fully agree that exposing a dependency-free api would be the best route to go.
A less controversial approach would be be to create a small collection of common animations `(FlipX, FlipY, Fade)` in their respective native code and expose a `Transition` module in the js library that references each native animation. 

```
import {Transition} from 'react-native-navigation'

this.props.navigation.showModal({
  screen: '...',
  transition: Transition.FlipY
})
```

I assume with this approach others could create transition libraries that could be used instead, but again my obj-c and java experience is non-existing, so this is only an assumption. My main concern is that this repo will start getting more PRs and issues to support more animations.

I'm going to think on this some more, but I think this new approach should simplify things further.

exentrich:
Please merge this commit! Back gesture over entire screen is massive UX improvement. Check this in Instagram app. Try once, and you can't live without it.

eliperkins:
I feel pretty passionate about not overriding default behaviors of the OS with this library. I'd prefer to see this functionality done _within_ consumer apps, by leveraging the API that the library exposes. The [manifesto of this project](https://github.com/wix/react-native-controllers) talks about how this project should "[allows] you to easily wrap core native UI components without compromising on the native experience." By overriding the native OS defaults, we're getting away from the core native experience.

I also feel passionate about keeping this project focused on wrapping the core native UI components, rather than extending them. The extending them should exist in other projects outside of this one, just as React Native works. React Native is a core library with the ability to hook in and extend it in other ways. I see this project in the same vein.

MrLoh:
I understand that the default should be the apple iOS behavior but having this as an option would be nice. It's such a common pattern on iOS: Twitter, Facebook, Reddit, Instagram all use it, just to name a few. So it's certainly the most common navigation behavior on the biggest iOS apps except for native apple apps like Mail and iMessage.

MrLoh:
@gevorg94 Would it be possible to refactor this code into a separate package that acts as a plugin to this one?

fokoz:
@gevorg94 There is a weird bug here when popping, the screen position seems wrong and then stretch back to normal position.

Issue #1492, opened 2017-07-11
Can't find React headers on iOS


tspecht:
### Issue Description

I'm currently trying to integrate React-Native into my existing Swift iOS project. I'm using Cocoapods already so I followed the documentation on the topic by adding the following to my Podfile:

```
# React Native
react_native_path = "../node_modules/react-native"
pod "Yoga", :path => "#{react_native_path}/ReactCommon/yoga"
pod "DoubleConversion", :podspec => "#{react_native_path}/third-party-podspecs/DoubleConversion.podspec"
pod "GLog", :podspec => "#{react_native_path}/third-party-podspecs/GLog.podspec"
pod "Folly", :podspec => "#{react_native_path}/third-party-podspecs/Folly.podspec"
pod ‘React’, :path => react_native_path, :subspecs => [
  ‘Core’,
  'CxxBridge',
  'DevSupport',
  ‘RCTImage’,
  ‘RCTNetwork’,
  ‘RCTText’,
  ‘RCTWebSocket’
]
```
This all works well enough and I'm able to build my app both in the simulator as well as on the device and play around with it.

In a second step I wanted to add react-native-navigation. I added the module to package.json, ran npm install and afterwards react-native-link. All these commands finished successfully. Unfortunately when now trying to build the project it can't find any of the React header files while compiling the react-native-navigation dependency. I already tried various combinations of downgrading react-native again, adapting HEADER_SEARCH_PATHs and so on but with no success ...

react-native version: 0.46.1
 react version: 16.0.0-alpha.12 
react-native-navigation version: 1.1.134

Any idea what could cause this?

### Steps to Reproduce / Code Snippets / Screenshots
See above
---
### Environment
* React Native Navigation version: 1.1.134
* React Native version: 0.45.1
* Platform(s) (iOS, Android, or both?): iOS
* Device info (Simulator/Device? OS version? Debug/Release?): Everything

andriantosg:
Hi @tspecht, 
If you are using Cocoapods and already link this react-native-navigation to your project, then you just need to do additional steps to make it works.

1. Open your Xcode, find `Libraries/ReactNativeNavigation.xcodeproj`
2. Select ReactNativeNavigation on TARGETS section then go to Build Settings
3. Find Header Search Paths then add `$(SRCROOT)/../../../ios/Pods/Headers/Public` with **recursive** option.
4. Build your project

Hopefully can help solve your problem.

Note: We need to do this until It officially supports Cocoapods.

[...]

PR #1468, opened 2017-07-06
Add CocoaPods support to v1

ceyhuno:

PR #1496, opened 2017-07-12
Support Cocoapods

andriantosg:
We need to support react-native projects via Cocoapods installation. This PR will solve issue #1492.


ceyhuno:
There is #1468  already. Is there a problem with that?

PR #1508, opened 2017-07-15
Feature/v2 like api

yusufyildirim:


yusufyildirim:
That was mistake, sorry about that...

PR #1520, opened 2017-07-18
Added overrideBackPress to LightBox

BrendonSled:


guyca:
Thanks! @BrendonSled Correct me if I'm wrong, but the hardware back button click isn't sent to the lightbox, right?

BrendonSled:
Yes that's right

PR #1521, opened 2017-07-18
Fix app start from background crash

BrendonSled:


FRizzonelli:
Hello there, is this fix working? Will it be added to master?

PR #1529, opened 2017-07-19
v2 toggleTabs support

yogevbd:


DanielZlotin:
This PR needs to change:
this command shouldn't be part of the Navigation static API, it should be inside the styles (NavigationOptions)
something like
```
static navigationOptions: {
    tabs: {
        visible: false/true,
        animated: false/true
    }
}
```

Issue #1259, opened 2017-05-17
Navbar not translucent when using background color


chrise86:
I have the following `navigatorStyle`:

```
navBarBackgroundColor: 'rgba(210, 29, 53, 0.76)',
navBarTranslucent: true,
drawUnderNavBar: true
```

but the content does not seem to actually draw under and shine through the navbar. Removing the background colour fixes this, however the navbar then turns a grey colour which is not what I want. 

Is there any way of setting a navbar tint colour?

Ehesp:
Yeah so this is something called immersive mode which is currently not supported (Android). I'm not sure what the iOS equivalent is called. ccing @guyca as I've spoke with him about this in the past.

chrise86:
@Ehesp I think from scouting around StackOverflow that there is a hack to get this effect, which involves setting a transparent background image :-/

ZionChang:
@chrise86 
It works on both iOS and android, when I use the following style in my screen.
```javascript
static navigatorStyle = {
    drawUnderNavBar: true,
    navBarTranslucent: true,
    navBarTransparent: true,
    navBarBackgroundColor: 'transparent'
};
```

chrise86:
@ZionChang the issue is not with getting a transparent navbar, the issue is not having translucency when using a background colour (other than transparent, which isn't necessary)

ZionChang:
@chrise86 I totally got it wrong about what you said, sorry.

huhu087:
@chrise86 here is the pr: https://github.com/wix/react-native-navigation/pull/1548

PR #1548, opened 2017-07-21
Add navBarTransparency param.

huhu087:
Usage:
like this: navigatorStyle: {
                  navBarBackgroundColor: '#e6514a',
                  navBarTextFontSize: 18,
                  navBarTextColor: 'white',
                  navBarTranslucent: true,
                  drawUnderNavBar: true,
                },

Test passed.


henrikra:
Can you explain bit more or/and provide screenshots?

brenwell:
I tried this, and it appears to work. in my app its so subtle its a bit hard to tell. But would be nice to solve #1259

PR #1581, opened 2017-07-27
Currently visible screen

guyca:
This PR adds api to get the currently visible screen.

Usage:

1. From a screen, using navigator instance: 
    `const isVisible = await this.props.navigator.isCurrentlyVisibleScreen()`

2. Globally, using `Navigation` object:
     `const visibleScreenInstanceId = await Navigation. getCurrentlyVisibleScreenId()`


aarongreenwald:
@guyca @artald THANK YOU!!!

brunolemos:
> In order to have any use of this method, you'd need to map instanceId to screens your self.

Any example of this? I can't find in the docs how to get the instance id. 

BTW why doesn't it return the `screen` name that I used when registering (e.g. `org.brunolemos.HomeScreen`) instead of `screenInstanceID2`?

PR #1600, opened 2017-07-29
Update top-level-api.md (Added forgotten styles)

berkaybeyaz1:
Added forgotten drawer style props.


docs/top-level-api.md
```diff
@@ -66,6 +66,12 @@ Navigation.startTabBasedApp({
       screen: 'example.SecondSideMenu', // unique ID registered with Navigation.registerScreen
       passProps: {} // simple serializable object that will pass as props to all top screens (optional)
     },
+    style: {
```
guyca: We should mention that these styles (the entire `style` object actually) are for iOS only

PR #1601, opened 2017-07-29
DOCS: Fixed android installation

brapifra:
The documentation for the android installation of v2 was wrong.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,7 +140,7 @@ NSURL *jsCodeLocation = [[RCTBundleURLProvider sharedSettings] jsBundleURLForBun
 
 	```groovy
 	include ':react-native-navigation'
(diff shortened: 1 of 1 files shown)


guyca:
This is wrong, MainActivity should extend SplashActivity as the docs state.

brapifra:
In v2 there is no SplashActivity, it was replaced by NavigationActivity. MainActivity tries to extend SplashActivity but fails because it doesn't exist.
https://github.com/wix/react-native-navigation/blob/v2/lib/android/app/src/main/java/com/reactnativenavigation/NavigationActivity.java

guyca:
Sorry, didn't notice this PR was for v2 branch 👍

DanielZlotin:
We still need to add `createAdditionalReactPackages` or similar mechanism

PR #1616, opened 2017-08-02
changed setOptions to receive RNNNavigationOptions and added topBarTextColor

bogobogo:


lib/ios/RNNModalManager.m
```diff
@@ -33,7 +34,7 @@ -(void)dismissAllModals {
 -(void)removePendingNextModalIfOnTop {
 	NSString *containerId = [[_store pendingModalIdsToDismiss] lastObject];
 	
-	UIViewController *modalToDismiss = [_store findContainerForId:containerId];
+	RNNRootViewController *modalToDismiss = [_store findContainerForId:containerId];
```
DanielZlotin: why? containerId can be any viewcontroller, not just RNNRootViewController

Issue #1446, opened 2017-07-02
[iOS] ScrollView/FlatList *BUG* after changing route/pushing/showing modal while there is a momentum scroll


grundmanise:
### Issue Description

There is a bug when you have a `ScrollView` / `FlatList` (haven't tried with others) on one screen and you switch (/ push / show) to a different screen while `ScrollView` / `FlatList` is scrolling with a momentum, when you return back to the prev. screen (screen with `ScrollView` / `FlatList`) it's children / cells would not respond to touches until: a) you scroll `ScrollView`to the end/top (with a bounce) b) the cell item re-renders (FlatList)

### Steps to Reproduce / Code Snippets / Screenshots

https://github.com/grundmanise/react-native-navigation-scrollview-bug
1. Scroll few items (ScrollView should not bounce, i.e. do not scroll till the end of the list)
2. While there is a momentum scrolling => switch to the 2nd tabbar screen
3. Return to the 1st screen and try to click a `FlatList` cell => cells are not responsive.

---
### Environment
* React Native Navigation version: 1.1.104
* React Native version: 0.44.0
* Platform(s) (iOS, Android, or both?): iOS
* Device info (Simulator/Device? OS version? Debug/Release?): Simulator & Device, 10.3,  Debug & Release

varungupta85:
+1 I am able to reproduce this issue in the app. I have seen it a couple of times earlier also but I couldn't put a finger on the steps to reproduce. Thanks @grundmanise for pinpointing the issue. I am not sure if this has always been there or introduced later.

varungupta85:
@DanielZlotin Do you guys also face this bug in the Wix app?

guyca:
@gran33 Think you could look into this during guild day tomorrow? It's reproduces constantly in our app as well.

gran33:
@DanielZlotin yes. I can reproduce it over and over again. 

I try to put some light on this limbo case (yes super important). 
Any extra info will be appreciated.

guyca:
Fixed in v1.1.169

blairvanderhoof:
I think this is still broken for TextView

![image](https://user-images.githubusercontent.com/833890/28904693-9ddab564-77c1-11e7-9940-06f7db03229f.png)


```
Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '-[RCTTextView scrollViewDidEndDecelerating:]: unrecognized selector sent to instance 0x7fd51a0641c0'

```

[...]

PR #1618, opened 2017-08-02
walk-around for React-Native bug with multiple RCTRootView

gran33:
https://github.com/wix/react-native-navigation/issues/1446

### DIFF
--- a/ios/RCCViewController.m
+++ b/ios/RCCViewController.m
@@ -269,6 +269,22 @@ const NSInteger TRANSPARENT_NAVBAR_TAG = 78264803;
   return [NSString stringWithFormat:@"%lld", milliseconds];
 }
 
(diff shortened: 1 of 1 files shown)


gran33:
This bugFix is a walk-around limbo case in react-native when the using multiple `RCTRootView`.

The bug description:
Having bottom tabs (tab bar) with 2 tabs (at least).
In tab #1 having `ScrollView` and perform scroll. While the scrolling is still in progress (haven't stopped yet) switch tab #2. Back to tab #1 and try to click on any button in the `ScrollView`. They aren't clickable (the native get this event but the JS block the app from getting it).
In order to get out of this limbo case, simply scroll with momentum and wait the `ScrollView` to stop. Now the buttons are clickable.

PR #1619, opened 2017-08-02
update docs for iOS drawer animations in top-level-api section

rajaishwary:


henrikra:
Wow did not know you can use these kind of animations :O

PR #1628, opened 2017-08-02
Remove createJSModules @override marker - RN 0.47 compatibility

radko93:
[`createJSModules` is now not required on Android](https://github.com/facebook/react-native/commit/ce6fb337a146e6f261f2afb564aa19363774a7a8) from RN 0.47. This is backwards compatible according to my tests.


guyca:
Thanks

PR #1630, opened 2017-08-03
Fix crash with scrollViewDidEndDecelerating and RCTTextView

blairvanderhoof:
See https://github.com/wix/react-native-navigation/issues/1446#issuecomment-319858633


gran33:
@blairvanderhoof Thanks man! Great catch.

PR #1644, opened 2017-08-05
[iOS] Fix conditions for tabBarSelectedLabelColor & selectedButtonColor

datso:
Without that fix it is not possible to show original iconImageSelected and also tabBarSelectedLabelColor uses invalid condition.


ios/RCCTabBarController.m
```diff
@@ -166,6 +166,10 @@ - (instancetype)initWithProps:(NSDictionary *)props children:(NSArray *)children
     id selectedIcon = tabItemLayout[@"props"][@"selectedIcon"];
     if (selectedIcon) {
       iconImageSelected = [RCTConvert UIImage:selectedIcon];
+      
+      if (!selectedButtonColor) {
```
guyca: Can you please clarify this change?

datso: I can't set an original icon for selected state without UIImageRenderingModeAlwaysOriginal. Without this condition, selectedIcon are always colored with the default (light-blue) color.

guyca: Sorry to trouble you, Can you please include code samples to help me understand the flow this fixes?

datso: Sorry for long delay, without this fix it is not possible to show selected icon as is.

Without fix:
without_fix
With fix:
with_fix

datso: The same for tabBarSelectedLabelColor.

datso: @guyca

krystofcelba: Please merge this. I have the same problem.

PR #1647, opened 2017-08-06
Export `Navigator` class for public use

pcfutures:
When using ESLint rule `react/prop-types`, an error is show when using `this.props.navigator` if `navigator` is not specified in the components propTypes.

Changing this library to export the `Navigator` class allows users to use it in propTypes, i.e:

```JavaScript
import PropTypes from 'prop-types';
import { Navigator } from 'react-native-navigation';

...

Component.propTypes = {
    navigator: PropTypes.instanceOf(Navigator),
};
```

Whereas before you had 2 options, either:
1. Use `PropTypes.object`, which is another ESLint error (`react/forbid-prop-types`),

or

2. create your own `PropTypes.shape()`, but that feels convoluted and error-prone.

Hope this all makes sense.


tayfunyasar:
It still giving an error.

`
import { Navigator } from 'react-native-navigation';
import PropTypes from 'prop-types';

Component.propTypes = {
  dispatch: PropTypes.func.isRequired,
  navigator: PropTypes.instanceOf(Navigator),
};

`

error:
https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/require-default-props.md

PR #1671, opened 2017-08-09
Reworked React implementation to closely follow Androids lifecycle

BrendonSled:
This is a major refactor.. Not everything may work and I'm looking for feedback.

**Breaking change**
Splash screen - createSplashLayout _removed_ - Instead use getSplashLayout

**Why**
Currently React Native Navigation does not handle Android's garbage collection.
You can repeat these issues with master by going to Settings -> Developer options -> Limit background processes and settings that to No background processes.

Every time GC runs the app is rendered unusable.

Also there are lots of crashes around GC that I have noticed in fabric

I have moved the entire JS thread creation to the only activity that should be around and drastically cleaned up the lifecycle management.


ncnlinh:
Any progress with this?

PR #1674, opened 2017-08-10
[iOS/Android] Add initialTabIndex support

Eluss:
This pull requests introduces ability to set tab index that is selected when component is rendered for both iOS and Android. 
iOS solution is based on Proteye's [pull request](https://github.com/wix/react-native-navigation/pull/553)


Eluss:
Thanks for merging this pr!

PR #1695, opened 2017-08-15
fix for breaking change in React Native 0.47

JPig:
fix for breaking change in React Native 0.47 the deprecated createJSModules Override


JPig:
Not too sure why all the checks failed.

JPig:
Seems the Tests Failing are totally unrelated to the PR upon actually checking the CI logs.

DanielZlotin:
Thanks for the PR
why not remove this method alltogether?

DanielZlotin:
additionally, if you checked that this is the only change needed for latest stable RN, if you can, please update `playground`'s `plackage.json` to use correct RN and react versions (along with dependencies) and update the yarn.lock

JPig:
I could be wrong but i think the method needs to stay for backwards compatability?
And when i get a chance i can look into doing that.

JPig:
just investigated it and removing the method all together causes issues for anything < 0.47. so i feel keeping the method would be best.

PR #1697, opened 2017-08-16
add adjust soft input optional config for Lightbox on android platform

onigirisan:
The window of LightBox is different with the activity，adjust soft input mode in onActivityCreated life cycle callback will have no effect


guyca:
This is great. Can you please add it to the [docs](https://wix.github.io/react-native-navigation/#/screen-api?id=showlightboxparams-) as well? Otherwise this api will be left out of v2

onigirisan:
Hi guyca, I have updated doc, Can you please check it again

PR #1743, opened 2017-08-25
[Android] fix conditions for  destroyJsIfNeeded

onigirisan:
Currently  ReactNativeHost will be cleared at  onDestroy callback when currentActivity is null or finishing. Mostly, it works well, but in some case, it will results in the app lost interaction with js context, Think about the case below 

steps:

1. startSingleScreenApp(A)
2. restart app by startSingleScreenApp(B)
3. start native activity C immediately
4. do some thing then fish C

complete lifecycle flow:

1. A onCreate - A onStart - A onResume 
2. A onPause - B onCreate - B onStart - B onResume 
3. **B onPause** - C onCreate - C onStart - C onResume - A onStop - **A onDestory**
4. C onStop - C onDestory - **B onResume**

It can be seen that B onPause before A onDestory，so the static reference currentActivity will set to null which results in ReactNativeHost be cleared onDestory

It an uncommon case I know, but I really meet the problem and need solve it

### DIFF
--- a/android/app/src/main/java/com/reactnativenavigation/controllers/NavigationActivity.java
+++ b/android/app/src/main/java/com/reactnativenavigation/controllers/NavigationActivity.java
@@ -53,6 +53,7 @@ public class NavigationActivity extends AppCompatActivity implements DefaultHard
      * Along with that, we should handle commands from the bridge using onNewIntent
      */
     static NavigationActivity currentActivity;
+    private static int activityCount;
 
     private ActivityParams activityParams;
     private ModalController modalController;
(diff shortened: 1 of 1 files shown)


guyca:
Hey @onigirisan Thanks for the pr, I'll get to it by the end of the week

onigirisan:
Hi @guyca Sorry to bother you, Is this pr can be merged?

guyca:
Hey @onigirisan 
I've dedicated this Thursday to going over rnn PR's and issues. I'll review your pr first thing on Thursday.
Sorry for the delay

android/app/src/main/java/com/reactnativenavigation/controllers/NavigationActivity.java
```diff
@@ -53,6 +53,7 @@
      * Along with that, we should handle commands from the bridge using onNewIntent
      */
     static NavigationActivity currentActivity;
+    private static int activityCount;
```
henrikra: Why this needs to be static?

guyca: So that the counter is shared between all NavigationActivity instances, and destroy only when the last one is destroyed.

PR #1746, opened 2017-08-25
setButtons for iOS

Johan-dutoit:
I've also done the event side on Android, but will tackle the full implementation once this gets merged and PR it when it's ready.


henrikra:
So what I can do after this PR?

PR #1820, opened 2017-09-09
[v2] TabBarItem support for iOS

joonhocho:
I am adding customizing TabBarItem for iOS.
Most of codes are ported from v1.

### DIFF
--- a/lib/ios/RNNControllerFactory.m
+++ b/lib/ios/RNNControllerFactory.m
@@ -115,6 +115,11 @@
 	}
 	[vc setViewControllers:controllers];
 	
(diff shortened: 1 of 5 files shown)


DanielZlotin:
Thank you for the PR, @bogobogo please review, seems like some things are duplicated with our current codebase

joonhocho:
Also added support for default selected index. It's as easy as putting 'selected: true' to your tab definition:
```javascript
{
selected: true,
container: {...},
tabItem: {...}
}
```

larryranches:
@joonhocho Does this support changing background colors for an individual tab?

yogevbd:
fixed in #2571

Issue #1514, opened 2017-07-17
App crash on reload (cmd+r) v2


masterpetit:
### Issue Description

App crashes on reload.

```2017-07-17 11:53:30.917 [info][tid:com.facebook.react.JavaScript] Running application "navigation.appname.root" with appParams: {"rootTag":1,"initialProps":{"containerId":"Container2"}}. __DEV__ === true, development-level warning are ON, performance optimizations are OFF
2017-07-17 11:53:30.925 AppName[11327:101435] *** Assertion failure in -[RCTEventEmitter sendEventWithName:body:](), ..../node_modules/react-native/React/Modules/RCTEventEmitter.m:42
2017-07-17 11:53:30.942 AppName[11327:101435] *** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'bridge is not set. This is probably because you've explicitly synthesized the bridge in RNNEventEmitter, even though it's inherited from RCTEventEmitter.' 
```

Warning diplayed on startup:
```
Sending `RNN.appLaunched` with no listeners registered.

logToConsole @ index.ios.bundle:12332
logIfNoNativeHook @ index.ios.bundle:12315
__callFunction @ index.ios.bundle:5070
(anonymous) @ index.ios.bundle:4911
__guard @ index.ios.bundle:5044
callFunctionReturnFlushedQueue @ index.ios.bundle:4910
(anonymous) @ debuggerWorker.js:71
```

### Steps to Reproduce / Code Snippets / Screenshots

Just configure the library with the default react native init app.

---
### Environment
* React Native Navigation version: alpha
* React Native version: 0.46.3
* Platform(s) (iOS, Android, or both?): iOS (did not test on Android yet)
* Device info (Simulator/Device? OS version? Debug/Release?): Simulator/iOS 10/Debug.

masterpetit:
https://github.com/wix/react-native-navigation/issues/1397

rustybox:
I know next to nothing about react native or react native navigation, so this may well be a terrible way to solve the problem.

That said, making the following change in ReactNativeNavigation/RNNEventEmitter.m seems to work, so if this is killing your productivity you can roll with this till a more official solution comes through.

```ObjectiveC
# Before 
-(void)send:(NSString *)eventName body:(id)body {
	[self sendEventWithName:eventName body:body];
}

# After
-(void)send:(NSString *)eventName body:(id)body {
	if (self.bridge)
	{
[...]

PR #1836, opened 2017-09-11
[v2] fix bridge reload issue

joonhocho:
https://github.com/wix/react-native-navigation/issues/1514
There could be better solution than this, but I think it's good enough for now.


rustybox:
I really don't think this is good enough, that's why I made a comment instead of a pull request.

If this is to be merged in it should probably note the hack in a comment so someone can take a look in the future.

joonhocho:
@rustybox I agree. Added TODO comment. I still think it's 10x better than nothing since you can't reload without this.

suren1525:
It work for me too.

jslz:
hi, how do we follow which release commits get into? Are these fixes in 1.1.x versions or only 2.x versions? Thanks.

PR #1838, opened 2017-09-11
Stop calling host.clear() when activity is destroyed

guyca:
This is a breaking change and has to be opt-in. To prevent host.clear from being called,
Override `clearHostOnActivityDestroy` in MainApplication and return false:

```java
    @Override
    public boolean clearHostOnActivityDestroy() {
        return false;
    }
```

If host isn't cleared, the next time the app is opened react context might still be initialized.
In this case we emit appLaunched event which has to be handled in Js:

```js
  Promise.resolve(Navigation.isAppLaunched())
      .then((appLaunched) => {
        if (appLaunched) {
          startApp();
        }
        new NativeEventsReceiver().appLaunched(() => {
          startApp();
        });
      })
```

### DIFF
--- a/android/app/src/main/java/com/reactnativenavigation/NavigationApplication.java
+++ b/android/app/src/main/java/com/reactnativenavigation/NavigationApplication.java
@@ -118,6 +118,10 @@ public abstract class NavigationApplication extends Application implements React
 
     public abstract boolean isDebug();
 
+    public boolean clearHostOnActivityDestroy() {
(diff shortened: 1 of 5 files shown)


joshuapinter:
Hi All, is this PR still relevant in the `master` branch. Launching the example app shows the same loss of state when the Activity is destroyed in the background. Here's a GIF of the issue in action: 

May-05-2019 09-44-43

Issue #1858, opened 2017-09-14
Fix iOS11 safeArea effects on scrollView


gran33:
### Issue Description

There is a margin under the top bar that shouldn't be there.
Related to iOS11 and react-native issue:
https://github.com/facebook/react-native/issues/15681

### Steps to Reproduce / Code Snippets / Screenshots

Running any app with screen contains any scroll view, when push then pop, the back screen comes from the bottom.

---
### Environment
* React Native Navigation version: 1.1.85
* React Native version: 0.45.1
* Platform(s) (iOS, Android, or both?): iOS
* Device info (Simulator/Device? OS version? Debug/Release?): Simulator + real device

gran33:
fixed in #1859

DanielZlotin:
implement in v2

gameboyVito:
it's not working on iOS10 while building with Xcode9, though it works on iOS11. See this: https://github.com/facebook/react-native/issues/15957

gameboyVito:
Fixed, I will send you a PR if you don't mind.

DanielZlotin:
So the current solution is a bug. This forces the users to specify magic number in their code, instead of letting the system do it's thing.

LeoNatan:
Fixed in v2. Will back-port to v1.

PR #1859, opened 2017-09-14
fix iOS11 safeArea

gran33:
https://github.com/wix/react-native-navigation/issues/1858


DanielZlotin:
@gran33 do that in v2 as well?

gran33:
@DanielZlotin Open an issue :)

LeoNatan:
This is not the correct solution.

LeoNatan:
Fixed in v2. Will back-port to v1.

PR #1882, opened 2017-09-18
[V2] Android topbar font family

Cool04ek:


lib/android/app/src/main/java/com/reactnativenavigation/utils/TypefaceLoader.java
```diff
@@ -0,0 +1,54 @@
...
+import android.content.res.AssetManager;
+import android.graphics.Typeface;
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class TypefaceLoader {
```
DanielZlotin: cover with tests

Issue #1886, opened 2017-09-19
The scrollview breaks on iOS 10 when using RN-0.48


gameboyVito:
### Error:

> ***Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '-[RCTCustomScrollView setContentInsetAdjustmentBehavior:]: unrecognized selector sent to instance 0x7ffab9891a00'***

### Reproduce:
Build your app by XCode9-GM, then run it on your iOS10 device or below version. Make sure you have the stable version of react-native-navigation@1.1.228.

### Reference
https://github.com/facebook/react-native/issues/15957

### Env:
- `Device - iOS 10.3.1`
- `XCode 9.0 (9A235)`
- `react-native - 0.48.3`

gameboyVito:
Any update?

PR #1896, opened 2017-09-20
Bug fixed - the UIScrollView will break on iOS10 when building by XCode9

gameboyVito:
This only happens when you are using react-native@0.48.3, and build your APP by XCode9. When you run your APP in iOS10 or below, it will throw `unrecognized selector` error. The new iOS11 has new design on StatusBar which behaves weird on UIScrollView, so that the react-native team added this new selector to fix it. This PR will help to address this problem when you are also using the amazing `react-native-navigation`.

Here's the error details:
> Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '-[RCTCustomScrollView setContentInsetAdjustmentBehavior:]: unrecognized selector sent to instance 0x7ffab9891a00'

Reference:
https://github.com/facebook/react-native/issues/15957
https://github.com/wix/react-native-navigation/issues/1886

### DIFF
--- a/ios/RCCViewController.m
+++ b/ios/RCCViewController.m
@@ -292,7 +292,7 @@ const NSInteger TRANSPARENT_NAVBAR_TAG = 78264803;
 // rnn issue - https://github.com/wix/react-native-navigation/issues/1858
 - (void)_traverseAndFixScrollViewSafeArea:(UIView *)view {
 #ifdef __IPHONE_11_0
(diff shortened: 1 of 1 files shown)


SudoPlz:
I had that problem w/ XCode9 and react-native **0.46.3** so it does not only happen on 0.48.3

mehdijjz:
Can anyone please merge this ?

gran33:
Thanks man!

Issue #1861, opened 2017-09-14
When the app is bundled there are no fading animations.


voidstarfire:
Easily the best navigation library for React native (And I've used a few) keep up the good work.

Anyhow; my first issue with it!

The fading animation is disabled when going from screen to screen. It works fine on react-native run-android apps. (live dev build). Not ok on bundled builds.

Example:
```
this.props.navigator.push({
            screen: 'screen.SecondScreen',
            title: 'Some title',
            animated: true,
            animationType: 'slide-horizontal'
});
```

Now when I bundle this with 

```
react-native bundle --dev false --platform android --entry-file index.android.js --bundle-output android/app/build/intermediates/assets/release/index.android.bundle --assets-dest android/app/build/intermediates/res/merged/release
```

Or I bundle with debug
```
react-native bundle --dev false --platform android --entry-file index.android.js --bundle-output android/app/build/intermediates/assets/debug/index.android.bundle --assets-dest android/app/build/intermediates/res/merged/debug
```

### Environment
* React Native Navigation version: FILL THIS OUT
* React Native version: 0.47.0
* Platform(s) (iOS, Android, or both?): Android
* Device info, Both on real devices and emulator. Debug and release builds. Not on dev builds.

voidstarfire:
After digging, replacing 'AppRegistry.registerComponent('itsquizreactnative', () => App);' with 
'Navigation.startSingleScreenApp({' also causes the same bug.

voidstarfire:
This issue is to do with how you register the app witihn index.android.js 

I you don't use ```AppRegistry.registerComponent('itsquizreactnative', () => App);```

It will stop the fade animation.

voidstarfire:
After upgrading to RN 0.48 fade animation is completely vanished (all build versions including dev)

PR #1904, opened 2017-09-21
Fixed an issue with the React navigation animations

voidstarfire:
https://github.com/wix/react-native-navigation/issues/1861

### DIFF
--- a/android/app/src/main/java/com/reactnativenavigation/screens/ScreenAnimator.java
+++ b/android/app/src/main/java/com/reactnativenavigation/screens/ScreenAnimator.java
@@ -23,7 +23,6 @@ class ScreenAnimator {
     private static final int DURATION = 250;
     private static final int ALPHA_START_DELAY = 100;
     private static final int ALPHA_SHORT_DURATION = 150;
-    private static final LinearInterpolator LINEAR_INTERPOLATOR = new LinearInterpolator();
     private static final DecelerateInterpolator DECELERATE_INTERPOLATOR = new DecelerateInterpolator();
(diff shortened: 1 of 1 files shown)


guyca:
Hey @voidstarfire,
The alpha animation was actually intentionally removed from `slide-horizontal` animation. Personally, I think the animation looks better with out the alpha. Also I've noticed that Samsung has changed the default activity animation on their devices to slide animation, which doesn't have an alpha component. Of course the iOS slide animation also doesn't have alpha.

I think we should keep the `slide-horizontal` animations as is, with out alpha. If you'd like, you're welcome to add a new animation type with alpha. wdyt?

voidstarfire:
Maybe there should be an option? (Like a boolean flag for both imcomming and outgoing)

Personally I think the alpha looks much better because the slide animation sometimes lags and lowers the quality of the app's appearance.

Also the slight fade gives the impression that the app is silky smooth and going backwards feels like a much nicer process. At the moment it feels very iOS'y.

This was originally raised as bug by my QA team hehe. (Why are screens not fading).
According to material design sliding in and direction where-by a screen is destroyed should be accompanied by a fade at least.

guyca:
Adding a new type `slide-horizontal-fade` sounds like a good idea 👍

voidstarfire:
Great
I'll get into it next week :-)
When was this removed? (I'd like to reference old source code)

guyca:
89eb30a54290950426b94ebe0b9fc5c7560cb259

voidstarfire:
FYI I will begin work on this tomorrow, been on vacation (Just letting you know so you don't close)

voidstarfire:
I don't have time do this now, (currently working 60 hour weeks) can somebody else take it over thanks?

guyca:
Closing the pr as custom animations are supported in v2

Issue #1502, opened 2017-07-14
You cannot render into anything but top root


diennguyentien:
I get this error after update RN. The showLightBox is not shown

---
### Environment
* React Native Navigation version: 1.1.134
* React Native version: 0.46.2
* Platform(s) (iOS, Android, or both?): both
* Device info (Simulator/Device? OS version? Debug/Release?): Release

Bhullnatik:
I have somewhat the same issue, one the 2 errors come when dismissing a LightBox on Android.

### Environment
* React Native Navigation version: 1.1.134
* React Native version: 0.46.2
* Platform(s) (iOS, Android, or both?): Android
* Device info (Simulator/Device? OS version? Debug/Release?): *

![screenshot_1500323207](https://user-images.githubusercontent.com/2040807/28288379-4b7f1a0a-6b3f-11e7-9061-922fee1c21ea.png)
![screenshot_1500323299](https://user-images.githubusercontent.com/2040807/28288384-4d6987ba-6b3f-11e7-9948-4016745d1883.png)

lipingruan:
@diennguyentien @Bhullnatik 

node_modules/react-native-navigation/android/app/src/main/java/com/reactnativenavigation/layouts/:

class SingleScreenLayout & BottomTabsLayout

rewrite function : 

```java

    @Override
    public void dismissLightBox() {
        if (lightBox != null) {
            lightBox.dismiss();
            // lightBox.hide();
            lightBox = null;
        }
    }

```

diennguyentien:
@lipingruan thanks you

rauliyohmc:
Does @lipingruan workaround work for you guys?

sijad:
@lipingruan's workaround works.

rauliyohmc:
That will only work if you want to dismiss the lightbox from your screen, meaning by calling `this.props.navigator.dismissLightBox()` from your component.

However, if you wanna also close the lightbox by tapping in the background (common use case) using `tapBackgroundToDismiss` it will still break, i.e:

```javascript
this.props.navigator.showLightBox({
  screen: screens.lightBox.sort,
  style: {
    backgroundBlur: 'dark',
    backgroundColor: 'rgba(0, 0, 0, 0.7)',
    tapBackgroundToDismiss: true,
  },
});

```
[...]

PR #1734, opened 2017-08-23
Add hacky fix to make Lightbox dismiss not crash on Android for recent RN versions

rawrmaan:
Mainly submitting this PR to re-state the urgency of fixing #1502 and start a discussion about a better way to solve this issue than this hacky fix (proposed by @rauliyohmc).

EDIT: Upon further testing, this hack only fixes the problem some of the time.

PR #1910, opened 2017-09-22
fix android lightbox error when dismissing

1mike12:
hack that was mentioned to have fixed issue in https://github.com/wix/react-native-navigation/issues/1502

This may interfere with `componentWillUnmount` 

Taken from a comment in the thread: 

@hemedani Thanks, your workaround works quite nice. But within my lightbox I'm dependent on the componentWillUnmount method beeing called. With your workaround this method is never called for my lightbox. So currently I have a really ugly hack for this:


bdauria:
Looks very much like https://github.com/wix/react-native-navigation/pull/1734/commits

1mike12:
dupe of pre-existing

PR #1916, opened 2017-09-23
Support abstract method getJSBundleFile() in NavigationReactNativeHost to use Code Push

moonjava2005:
Dear all,
I found that NavigationReactNativeHost doesn't have getJSBundleFile() abstract method. So, we cannot use the Code Push with react-native-navigation v2


guyca:
@moonjava2005 You've branched out of master and submitted the pr to v2 branch. Please submit a new pr from the correct branch.
Thanks

PR #1923, opened 2017-09-24
Apply styled buttons to all UIControlState

rgoldiez:
Applying a custom size/font/etc to navBar buttons is now applied to all `UIControlState`s.  Previously it was only applied to the normal state (ie, not select/focused/highlighted/disabled states).

This PR fixes that.


rgoldiez:
No longer needed.  A similar PR was just accepted.

Issue #1710, opened 2017-08-18
How to disable TabBar button color? Want to use the original color of the icon image.


spikeisryu:
### Issue Description

Need to use images for TabBar buttons like these.
Normal
![timeline](https://user-images.githubusercontent.com/23102875/29449199-c844acc2-8434-11e7-9869-9a817f2a6271.png)
Selected
![timeline_rv](https://user-images.githubusercontent.com/23102875/29449217-d7e9d6e8-8434-11e7-9e3c-495f21ef6805.png)
and the background color of the TabBar needs to be a certain one.

Therefore, I used tabsStyle like this.

            tabsStyle: {
                tabBarBackgroundColor: '#AB013E',
            },

Then, the result shows like this.
![image](https://user-images.githubusercontent.com/23102875/29449328-34d56624-8435-11e7-8af6-2afdb4c29564.png)

It seems that the default button color is blue and it overwrite the color of the original image.

And I when I set a specific color, like 'white', then it still overwrite the color of the original image, like this.
![image](https://user-images.githubusercontent.com/23102875/29449463-d67cebd2-8435-11e7-9965-09b198c7bcfa.png)

I tried not to use tabsStyle, but it that case, I can't set the background color of the tab and the default color of the button, 'blue', still overwrites the image, like this.
![image](https://user-images.githubusercontent.com/23102875/29449560-2fd4a210-8436-11e7-84b8-ab101cab0537.png)

How can I use the original color of the button, not the tabsStyle settings?

---
### Environment
* React Native Navigation version: 1.1.130
* React Native version: 0.47.1
* Platform(s) (iOS, Android, or both?): iOS
* Device info (Simulator/Device? OS version? Debug/Release?): Simulator iPhone 7, iOS 10.3

saxahan:
This is exactly my deserving issue recently. Can anyone achieve this?

ianataffinity:
You'll need a pull request to implement`disableIconTint` for TabBar icons.

For example in the docs:
https://github.com/wix/react-native-navigation/blob/d224bf4d3903ce16b287f5243c736aee67d30824/docs/adding-buttons-to-the-navigator.md

And implemented here in iOS (in Android too):
https://github.com/wix/react-native-navigation/blob/c62d944c2b10a841e531c05b45f88f18f902042a/ios/RCCNavigationController.m#L386

gvillenave:
[...]

PR #1943, opened 2017-09-28
Add disableIconTint option for TabBar icons (#1710)

gvillenave:
This PR adds a `disableIconTint` and `disableSelectedIconTint` options in AppStyle to prevent the Tab Bar icons to be forced to `tabBarButtonColor` and `tabBarSelectedButtonColor`. As discussed in #1710, these properties live in AppStyle even for iOS to be able to access those when calling `setTabButton`. I think eventually all the Tab Bar styling settings should live in only one place for both platforms (whether it's TabsStyle or AppStyle) to make usage less confusing.


gvillenave:
I went a little too fast, there's still a caveat for setTabButton on Android. Will submit a new PR once I figure it out.

Noitidart:
Does this affect Android too?

I have tried `disableIconTint` as true and false however it is always taking the color of the image. I want it to take the nav bar button color automatically.

PR #1945, opened 2017-09-29
Error in documentation - AppDelegate.m has no return statement.

itsyogesh:
Since the docs suggested we remove everything inside the `didFinishLaunchingWithOptions`, without the return statement there was an `Control reaches end of non-void function` error since the function expects a boolean return. After adding this, the error goes away.


henrikra:
I noticed this too so I can approve this 👍

PR #1959, opened 2017-10-03
condition -fix for case of top level passProps in single-screen-app

rajaishwary:
…en-app


guyca:
Closing as we're not accepting PR's to v1 anymore.

PR #1968, opened 2017-10-04
[Android] Allow setting statusBarTextColorScheme globally via appStyle

Agontuk:
Currently statusBarTextColorScheme is ignored if provided from appStyle. In my case, I was trying to set white status bar with dark "statusBarTextColorScheme" globally and it wasn't working. So made this quick fix.


guyca:
Thanks a lot for the pr. Unfortunately I missed it, and resolved it last week.

PR #1970, opened 2017-10-04
Added support for RTL sideMenu in Android

nidongara:


android/app/src/main/java/com/reactnativenavigation/views/SideMenu.java
```diff
@@ -76,11 +76,23 @@ public void setEnabled(boolean enabled, Side side) {
     }
 
     public void openDrawer(Side side) {
-        openDrawer(side.gravity);
+        try{
+            openDrawer(Gravity.LEFT);
```
guyca: Is there no other way to determine the menu we should open?

yershalom:

android/app/src/main/java/com/reactnativenavigation/views/SideMenu.java
nidongara: @guyca  I could not find any other way. this looked like the easiest way. Do we have any other work around?

PR #1973, opened 2017-10-05
Fixes popTo signature

tim-dev:
The only argument to `popTo` is the containerId to pop to, not the current containerId as well.  Also showing `this.props.previousContainerIds` lead me to believe that was included automatically (similar to `this.props.containerId`) so I changed that to avoid confusion.


DanielZlotin:
My guess is in the future we will have `params` as a second argument, but currently you're right. Thanks

PR #2043, opened 2017-10-19
Prevent unwanted app destroy on Android

eeynard:
Hi,

I was running into an annoying issue while trying to ask user permissions in my home screen on android (I have a map on the home screen so I need to ask the user for his location). I am using react-native-permissions and since it use react-native PermissionsAndroid API you must call the `Permissions.request` function after starting an activity or you will end up with the error `Tried to use permissions API while not attached to an Activity`.

The problem is that even when you call the `Permissions.request` function in the `componentDidMount` callback of your home screen, your app ends up crashing when the user accept or decline the permission with the error shown above.

So after some investigations I realize that when the native permissions pop up shows up, the app is put in pause and the `currentActivity` is null. I forgot to mention that I have a Login flow first and at the end I call `startSingleScreenApp` so the first activity is destroyed and replaced by the new one. The problem here is that the `onDestroy` callback is called after the `onPause` of the new screen. So when is get trough `destroyJsIfNeeded` function, `currentActivity` is null and the app is destroyed, causing the `Tried to use permissions API while not attached to an Activity`.

I apologize if it is not very clear but it is pretty hard to explain !
I don't have a global view of the project so please tell me if I am doing something wrong.


android/app/src/main/java/com/reactnativenavigation/controllers/NavigationActivity.java
```diff
@@ -133,6 +135,7 @@ protected void onNewIntent(Intent intent) {
     @Override
     protected void onPause() {
         super.onPause();
+        lastActivity = currentActivity;
```
guyca: You're leaking the activity by saving it in a static field after `onPause`.

Note that `currentActivity` is static, `onDestroy` of first activity is called after `onResume` of second activity. Perhaps we can change the conditional in `destroyJsIfNeeded` to to:
`(currentActivity == null || (currentActivity.isFinishing() && currentActivity == this))` this way `onDestroyApp` is called only if the finishing activity is the current activity.

eeynard: Oh I see .. 

The problem is not related to the second part of the conditional but with the first one: `currentActivity == null` In fact, if the app is paused between the second screen `onResume` and the first screen `onDestroy` the `currentActivity` will be null and de app will be destroyed.

I will try to find another way to handle this case.

eeynard:
@guyca what about this approach ? Not the best code I wrote but I don't see any other straightforward solution.

guyca:
The issue is basically that the react context is lost since `host.clear()` is called from, right?
You can override `clearHostOnActivityDestroy` in `MainApplication` and return `false`
Can you true that before we merge you pr?

eeynard:
I already tried that solution and it partially worked, in fact the app does not crash when the permissions are accepted/declined and you can use it without any problem but when you close it with the back button and try to relaunch it, it gets stucked on the SplashScreen. Is it another issue ? Or maybe I miss used this `clearHostOnActivityDestroy` ?

guyca:
Did you modify your Js code accordingly?

```js
  Promise.resolve(Navigation.isAppLaunched())
      .then((appLaunched) => {
        if (appLaunched) {
          startApp();
        }
        new NativeEventsReceiver().appLaunched(() => {
          // The event is dispatched from SplashScreen and starts the app.
          // This flow happens when the app was started while react context was 
          startApp();
        });
      })
```

At any case, lets merge the pr as users who didn't override `clearHostOnActivityDestroy` are affected by this.

android/app/src/main/java/com/reactnativenavigation/controllers/NavigationActivity.java
```diff
@@ -151,6 +153,7 @@ protected void onDestroy() {
         destroyLayouts();
         destroyJsIfNeeded();
         NavigationApplication.instance.getActivityCallbacks().onActivityDestroyed(this);
+        numberOfActivities -= 1;
```
guyca: I noticed you're incrementing activity count in onCreate and decrementing in onDestroy. This doesn't seem like what you want as the activity can be recreated after onStop

eeynard: You’re right, I was looking at the android [Activity Lifecycle](https://developer.android.com/guide/components/activities/activity-lifecycle.html) and I am afraid I can’t find any way to reliably use this counter.

eeynard:
@guyca I tried your advice and it seems to work on Android :)
To recap, from a clean react-native-navigation version (I mean without the stuff in my PR): 
- I overrided `clearHostOnActivityDestroy` in my MainApplication.java
```java
@Override
public boolean clearHostOnActivityDestroy() {
    return false;
}
```
- I changed my index.js for: 
```javascript
import 'react-native';
import { Navigation, NativeEventsReceiver } from 'react-native-navigation';

import app from './src/app';

Promise.resolve(Navigation.isAppLaunched())
.then((appLaunched) => {
  if (appLaunched) {
    app();
  }
  new NativeEventsReceiver().appLaunched(() => {
    app();
  });
});
```
And everything seems OK ! Except for iOS which throw `TypeError: _platformSpecificDeprecated2.default.isAppLaunched is not a function` Should we add a dummy function to `src/platformSpecific.ios.js` which always return true or should I update my code and have a different implementation for Android and iOS ?

eeynard:
This PR is no longer justified, I close it. Thanks for the help @guyca :)

guyca:
Hey @eeynard 
Just saw your comment, sorry for the delay. I don't think we should add a dummy function as that might be misleading. Glad it worked out.

PR #2048, opened 2017-10-20
Fix badgeColor for IOS < 10

freiserg:


freiserg:
Fixed https://github.com/wix/react-native-navigation/pull/2068

PR #2051, opened 2017-10-20
[iOS] Enable (Dis)Appear events for drawer

Royce:
`navigatorEventId` is required for `[self sendScreenChangedEvent:@"didAppear"];` etc., so we should also provide it to the drawer screens.


rgoldiez:
Yes!!!!!!

Jedidiah:
Thank you @Royce 🙌 
Was just about to dive in and add this too then found you PR. Hopefully will get merged soon.

yershalom:

rgoldiez:
Not sure why this isn't getting merged.  It works. @gran33 - Can you take a look and see what else is needed to accept this PR?

Issue #1728, opened 2017-08-22
screenIsCurrentlyVisible() always returning false in iOS


MrPeproni:
Hello,

I am using `const isVisible = await this.props.navigator.screenIsCurrentlyVisible();` in Android and it's working fine.

Now i am developing my app in iOS so this is returning false everytime
`const isVisible = await this.props.navigator.screenIsCurrentlyVisible();`

Please guide... where i am doing mistake and how can i solve this
Thanks in Advance :)

jbolter:
Same issue here.

PR #2052, opened 2017-10-20
[iOS] Fix getCurrentlyVisibleScreenId with drawer

Royce:
This fixes two issues with `Navigation.getCurrentlyVisibleScreenId` and `this.props.navigator.screenIsCurrentlyVisible()` on iOS.

- The wrong "controllerId" was being set for the initial screen/layout. `controllerId` is only used to return a `screenId` for Navigation.getCurrentlyVisibleScreenId.
- Take into account drawers, when trying to find the visible viewController.

### DIFF
--- a/ios/RCCManagerModule.m
+++ b/ios/RCCManagerModule.m
@@ -371,6 +371,14 @@ RCT_EXPORT_METHOD(
     {
         return [self getVisibleViewControllerFor:vc.presentedViewController];
     }
+    else if ([vc isKindOfClass:[TheSidebarController class]]) {
+        TheSidebarController *drawerController = (TheSidebarController*) vc;
+        return [self getVisibleViewControllerFor:drawerController.contentViewController];
+    }
+    else if ([vc isKindOfClass:[MMDrawerController class]]) {
+        MMDrawerController *drawerController = (MMDrawerController*) vc;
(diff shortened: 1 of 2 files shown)


Royce:
Fixes #1728

rynop:
@Royce `this.props.navigator.screenIsCurrentlyVisible()` is still always `false` for me on iOS.  react-native-navigation `v1.1.316`.  I'm not using drawer mode, I'm only using `Navigation.startTabBasedApp()`.

I've tried using `this.props.navigator.screenIsCurrentlyVisible()` both on a screen that is opened from a tab AND a screen that is opened in a modal.

Whats strange is calling `const visibleScreenInstanceId = await Navigation.getCurrentlyVisibleScreenId()` always returns a strange screen id like `controllerID1_nav0` instead of the screen ID that i registered via `Navigation.registerComponent('my.screen', () => ThingFinder, store, Provider);`

Royce:
await Navigation.getCurrentlyVisibleScreenId() can't/won't return your
'my.screen'
because a 'my.screen' screen could be mounted multiple times. I don't
think await
Navigation.getCurrentlyVisibleScreenId() is really meant to be part of
the public API.

rynop:
@Royce thanks makes sense. Shouldn't `this.props.navigator.screenIsCurrentlyVisible()` work though? (not always return `false`)?

Issue #1765, opened 2017-08-30
BackHandler problem


jsina:
#### I wanna do sth on screen after the android back button fired but there is no effect effect I've done it.
```
componentWillMount() {
    this._backPress = 0;

    BackHandler.addEventListener('backPress', () => {
      console.log('back ****');
      setTimeout(() => {
        this._backPress = 0;
      }, 3000);

      this._backPress += 1;
      if (this._backPress <= 1) {
        ToastAndroid.showWithGravity(strings.BACK_BUTTON_ALERT, ToastAndroid.SHORT, ToastAndroid.CENTER);
        return true;
      }
      return false;
    });

  }
```

MrLoh:
I'm having the same issue with `BackHandler.addEventListener('hardwareBackPress', this.myHandler);` I need some way to override this in a youtube player, which displays back buttons on android in fullscreen, but I don't want to always go back a screen, just to disable fullscreen with the button.

Hizoul:
i have the same problem.
I also tried the overrideBackPress / overrideBackPressJs param. It works in that no backpress does anything anymore, but the "backPress" event that should be fired somehow never reaches my navCallback :/
https://github.com/wix/react-native-navigation/blob/90df99030cb7495a4f862efd538aab8b0456bce1/android/app/src/main/java/com/reactnativenavigation/screens/ScreenStack.java#L398
https://wix.github.io/react-native-navigation/#/screen-api?id=setonnavigatoreventcallback

jsina:
@Hizoul could you send us an example code?

Hizoul:
Whenever you call the navigator.pop or navigator.push method make sure to do it like this
```
navigator.pop({overrideBackPress: true})
navigator.push({ screen: "abc", overrideBackPress: true})
```

then on each root component of a route you call the following
```
    navigator.setOnNavigatorEvent((event) => {
      switch (event.id) {
        case `backPress`: {
console.log("received backpress on android") 
        }
      }
    })
```

but thats just what i already wrote in pseudocode

truongthanh1303:
Hi,
I have an issue I think it also related to this issue. Do you have a look at it, please?
https://github.com/wix/react-native-navigation/issues/749#issuecomment-331627087

Regards,

Hizoul:
Did you also pass `overrideBackPress` as a param to the `startSingleScreenApp` function?

truongthanh1303:
Yes I did, but it didn't work on the first screen?

[...]

PR #2057, opened 2017-10-22
Fix issues with Android BackHandler

artald:
Fix issues with Android BackHandler listeners not getting called, not allowing to override the back functionality from the JS side

relates to #1765

### DIFF
--- a/android/app/src/main/java/com/reactnativenavigation/controllers/NavigationActivity.java
+++ b/android/app/src/main/java/com/reactnativenavigation/controllers/NavigationActivity.java
@@ -172,12 +172,14 @@ public class NavigationActivity extends AppCompatActivity implements DefaultHard
 
     @Override
     public void invokeDefaultOnBackPressed() {
-        super.onBackPressed();
+        if (layout != null && !layout.onBackPressed()) {
+            super.onBackPressed();
+        }
     }
 
     @Override
     public void onBackPressed() {
-        if (layout != null && !layout.onBackPressed()) {
+        if (getReactGateway().isInitialized()) {
             getReactGateway().onBackPressed();
         }
     }


sijad:
this breaks our app, it makes `overrideBackPress` obsolete and now `backPress` handler called every time. I wish you introduce breaking changes more carefully.

guyca:
@sijad Will fix today

sijad:
BackHandler works. We were using overrideBackPress to switch between rnn backhandler and ours. We need rnn backhandler works as it is with ability to be overrided in some screens.
But with this changes we couldn't find a good way to make our handler only work in few specified screen. We were doing it via overrideBackPress before.

guyca:
Hey @sijad 
Can you please explain how I can reproduce the issue you are experiencing?
1. Push a screen with overrideBackPress true
2. Set a back handler in that screen
3. press back

expected: `backPress` event is received, BackHandler doesn't get called
result: `backPress` event isn't received, BackHandler handles the event

Is this the issue?

guyca:
@sijad I've reverted the problematic commit for now

sijad:
@guyca Thanks!

our app is tab based, this is what we're experiencing:

1. Set a back handler in one of tab screens (for asking user to confirm before exist application)
2. Push a screen with overrideBackPress `false`
3. press back

expected: navigation should pop opened screen
result: backPress event is received, BackHandler handles the event and exit confirmation shows up, but it shouln't

peakyDicers:
I am having this issue now, has anybody found a solution?

sijad:
this is how we ask user to press back button twice to exit:

```js
import {
  BackHandler,
  ToastAndroid,
} from 'react-native';
import { ScreenVisibilityListener } from 'react-native-navigation';

let backPress = 0;
let currentScreen = null;

const tabs = [
  'screen.tab.one',
  'screen.tab.two',
  'screen.tab.three',
];

const screenVisibilityListener = new ScreenVisibilityListener({
  didAppear: ({screen}) => {
    currentScreen = screen;
  },
});

BackHandler.addEventListener('backPress', () => {
  if (tabs.includes(currentScreen)) {
    setTimeout(() => {
      backPress = 0;
    }, 3000);

    backPress += 1;
    if (backPress <= 1) {
      ToastAndroid.showWithGravity(
        'Press back again to exit',
      );
      return true;
    }
  }
  return false;
});
```

we have this code in out `app.js`

PR #2067, opened 2017-10-24
Added a hiding option to the status bar when showing the in app notification

eladgel:


ios/RCCNotification.m
```diff
@@ -20,35 +22,38 @@ -(id)initWithParams:(NSDictionary*)params
     {
         self.params = params;
         self.yellowBoxRemoved = NO;
-
```
eliperkins: Please revert these whitespace changes as they are out of the scope of this PR

ios/RCCNotification.m
```diff
@@ -335,6 +340,11 @@ -(void)performShowWithCompletion:(void (^)(void))completion
 
 -(void)showWithCompletion:(void (^)(void))completion
 {
+    if(self.hideStatusBar){
+        UIWindow *window = [[RCCManager sharedInstance] getAppWindow];
+        window.windowLevel = UIWindowLevelAlert;
```
eliperkins: I don't think this is a safe way to implement this, as we have no guarantee about the other window levels... I'd rather see the alert be displayed on a new `UIWindow`, rather than reusing the existing one, to implement this feature.

yershalom:

PR #2072, opened 2017-10-24
[v2] Pull tab title from screens navigationOptions

neilkimmett:
Does what it says on the tin. I looked at adding it in `-[RNNavigationOptions applyOn:]` but because that only gets called in `viewWillAppear` the tab title doesn't get set until you navigate to the tab.

### DIFF
--- a/lib/ios/RNNControllerFactory.m
+++ b/lib/ios/RNNControllerFactory.m
@@ -101,8 +101,10 @@
 	NSMutableArray* controllers = [NSMutableArray new];
 	for (NSDictionary *child in node.children) {
 		UIViewController* childVc = [self fromTree:child];
-		
-		UITabBarItem* item = [[UITabBarItem alloc] initWithTitle:@"A Tab" image:nil tag:1];
+
+		NSDictionary* firstChild = [child[@"children"] firstObject];
+		NSString* title = firstChild[@"data"][@"navigationOptions"][@"title"] ?: @"A Tab";
+		UITabBarItem* item = [[UITabBarItem alloc] initWithTitle:title image:nil tag:1];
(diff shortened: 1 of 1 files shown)


billyohgren:
Can we get this merged? Looks lite a much better solution than "A tab"

lib/ios/RNNControllerFactory.m
```diff
@@ -101,8 +101,9 @@ -(RNNTabBarController*)createTabs:(RNNLayoutNode*)node {
 	NSMutableArray* controllers = [NSMutableArray new];
 	for (NSDictionary *child in node.children) {
 		UIViewController* childVc = [self fromTree:child];
-		
-		UITabBarItem* item = [[UITabBarItem alloc] initWithTitle:@"A Tab" image:nil tag:1];
+
+		NSString* title = child[@"children"][0][@"data"][@"navigationOptions"][@"title"] ?: @"A Tab";
```
eliperkins: The access to `[0]` freaks me out a little bit here. What do you think about pulling this line out into something that checks for the first object here, rather than accessing it blindly?

neilkimmett: Sure thing 👍

neilkimmett:
@eliperkins Pulled that `[0]` out into a call to `firstObject` instead. We shouldn't need to explicitly check for nil, it'll just fall through and default to "A Tab".

yogevbd:
Already fixed in latest version.

```
Navigation.setRoot({
      bottomTabs: {
        children: [{
          component: {
            name: 'navigation.playground.TextScreen',
            options: {
              bottomTab: {
                title: 'Tab 1'
              }
            }
          }
        }]
      }
    });
```

Issue #1643, opened 2017-08-05
[Feature request] iOS 11 Style Navbar Titles 


hammadj:
### Issue Description

Almost every iOS 11 app has this new style for navigation titles:

![ios11 nav title](https://user-images.githubusercontent.com/7143583/28991088-f617b5e4-7940-11e7-92ca-ca14ba6b4d5d.gif)

It would be great to support this natively with react-native-navigation since it seems to be the new UI standard for iOS. It shouldn't be too difficult to accomplish either because I believe iOS has a simple API to enable it: https://developer.apple.com/documentation/uikit/uinavigationbar/2908999-preferslargetitles

chrise86:
There are 2 parts to this, the large titles and the ability to set a `searchController` on the `navigationItem` - both would be a very welcome addition 👍

B3rry:
+1

phuongwd:
+1

PavSidhu:
the API you suggested can't be used because React-Navigation is purely Javascript. This would need to be recreated from scratch.

jarvisluong:
@PavSidhu did you just go to the wrong repo? this is react-native-navigation, not react-navigation

PavSidhu:
@jarvisluong Yeah, I just realised my mistake and removed my comment.

joaqtor:
+1

eliperkins:
Configuring `prefersLargeTitles`, `largeTitleTextAttributes` and `largeTitleDisplayMode` should be pretty straight forward, since they're all similar properties to what we expose already.

Configuring `searchController` will not be as straight forward, since we don't have an API for search controllers anywhere in React Native.

Also, as a reminder to all those commenting "+1", please use the reactions instead of comments, so as not to spam this thread with unhelpful comments. Pull requests to implement this functionality would be more than welcome!

gnihphp:
+1

jarvisluong:
Any update from this feature?

Sauri89:
+1

Sauri89:
+1

Burnsy:
@gran33 Hello, Is there any chance you could give us some insight why the support was removed and if and when it could be re added? Thank you!

anonrig:
Any eta on this?

xmarianox:
Any updates on this issue?

oferRounds:
+

chrise86:
[...]

PR #2090, opened 2017-10-28
Support iOS 11 prefersLargeTitles #1643

yomybaby:
Please, see #1643

### DIFF
--- a/ios/RCCViewController.m
+++ b/ios/RCCViewController.m
@@ -645,14 +645,10 @@ const NSInteger TRANSPARENT_NAVBAR_TAG = 78264803;
     }
   }
   
-  if ([[[UIDevice currentDevice] systemVersion] floatValue] >= 11.0) {
+  if ([self.navigationController.navigationBar respondsToSelector:@selector(setPrefersLargeTitles:)]) {
     NSNumber *prefersLargeTitles = self.navigatorStyle[@"prefersLargeTitles"];
     BOOL prefersLargeTitlesBool = prefersLargeTitles ? [prefersLargeTitles boolValue] : NO;
-    if (prefersLargeTitlesBool) {
-      [self.navigationController.navigationBar setPrefersLargeTitles:YES];
-    } else {
-      [self.navigationController.navigationBar setPrefersLargeTitles:NO];
-    }
+    self.navigationController.navigationBar.prefersLargeTitles = prefersLargeTitlesBool;
   }
 }
 
--- a/ios/RCCViewController.m
+++ b/ios/RCCViewController.m
@@ -644,6 +644,16 @@ const NSInteger TRANSPARENT_NAVBAR_TAG = 78264803;
       self.navigationItem.titleView.clipsToBounds = YES;
     }
   }
+  
+  if ([[[UIDevice currentDevice] systemVersion] floatValue] >= 11.0) {
+    NSNumber *prefersLargeTitles = self.navigatorStyle[@"prefersLargeTitles"];
+    BOOL prefersLargeTitlesBool = prefersLargeTitles ? [prefersLargeTitles boolValue] : NO;
+    if (prefersLargeTitlesBool) {
+      [self.navigationController.navigationBar setPrefersLargeTitles:YES];
+    } else {
+      [self.navigationController.navigationBar setPrefersLargeTitles:NO];
+    }
+  }
 }
 
 


ios/RCCViewController.m
```diff
@@ -644,6 +644,16 @@ -(void)setStyleOnAppearForViewController:(UIViewController*)viewController appea
       self.navigationItem.titleView.clipsToBounds = YES;
     }
   }
+  
+  if ([[[UIDevice currentDevice] systemVersion] floatValue] >= 11.0) {
+    NSNumber *prefersLargeTitles = self.navigatorStyle[@"prefersLargeTitles"];
+    BOOL prefersLargeTitlesBool = prefersLargeTitles ? [prefersLargeTitles boolValue] : NO;
+    if (prefersLargeTitlesBool) {
+      [self.navigationController.navigationBar setPrefersLargeTitles:YES];
```
eliperkins: These setters can be written as `self.navigationController.navigationBar.prefersLargeTitles = prefersLargeTitlesBool ` as well, to use Objective-C 2.0 syntax, as well as reuse the `prefersLargeTitlesBool` variable to match the logic.

ios/RCCViewController.m
```diff
@@ -644,6 +644,16 @@ -(void)setStyleOnAppearForViewController:(UIViewController*)viewController appea
       self.navigationItem.titleView.clipsToBounds = YES;
     }
   }
+  
+  if ([[[UIDevice currentDevice] systemVersion] floatValue] >= 11.0) {
```
eliperkins: What do you think about using a `if (@available(iOS 11.0, *))` check here instead of checking the device's system version?

gran33: Since the `@available` can only be used in xcode9 and higher (backward compatible wise), it's better to use device's system version

eliperkins: If the argument is for backwards compatibility, I would argue that using `[self.navigationController.navigationBar respondsToSelector:@selector(setPrefersLargeTitles:)]` would be a better alternative than a version check.

eliperkins:
What can I do to help get this PR merged in?

gran33:
@eliperkins can u plz fi it to 
`[self.navigationController.navigationBar respondsToSelector:@selector(setPrefersLargeTitles:)]`
?

eliperkins:
@gran33 Since I'm just a contributor and not a collaborator, I can't make edits to this PR.

@yomybaby can you make that change?

Otherwise, I'll need to open a different PR or we can merge this and fix it in master.

yomybaby:
Sorry for late reply. I’ll do that tonight. 🙏
2017년 11월 14일 (화) 오전 8:13, Eli Perkins <notifications@github.com>님이 작성:

yomybaby:
@eliperkins I just change some code. Thank for your suggestion.

birkir:
The ScrollView behaviour is a little weird.
The complete animation happens on like 5 pixel scroll.

Works fine if the scrollview content has less height than the scrollview itself.

http://gph.is/2j2FBtP vs http://gph.is/2j1PRT1

eliperkins:
@birkir I don't think this is an issue with this library, but rather just an iOS quirk.

If you find the root cause of this is in `react-native-navigation`, I would encourage you to open a separate issue.

birkir:
Yeah I'm more suspecting React Native in general.

Maybe this [recently added method](https://github.com/facebook/react-native/pull/15609) can be used to fix the problem

icodesign:
@yoavaa @eliperkins  Should we get it merged?

birkir:
What happened? @gran33

gran33:
Sorry guys, This PR breaks xcode8 users, and force xcode8 to update to xcode9.

eliperkins:
Can we open a discussion to updating this library to require Xcode 9? It seems like quite a big deal to cut features of the most recent operating system in order to support an outdated toolchain.

brunolemos:
Well, Xcode 9 is the latest stable version, that's what we should support.

gran33:
@yomybaby can u plz try to build it in xcode8 and add 
```
#ifdef __IPHONE_11_0

YOUR_CODE_HERE

#endif
```

birkir:
About the scroll issue... I think you may be right about it being a iOS quirk. I saw one app that has the same bug [Parcel](https://itunes.apple.com/us/app/parcel-delivery-tracking/id639968404?mt=12) (I don't know if it's an react native app or not).

But somehow Apple makes it look smooth in their apps, I deeply want to fix the problem, so any ideas are greatly appreciated, I don't know where to look or start debugging.

dkast:
Any updates on this issue?

guyca:
@yomybaby Sorry to trouble you, could you resubmit the PR with support for xcode 8?

brunolemos:
@guyca I think if you `git cherry-pick` his commit add add the xcode 8 support would be enough, if he is not available

anonrig:
It seems that this PR is reverted in master, any ETA on the status? I couldn't keep track.

brunolemos:
Where can I add this code on v2?

I'm using [patch-package](https://github.com/ds300/patch-package) and picking some userful PRs.

doertli:
Is this PR likely to get remerged anytime soon?

gran33:
Yes, will update. 
cc: @yogevbd  @guyca

brunolemos:
It was already remerged yesterday on #2593 AFAIK.

Issue #1598, opened 2017-07-28
[IOS]switchToTab error - could not find UITabBarController


neuberoliveira:
### Issue Description
My app was built this way:

	Tab1
		ProductsList
			ProductDetail
				AnotherProductInfo #change to Tab2 from here (switchToTab({ tabIndex: 1 }))
	Tab2
		ListOfChoices #choices picked from 'AnotherProductInfo'

Now i want to navigate from `AnotherProductInfo` directly to `Tab2`, for now the code bellow work only on Android

	this.props.navigator.dismissAllModals({ animationType: "slide-down" });
	this.props.navigator.switchToTab({ tabIndex: 1 });

In the IOS i can reach to the Tab1 with `dismissAllModals()`, but when the `switchToTab()` is executed i get this error:

	Possible Unhandled Promise Rejection (id: 0):
	Error: could not find UITabBarController
	Error: could not find UITabBarController
	    at createErrorFromErrorData (http://localhost:8081/index.ios.bundle?platform=ios&dev=true&minify=false&hot=true:4584:15)
	    at http://localhost:8081/index.ios.bundle?platform=ios&dev=true&minify=false&hot=true:4542:25
	    at MessageQueue.__invokeCallback (http://localhost:8081/index.ios.bundle?platform=ios&dev=true&minify=false&hot=true:4864:16)
	    at http://localhost:8081/index.ios.bundle?platform=ios&dev=true&minify=false&hot=true:4713:16
	    at MessageQueue.__guard (http://localhost:8081/index.ios.bundle?platform=ios&dev=true&minify=false&hot=true:4800:9)
	    at MessageQueue.invokeCallbackAndReturnFlushedQueue (http://localhost:8081/index.ios.bundle?platform=ios&dev=true&minify=false&hot=true:4712:12)
	    at http://localhost:8081/debuggerWorker.js:71:58

I found an related issue #284 but its too old and not solved yet.

---
### Environment
* React Native Navigation version: 1.1.134
* React Native version: 0.45.1
* Platform(s): IOS
* Device info: Simulator iPhone 5s 10.3 Debug

ISevenSinsI:
Hi neuberoliveira,

It seems to be the issue has to do with `dismissAllModals()`
For us it fixed to to use `dismissModal` instead.

The last post in #284 (The issue you mentioned) explains it has something to do with this function unsetting several controllers.

neuberoliveira:
[...]

Issue #418, opened 2016-11-08
switchToTab stops working after calling dismissAllModals


indivisable:
### Description
The tab switching seems to stop working after you use the dismissAllModals method, I was using this to assure all modals were cleared but it works when I only use dismissModal.

Rel: https://github.com/wix/react-native-navigation/issues/284

### Steps 

1) Use the example code from latest 2.0 branch (Tabbed Application)
2) Create a button that triggers the switchToTab Method
3) Create a Modal and a button to toggle it
4) Dismiss the Modal with dismissAllModals
5) Test the switchTab button 

---
### Environment
react: 15.3.2
react-native: 0.36.0
IOS 10
iPhone 6 - On device.

norman-kapschefsky:
I do have a related issue. After calling `dismissAllModals` I cant push a new screen on the stack that was under that dismissed modal. When using `dismissModal` its fine. **Note** that this happens on **iOS only**.

fuatsengul:
I have the same issue, Is there any workaround at the moment?

ScrimpyCat:
Came across the same issue with dismissAllModals preventing any new screens from being pushed onto the stack underneath. Anyone figured out why dismissAllModals isn't working?

@fuatsengul Only workaround I'm aware of is to call dismissModal the required amount of times (avoid animation on all subsequent dismissals to avoid any weird UI animations being shown to the user). This however is a bit of a hack.

asimolmez:
👍

PR #2093, opened 2017-10-28
Fixed dismissAllModals for iOS

sotownsend:
Fixed dismissAllModals for iOS - Previously it would unregister things like the tab bar that it shouldn't have. #284 #1598 #418


shuse2:
when will this be published on npm?

CherishIt:
could this PR be merged?

jasonmerino:
I too would love to see this merged. It would be more useful for handling deep links that come into the app and we don't want to keep track of how many modal layers deep we are on the off chance that there are two displayed.

sotownsend:
We've been using this in production for a few weeks without any mishaps.

gran33:
Solved in 
https://github.com/wix/react-native-navigation/pull/2255

PR #2097, opened 2017-10-30
V2 overlay skeleton

Cool04ek:


lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/overlay/AlertOverlay.java
```diff
@@ -0,0 +1,33 @@
...
+import android.support.v7.app.AlertDialog;
+
+import com.reactnativenavigation.parse.OverlayOptions;
+
+public class AlertOverlay implements OverlayInterface {
+
+	private AlertDialog dialog;
+
+	@Override
+	public void init(Context context, OverlayOptions options) {
```
DanielZlotin: "new or new-not; there is not try"...

maybe a cleaner way would be to replace the Overlay enum with a factory, with a `create` method, receiving all needed params (as well as context)
then you can in constructor do all the "init" stuff

Cool04ek: Ok. I got your point. I don't want to remove enum, because it helps to have all types in one plays. So, I will try to change getOverlay method in that way, so we create only the one, we need with proper options

DanielZlotin: 👍 
`getOverlay` probably rename to `create` or something.. factory method conventions..

Cool04ek: yeah, sure

PR #2103, opened 2017-10-31
Set modal status bar style in constructor

guyca:
Since status bar style was applied when the screen was created, this caused a noticeable
artifact in status bar color.


koenpunt:
Good that the status bar no longer "flashes" when a modal is opened, but now a new issue appeared; if the content in the modal doesn't render instantaneously the statusbar flies by before the content is shown.

See this video for what I mean: https://www.youtube.com/watch?v=MIcDfcEnwiQ

 As a workaround, it possible to set the background of a modal screen?

PR #2108, opened 2017-11-01
Fixed NavigationBar Background change on Pop

atkit:
I noticed if I use different UINavigationBar backgroundColors it shows navigationBar as white on pop back (for a short period of time, but still noticeable) (and it shows it correctly with drag gesture)

actual behaviour:
bug

fixed, and expected behaviour:
fixed

Any suggestions on solution are welcome

### DIFF
--- a/ios/RCCViewController.m
+++ b/ios/RCCViewController.m
@@ -311,6 +311,23 @@ const NSInteger TRANSPARENT_NAVBAR_TAG = 78264803;
     [self setStyleOnAppearForViewController:self appeared:true];
 }
 
+// the only way I found so far to make it possible to solve the problem with changing navigation bar color.
+// otherwise chaning color of navigation bar is delayed on pop and it is visually distructing
+- (BOOL) isViewLoaded {
+    [self updateNavigationBarTintColor];
+    return [super isViewLoaded];
+}
+
+- (void) updateNavigationBarTintColor {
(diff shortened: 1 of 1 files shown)


ios/RCCViewController.m
```diff
@@ -351,6 +351,19 @@ - (void)updateStyle
   [self setStyleOnAppearForViewController:self appeared:true];
 }
 
+- (BOOL) isViewLoaded {
```
eliperkins: Overriding `isViewLoaded` here to have a side-effect doesn't seem like a safe idea. Access to this property is not guaranteed.

Can you set a breakpoint in this method and figure out what lifecycle we should be hooking into (aka, the one that's calling this)?

It looks like we run this logic already on `viewWillAppear` and `viewDidAppear`, but perhaps we need to call this on `viewDidLoad` as well.

atkit: agree, I tried different techniques, but this the only one worked for me. 
But I'll take a look a bit more...

yershalom:

atkit:
@yershalom I agree that solution is not optimal, but I'm using it in production (via own fork). I did not find better solution so far. And version 1.x is not a priority now. I hope to switch to 2.x soon.

atkit:
rebased and tried other things, but nothing seem working better in a way. 

open for suggestions. 

P.S. as I maintain own fork and use it in the project for a half a year already I did not notice any issues with this solution yet.

PR #2119, opened 2017-11-05
[v2] [iOS] Add subtitle settings and styling

krystofcelba:
Hi, there! I would like to help a bit with this great project. I first picked the simpler functionality like subtitle settings. I hope, there will be everything OK with that.


yershalom:

PR #2133, opened 2017-11-08
[iOS] Add support for bar button system items

eliperkins:
UIKit supplies a range of glyphs to be used in navigation bar buttons, so let's expose them to developers to make better native apps!

This adds support for the `systemItem` key-value in the button config to allow for placing this type of `UIBarButtonItem`s in the navigation bar.

More information about these glyphs can be found in [Apple's Human Interface Guidelines](https://developer.apple.com/ios/human-interface-guidelines/icons-and-images/system-icons/).


guyca:
Hey @eliperkins, is this pr compatible with xcode8?

eliperkins:
Yup. Nothing specific to Xcode 9 here, it should work fine in Xcode 8.

smooJitter:
Can you provide an example of how to access the bar button system item

PR #2135, opened 2017-11-08
Fixed correct key name

MrVisitor:
navBarTitleFontSize - is deprecated key. Now this key is not functioning.


MrVisitor:
What do you say about this PR?

yershalom:

PR #2136, opened 2017-11-08
Added navBarTextFontBold to iOS

MrVisitor:
Android has navBarTextFontBold property. I added this property to iOS

### DIFF
--- a/android/app/src/main/java/com/reactnativenavigation/layouts/BottomTabsLayout.java
+++ b/android/app/src/main/java/com/reactnativenavigation/layouts/BottomTabsLayout.java
@@ -380,12 +380,17 @@ public class BottomTabsLayout extends BaseLayout implements AHBottomNavigation.O
 
     @Override
     public void pop(final ScreenParams params) {
-        getCurrentScreenStack().pop(params.animateScreenTransitions, params.timestamp, new ScreenStack.OnScreenPop() {
+        performOnStack(params.getNavigatorId(), new Task<ScreenStack>() {
             @Override
-            public void onScreenPopAnimationEnd() {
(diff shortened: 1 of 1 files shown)


MrVisitor:
What do you say about this PR?

ios/Helpers/RCCTitleViewHelper.m
```diff
@@ -194,10 +194,17 @@ -(UILabel*)setupTitle:(NSDictionary*)style
...
     
-    id fontSize = style[@"navBarTitleFontSize"];
+    id fontSize = style[@"navBarTextFontSize"];
+    id fontBold = style[@"navBarTextFontBold"];
+    
     if (fontSize) {
         CGFloat fontSizeFloat = [RCTConvert CGFloat:fontSize];
-        titleFont = [UIFont boldSystemFontOfSize:fontSizeFloat];
+        
+        if (fontBold == 1) {
```
eliperkins: I'm not sure I understand what this check is for. Do you mean `if (fontBold)` to indicate that the property is non-null?

ios/Helpers/RCCTitleViewHelper.m
```diff
@@ -194,10 +194,17 @@ -(UILabel*)setupTitle:(NSDictionary*)style
     
     UIFont *titleFont = [UIFont boldSystemFontOfSize:17.f];
     
-    id fontSize = style[@"navBarTitleFontSize"];
```
eliperkins: This change would be a breaking change for all API users. We should not break the API, especially for something like this, because it would break our users' apps.

yershalom:

PR #2140, opened 2017-11-09
increase nav bar custom icons tap area

Cool04ek:


ios/RCCNavigationController.m
```diff
@@ -406,13 +406,16 @@ -(void)setButtons:(NSArray*)buttons viewController:(UIViewController*)viewContro
     }
   }
   
+  UIBarButtonItem *itemToIncreaseInsets = [barButtonItems lastObject];
```
eliperkins: Why does this call `lastObject`? I don't understand why we only need to change insets for the last item.

Cool04ek: In other case the previous ones insets may overlap with the next ones

eliperkins:
I'm not sure I understand why this change is needed. Can you elaborate more as to what bug this is fixing?

Cool04ek:
It's not a bug fix. It's feature, asked by product manager.

eliperkins:
This doesn't seem like much of a feature if it's something that will affect all consumers of this library, regardless of product. It would be great if we could figure out what the feature is that you're trying to add in and find a way to get it in that is backwards compatible for other users.

Cool04ek:
Basically, it increases tap area of buttons. It's that simple

guyca:
@Cool04ek Can we make this change opt in?

PR #2141, opened 2017-11-09
Fixing iOS installation steps

douglasjunior:
Problem:
captura de tela 2017-11-09 as 09 29 00

### DIFF
--- a/docs/installation-ios.md
+++ b/docs/installation-ios.md
@@ -14,7 +14,7 @@
 
 4. In Xcode, in Project Navigator (left pane), click on your project (top), then click on your *project* row (on the "project and targets list") and select the `Build Settings` tab (right pane). In the `Header Search Paths` section add `$(SRCROOT)/../node_modules/react-native-navigation/ios`. Make sure on the right to mark this new path `recursive` ([screenshots](https://facebook.github.io/react-native/docs/linking-libraries-ios.html#step-3))
 
-5. In Xcode, you will need to edit this file: `AppDelegate.m`. Is the main entry point for your app. Its content must be replaced with the content of this [reference](https://github.com/wix/react-native-navigation/blob/master/example/ios/example/AppDelegate.m).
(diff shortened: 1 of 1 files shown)


docs/installation-ios.md
```diff
@@ -14,9 +14,8 @@
 
 4. In Xcode, in Project Navigator (left pane), click on your project (top), then click on your *project* row (on the "project and targets list") and select the `Build Settings` tab (right pane). In the `Header Search Paths` section add `$(SRCROOT)/../node_modules/react-native-navigation/ios`. Make sure on the right to mark this new path `recursive` ([screenshots](https://facebook.github.io/react-native/docs/linking-libraries-ios.html#step-3))
 
-5. In Xcode, you will need to edit this file: `AppDelegate.m`. 
-This function:
-````````
-- (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions { ... }
-````````
-is the main entry point for your app. Its content must be replaced with the content of this [reference](https://github.com/wix/react-native-navigation/blob/master/example/ios/example/AppDelegate.m)
+5. In Xcode, you will need to edit this file: `AppDelegate.m`. Is the main entry point for your app. Its content must be replaced with the content of this [reference](https://github.com/wix/react-native-navigation/blob/master/example/ios/example/AppDelegate.m).
```
eliperkins: What do you think about wording this as `In Xcode, you'll need to edit the file `AppDelegate.m`. It is the main entry point for your app.`?

As it's currently written, the second sentence is an incomplete sentence.

douglasjunior: Thanks for review.

I removed the code that was in the middle of the text to make it easier to format. In the second sentence I refer to the change described at the first sentence. But this really was confusing.

Maybe the following is better?

5. In Xcode, you will need to edit the function `didFinishLaunchingWithOptions` in the `AppDelegate.m`. Its content must be replaced with the content of this [reference](https://github.com/wix/react-native-navigation/blob/master/example/ios/example/AppDelegate.m). 
   ```objc
   - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions { ... }
   ```

eliperkins: Looks good to me!

yershalom:

PR #2156, opened 2017-11-13
Fixing a crash upon rotating from locked landscape orientation to portrait

lelikg:
The crash is:

*** Terminating app due to uncaught exception 'UIApplicationInvalidInterfaceOrientation', reason: 'Supported orientations has no common orientation with the application, and [RCCTabBarController shouldAutorotate] is returning YES'

Setting shouldAutorotate to return NO fixes the crash

### DIFF
--- a/ios/RCCTabBarController.m
+++ b/ios/RCCTabBarController.m
@@ -413,5 +413,8 @@
 }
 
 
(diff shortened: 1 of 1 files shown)


lelikg:
hi guys, anyone merging this? :)

PR #2172, opened 2017-11-16
Transparent splash background color

ywongweb:
Add the ability to set the splash activity's background color to transparent. Defaults to white as it is now.

Add the following to AndroidManifest.xml
```
        <meta-data
                android:name="splashBackgroundColor"
                android:value="transparent"
        />
```

#### Notes:  
Not sure if setting this in AndroidManifest is the best way but I can't get to the styleParams because react context is not yet initialise?. All suggestions will be appreciated.

### DIFF
--- a/android/app/src/main/java/com/reactnativenavigation/controllers/SplashActivity.java
+++ b/android/app/src/main/java/com/reactnativenavigation/controllers/SplashActivity.java
@@ -94,7 +94,18 @@ public abstract class SplashActivity extends AppCompatActivity {
      */
     public View createSplashLayout() {
         View view = new View(this);
-        view.setBackgroundColor(Color.WHITE);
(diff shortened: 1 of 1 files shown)


guyca:
Hey @ywongweb 
Thanks for the PR, This is an interesting approach. Changing the background color can be done by [overriding](https://wix.github.io/react-native-navigation/#/android-specific-use-cases?id=splash-screen) `createSplashLayout` in your `MainActivity` and returning a transparent view. For example:

```java
@Override
public View createSplashLayout() {
    View transparentView = new View(this);
    transparentView.setBackgroundColor(Color.TRANSPARENT);
    return transparentView;
}
```

We're trying not to add new API's which will make the transition to v2 difficult, since the current api allows setting a transparent background and v2 already supports the same API (more or less), I think its best we don't add this api.

PR #2176, opened 2017-11-16
Fix invalid props access in componentWillReceiveProps

EdenGottlieb:


guyca:
resolved in #2179, Thanks @EdenGottlieb

PR #2186, opened 2017-11-19
3D Touch Preview API

birkir:
Added the preview and pop feature.

It is currently implemented so the javascript API is untouched (uses passProps) but I would like to see if anyone has interest in this feature so we can give it some API love.

To use the feature you simply call `navigator.push()` but with additional parameter `previewViewID` which is a React Node.

```js
<TouchableOpacity
  onPressIn={() => this.props.navigator.push({
    screen: 'screenid',
    previewView: this.viewRef,
    previewActions: [{
      title: 'Foo',
      style: 'selected', // none, selected, destructive
      actions: [{ title: 'Bar' }],
    }],
 })
/>
```
previewandpop

Going forward with this would include some of the next steps:
 - [x] Move previewViewID from passProps to it's parent.
 - [x] Added some fail-safe wrappers around the added class methods
 - [x] Added preview actions (with groups).
 - [x] Moved away from passProps to actionParams.
 - [x] Added action navigator events.
 - [x] Conditionally commit to pop on deep press.
 - [x] Option to set explicit height for the preview on shallow press.
 - [x] Documentation


messense:
This is awesome!

ios/RCCViewController.h
```diff
@@ -27,6 +27,8 @@ extern NSString* const RCCViewControllerCancelReactTouchesNotification;
 - (void)setStyleOnInit;
 - (void)updateStyle;
 - (void)setNavBarVisibilityChange:(BOOL)animated;
+- (void)setPreviewController:(RCCViewController *)controller;
```
eliperkins: This should be implemented as a `@property`, rather than explicit getters and setters with private ivars.

ios/RCCNavigationController.m
```diff
@@ -161,6 +162,19 @@ - (void)performAction:(NSString*)performAction actionParams:(NSDictionary*)actio
       [self setButtons:rightButtons viewController:viewController side:@"right" animated:NO];
     }
     
+    NSString *previewViewID = passProps[@"previewViewID"];
+    if (previewViewID) {
+      RCCViewController *topViewController = ((RCCViewController*)self.topViewController);
```
eliperkins: This seems a little scary. Do we have a way to guarantee that this view controller will be a `RCCViewController`? Perhaps we should wrap this cast with a check to see if the view controller responds to the specific selector that we're calling on it (`@selector(setPreviewController:)`).

birkir: Yes you're right, in the `if(keepStyleAcrossPushBool)` statement above it is  actually doing an `if ([self.topViewController isKindOfClass:[RCCViewController class]])`.

I will incorporate this, which one would you prefer?

ios/RCCViewController.h
birkir: Thanks!

ios/RCCNavigationController.m
eliperkins: Either `isKindOfClass` or `respondsToSelector` is fine with me!

birkir:
> So, as I understand it, we need to have already rendered our previewView somewhere in React, since we need to pass along the node handle in passProps? How will this work with cases where we want to preview a different view than one we already have rendered (say, a new view controller)?

1. `previewViewID` is actually a react view that will have the 3D touch enabled on it. When you press it it will blur the background out (Image 1)
2. When you press it harder, it will show the view controller given in the push configuration (Image 2)
3. And if you press even harder, it will "pop" the view controller as a newly pushed screen (Image 3)

sample

I like to think of it as a different way to `navigator.push`.

birkir:
> So, as I understand it, we need to have already rendered our previewView somewhere in React, since we need to pass along the node handle in passProps? How will this work with cases where we want to preview a different view than one we already have rendered (say, a new view controller)?

Thank you for the feedback!

1. `previewViewID` is actually a react view that will have the 3D touch enabled on it. When you press it it will blur the background out (Image 1)
2. When you press it harder, it will show the view controller given in the push configuration (Image 2)
3. And if you press even harder, it will "pop" the view controller as a newly pushed screen (Image 3)

sample

I like to think of it as a different way to `navigator.push`.

brunolemos:
Awesome!

API love would be great.
Ideally it should receive any normal component instead of an id, no?
But I can imagine it could lag if it wasn't already rendered.

Also add support for action buttons:

birkir:
Yes, that is what the Preview component would just do behind the scene (get the ref etc.)

It must be rendered before the force touch is started.

Actions is good and easy addon, but providing extensive api for it would require some work (like click action and show another set of actions)

Example of this is in the Apollo app, force press an item, click share, and you will be printed with another set of actions (share image / share link to image)

Use cases for this

birkir:
- [x] Added some fail-safe wrappers around the added class methods
 - [x] Added preview actions (with groups).
 - [x] Moved away from passProps to actionParams.
 - [x] Added action navigator events.

I am actually thinking about moving away from providing specific component that will act as previewViewID. The preview view component is only used to create and measure a rectangle on the screen to *not* blur on force press. We could remove the need for providing the actual react id and just provide the ref by handling `findNodeHandle` internally.

What do you think?

messense:
+1 on handling `findNodeHandle` internally, `findNodehandle` API might not be as well known as `ref`.

birkir:
Ok. Then I think we are good to go, at least to begin with. We could iterate on this if we get some feedback from people.

src/deprecated/platformSpecificDeprecated.ios.js
```diff
@@ -232,7 +234,11 @@ function navigatorPush(navigator, params) {
     console.error('Navigator.push(params): params.screen is required');
     return;
   }
+  let previewViewID;
   const screenInstanceID = _.uniqueId('screenInstanceID');
+  if (params.previewView instanceof Component) {
+    previewViewID = findNodeHandle(params.previewView)
+  }
```
messense: I think it should have an `else` branch to treat `previewView` as `previewViewID` directly?

brunolemos:
Looking great!

Did you test this two cases:

- The preview component will probably be a View hidden in the screen. How should the user proceed? Hide it's parent with height: 0 or what?
- Above the example above, did you test with dynamic content? E.g. If I force touch Row1 it will show Photo1 and if I force touch Row2 it will show Photo2 inside the same preview component

birkir:
1. I don't think it will. The previewView will always be something already visible on the screen (like a Row or a picture cell like in the GIF's above).

2. The content of the preview is always the screen you defined in the navigator.push (with all its settings). The screen will be immediately rendered in the background and transitioned into an preview based on the force of your touch.

With our current implementation we can replicate every app I've seen yet that has this feature.

Only limitation is the preview height (Like the image preview), but once we implement showLightbox, the preview controller should be able to use that height.

Also we need to implement a configurable boolean flag to skip the actual push on hard press.

Does this make sense to you?

src/deprecated/platformSpecificDeprecated.ios.js
birkir: Great idea, thanks!

birkir:
Ok after reading your second point again I realized that this could actually be done, but as to my understanding we don't have any easy way to update props of an RCCViewController yet?

But it should be easy to check if the previewController is is the same as the proposed one and just re-use it and update its styles and props.

ios/RCCViewController.m
```diff
@@ -814,4 +845,32 @@ -(BOOL)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer shouldRecogniz
   return !disabledSimultaneousGestureBool;
 }
 
+#pragma mark - UIViewControllerPreviewingDelegate
+- (UIViewController *)previewingContext:(id<UIViewControllerPreviewing>)previewingContext viewControllerForLocation:(CGPoint)location {
+  return self.previewController;
+}
+
+- (void)previewingContext:(id<UIViewControllerPreviewing>)previewingContext commitViewController:(UIViewController *)viewControllerToCommit {
+  [self.navigationController pushViewController:self.previewController animated:false];
```
birkir: Here we may want to provide the user to skip this operation if he doesn't want the screen to be pushed.

brunolemos:
> The previewView will always be something already visible on the screen

Ok sorry, I had misunderstood the API. I thought `previewView` was the small preview view that shows while you are force touching. But it seems to be the screen that will be pushed, right?

So I'll rephrase my question. Let's suppose this use case:

While you force touch on Instagram (peek), it shows a minimal preview screen with just 3 components: the profile photo, the profile name and the photo. When you finish force touching (pop), it shows the full page with all other contents (follow button, like button, etc).

Is this case supported?
By your example I get the idea that it shows the full screen with the navigation bar.

<div>


</div>

If not, one possible solution is to pass a prop to the screen saying that it's on the preview state.

birkir:
Gotcha! No problem.

We can pass a prop saying it's a preview and then send ScreenChanged event once it's pushed (didPreview etc.)

For the instagram case, what is really happening is just cropping + conditional rendering. How they do the height thing, IDK.

Maybe it would just work by setting fixed height to the first View inside the screen. I will verify and come back to you.

birkir:
Allright guys, I have tested all cases discussed in the thread. I added the option conditionally commit to pop the controller and to set explicit height to the preview with the recommended `preferredContentSize` suggested by Apple.

Please let me know if you have any more concerns about this PR, I'd like to get it merged to get some feedback from real users and see if anybody will find an limitation with the current implementation.

guyca:
@birkir This looks amazing. Lets get this merged asap so we can give more feedback.

1. I'm getting console.error on each push that doesn't have a [previewView](https://github.com/wix/react-native-navigation/pull/2186/files#diff-412ba139e8c40abd1d18f5a61eef9d89R246).
2. Is this feature compatible with xCode < 9 ?
3. Could you please add this to the [docs](https://github.com/wix/react-native-navigation/blob/master/docs/screen-api.md) so it's visible to more users? I'd hate if people miss out on this.

Also, thanks a lot @eliperkins @messense and @brunolemos for reviewing this pr 👍

birkir:
@guyca Thanks!

1. Whoops, fixed.
2. Yessir
3. I already documented configuration options, but also added a specific section for this explaining better how it works with a link to Apple's docs.

messense:
Tried this today, works great with single `previewRef`.
But I can't get it work with `FlatList` items (I have stored all item `ref`s in `this._refs`), not sure what I did wrong. I could see it pushed the screen with `isPreview = true` in logs but the preview didn't show up.

birkir:
Can you share an example code? Are you doing the push onPressIn?

enahum:
So I've been trying this today and it seems that it does not work for me

here is some code to show what I'm doing

```js
onPreview = () => {
        this.props.navigator.push({
            screen: 'About',
            title: 'New Screen',
            previewCommit: false,
            previewHeight: 250,
            previewView: this.previewRef,
            previewActions: [{
                id: 'action-cancel',
                title: 'Cancel'
            }, {
                id: 'action-delete',
                title: 'Delete',
                actions: [{
                    id: 'action-delete-sure',
                    title: 'Are you sure?',
                    style: 'destructive'
                }]
            }]
        });
    };
```

in my render function I have
```js
return (
            <TouchableHighlight
                underlayColor={changeOpacity(theme.sidebarTextHoverBg, 0.5)}
                onPress={this.onPress}
                onPressIn={this.onPreview}
            >
                <View
                    style={style.container}
                    ref={this.setPreviewRef}
                >
                    {extraBorder}
                    <View style={[style.item, extraItemStyle]}>
                        {icon}
                        <Text
                            style={[style.text, extraTextStyle]}
                            ellipsizeMode='tail'
                            numberOfLines={1}
                        >
                            {displayName}
                        </Text>
                        {badge}
                    </View>
                </View>
            </TouchableHighlight>
        );
```

this is what I get in the logs
```json
Running application About ({
    initialProps =     {
        commandType = Push;
        isPreview = 0;
        navigatorEventID = "screenInstanceID5_events";
        navigatorID = "controllerID1_nav";
        screenInstanceID = screenInstanceID5;
        timestamp = 1511532016454;
    };
    rootTag = 31;
})
index.bundle:24363 Running application "About" with appParams: {"rootTag":31,"initialProps":{"navigatorID":"controllerID1_nav","timestamp":1511532016454,"commandType":"Push","screenInstanceID":"screenInstanceID5","isPreview":false,"navigatorEventID":"screenInstanceID5_events"}}
```

and I have to point out that onPressIn is executed right away.

RN: 0.49.5
RNN: 1.1.295
Testing in simulator and in real device

any help would be appreciated

enahum:
So I made it work, but if I press touch and the preview shows and then I just let it go, onPress still fires, is there a way to prevent the onPress from being triggered?

birkir:
Good to hear, I have to experiment more with variations of press events, I know that onPressIn is the absolute first event in the sequence so I think that is fine, but it may interfere with cancelled preview.

rgoldiez:
@birkir - This is really awesome.  Thank you for your work on this.  However, like @enahum I'm seeing the `onPress` fire when the preview shows and then is dismissed from the `onPressIn`.  I suppose this is expected since `onPressIn` fires, followed by `onPress`, followed by `onPressOut`.  I'm using `onLongPress` for now and it seems to work well!

rgoldiez:
@birkir - Does the `previewView` need to be an instance of a `Component`?  I get an error if referencing a `View`?  So, in the case where I wanted to reference a `View`, I simply changed my component from a `View` to an `Animated.View` (even though I'm not animating it) and it worked just fine.

rgoldiez:
@birkir - so I’ve been playing around with this new functionality and it’s super easy to use. One thing that would be nice to have is a prop that goes true when a preview has been committed. I’m currently listening for the event, which works as expected, but if I want to conditionally block rendering of a component when in preview mode, I need to tract state for when a preview gets committed. Not a huge deal but this could be a small improvement that makes use even easier.

brunolemos:
> One thing that would be nice to have is a prop that goes true when a preview has been committed. I’m currently listening for the event, which works as expected, but if I want to conditionally block rendering of a component when in preview mode, I need to tract state for when a preview gets committed

Yep that's exactly what I meant with:

> pass a prop to the screen saying if it's on the peek/preview state or not. Another way would be by using events (might be boring to always have to listen to it and set the state accordingly

rgoldiez:
@brunolemos - Can you explain what you mean?

brunolemos:
@rgoldiez I meant: "+1"

birkir:
Cool thanks for the feedback guys. Can we update screen props somehow dynamically? I don't think this is done anywhere in RNN at the moment.

I know it's a hassle to add a listener and a state mechanism, but it's very common pattern with RNN in general (didAppear+didDisappear).

enahum:
@birkir lets say I have a list of components that onLongPress will do the 3D touch peek&pop and it looks like is working, while doing the peek I can see how it sometimes shows me a peek with the props that I've passed to a previews peek.

for instance:
long press component A -> shows peek of A
long press component B -> shows peek of B
long press component A -> shows peek of B

have you or anyone else seen this?

enahum:
from the logs

```
Running application ChannelPeek ({
    initialProps =     {
        channelId = 775qkerugp8cmk7cbx1r1em8sa;
        commandType = Push;
        isPreview = 1;
        navigatorEventID = "screenInstanceID5_events";
        navigatorID = "controllerID1_nav";
        previewViewID = 153;
        screenInstanceID = screenInstanceID5;
        timestamp = 1511806967339;
    };
    rootTag = 31;
})

Running application ChannelPeek ({
    initialProps =     {
        channelId = bkzo93oum38b8p3bzgspyguyyy;
        commandType = Push;
        isPreview = 1;
        navigatorEventID = "screenInstanceID6_events";
        navigatorID = "controllerID1_nav";
        previewViewID = 164;
        screenInstanceID = screenInstanceID6;
        timestamp = 1511806971471;
    };
    rootTag = 41;
})

Running application ChannelPeek ({
    initialProps =     {
        channelId = 775qkerugp8cmk7cbx1r1em8sa;
        commandType = Push;
        isPreview = 1;
        navigatorEventID = "screenInstanceID7_events";
        navigatorID = "controllerID1_nav";
        previewViewID = 153;
        screenInstanceID = screenInstanceID7;
        timestamp = 1511806975328;
    };
    rootTag = 51;
})
```

first preview = OK
second preview = OK
third preview = FAIL (this one shows the content of the previewViewID = 164 instead of 153

birkir:
Thanks. It seems that If you do a peek and then cancel quickly, the next peeking screen will actually be the top view controller and I think that is what is happening for you.

I will need some help debugging why this is happening, as apple says that the previewing api will handle deallocating the previewing view controller, but doesn't seem to be the case.

Maybe it would work to simply set the previewController to nil after successful push. I'll investigate but any help would be greatly appreciated.

enahum:
I really don't know much about object-C but I was looking at the docs and I'm wondering if you ever do `unregisterForPreviewing`?

enahum:
an example of how to use it https://github.com/alexhillc/AXPhotoViewer/blob/master/Example/AXPhotoViewerExample/TableViewController.swift#L62 dunno if this helps

birkir:
Yes, Apple's docs say:

> The system calls this method automatically when a 3D Touch-registered view controller is deallocated.
> In some circumstances, you must explicitly call this method. This is the case if a registered view controller’s view hierarchy changes state, or if presenting a preview is otherwise no longer possible. In such cases, call this method.

https://developer.apple.com/documentation/uikit/uiviewcontroller/1621395-unregisterforpreviewing

But yes, we may have to call it manually.

birkir:
@enahum https://github.com/wix/react-native-navigation/pull/2239

niktonic21:
@messense Did you run preview with FlatList? I have same issue. With single page its ok.

messense:
@niktonic21 Yes, I did run preview with `FlatList`.

birkir:
@messense @niktonic21 We are still waiting for the fix to be merged. It is ready!

guyca:
Hey guys, sorry about the delay. Merged now, please ping me on discord if I'm not responsive here 👍

anonrig:
Hello, does anybody have a working example of this preview 3d touch feature?

birkir:
https://github.com/wix/react-native-navigation/blob/master/example/src/screens/NavigationTypes.js#L132-L138

charliesbox:
Hi, I'm seeing a weird case. if you peek a view several times (without popping it), and then you actually decide to pop a view, it will do the animation, but it won't show the new view. Even if you just press the view, it won't go to the "details" view.

anonrig:
@charliesbox I face the same issue.

charliesbox:
@anonrig I opened an issue to track it better

https://github.com/wix/react-native-navigation/issues/2445

anonrig:
@charliesbox Can you also add my gif animation to the issue?

charliesbox:
@anonrig sure thing 👍

harisb2012:
What happened with this guys?

milesscherrer:
What happened with this?

PR #2188, opened 2017-11-19
addOnNavigatorEvent: defensive try-catch wrapping for handlers

yonigross:
When iterating over event handlers, it's possible one of them is in inactive state, or will fail for some reason. We do not want to terminate execution for the rest of the handlers queue in such a case.

### DIFF
--- a/src/Screen.js
+++ b/src/Screen.js
@@ -181,7 +181,13 @@ class Navigator {
     if (this.navigatorEventHandler) {
       this.navigatorEventHandler(event);
     }
-    this.navigatorEventHandlers.forEach(handler => handler(event));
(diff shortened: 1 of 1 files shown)


guyca:
Hey @yonigross 
What do you mean by "inactive state"?

yonigross:
Since we are stacking event handlers on the same navigator, one of the components added an event might be in an invalid or inactive state while an event fires (because it cannot predict at which state it will be called). For an example, let's say the 'screen' component listen to 'save button clicked' event (because it need to save), while some other component listen to the same event to do something else (e.g show some loader), but this component is unaware that the event can happen when it's not expecting that, and might cause an exception. We don't want such an exception to prevent the rest of the event handlers from executing.

PR #2189, opened 2017-11-19
(Android) Allow disableOpenGesture right or left in the drawer

Fuhrmann:
Hello! This is my first pull request here so apologize if I made any mistakes. I'm working on my first React  Native project and thought I could solve this problem that can happen to other people.

This pull request allows you to disable the open gesture only in one side of the drawer. In my current application, I want to have 2 drawers (on both sides). The left drawer is always enabled. The right drawer is only enabled in one screen.

This way I can startSingleScreenApp() and pass the two drawers, but only the left would be 'enabled':

```js
      screen: initialScreen,
      drawer: {
        left: {
          screen: DRAWER_SCREEN,
          disableOpenGesture: false,
        },
        right: {
          screen: SIDEBAR_FILTER_SCREEN,
          disableOpenGesture: true,
        },
      },
    });
```

If we pass disableOpenGesture to the drawer root object, like this: 

```js
      screen: initialScreen,
      drawer: {
        left: {
          screen: DRAWER_SCREEN,
          disableOpenGesture: false,
        },
        right: {
          screen: SIDEBAR_FILTER_SCREEN,
          disableOpenGesture: true,
        },
        disableOpenGesture: true <======
      },
    });
```

Then the two drawer positions would be gesture disabled and the specific disableOpenGesture would be ignored.

What do you guys think?

### DIFF
--- a/docs/top-level-api.md
+++ b/docs/top-level-api.md
@@ -104,10 +104,12 @@ Navigation.startSingleScreenApp({
     left: { // optional, define if you want a drawer from the left
       screen: 'example.FirstSideMenu', // unique ID registered with Navigation.registerScreen
       passProps: {}, // simple serializable object that will pass as props to all top screens (optional)
+      disableOpenGesture: false // optional, can the drawer be opened with a swipe instead of button
     },
     right: { // optional, define if you want a drawer from the right
       screen: 'example.SecondSideMenu', // unique ID registered with Navigation.registerScreen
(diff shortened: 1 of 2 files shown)


guyca:
Thanks for the PR Ricardo, I'll get to it by the end of the week. Can you please update the [docs](https://github.com/wix/react-native-navigation/blob/master/docs/top-level-api.md) as well?

Fuhrmann:
Hey @guyca, I`ve just updated the docs!

PR #2204, opened 2017-11-21
Revert "Support iOS 11 prefersLargeTitles #1643"

gran33:
Reverts wix/react-native-navigation#2090


anonrig:
Why did we revert this?

Issue #2183, opened 2017-11-18
[v2] test-unit-ios failed


vovkasm:
### Issue Description

Seems like I found a bug non-related to this issue, but when I try to dig into it, I discover that ios tests fails on my system.
Specifically it is `RNNCommandsHandlerTest/testAssertReadyForEachMethodThrowsExceptoins` that failed. I'am can't understand what is intent of this test, so this issue is more question to developers. This test try call all methods of `RNNCommandsHandler` class, but without needed arguments... is this possible in Objective C at all? And what it try to check after all?

### Steps to Reproduce / Code Snippets / Screenshots

As in https://github.com/wix/react-native-navigation/blob/v2/docs/docs/CONTRIBUTING.md
1. `brew tap wix/brew`
2. `brew install --HEAD applesimutils`
3. `npm install`
4. `npm run test-js`
5. `npm run test-unit-ios`

Expected that tests pass, but instead this output:
```
2017-11-18 16:33:58.231 xcodebuild[48492:1624107] [MT] IDETestOperationsObserverDebug: (071EFBDB-4403-4669-912B-6C0CCDDFA608) Beginning test session ReactNativeNavigationTests-071EFBDB-4403-4669-912B-6C0CCDDFA608 at 2017-11-18 16:33:58.231 with Xcode 9B55 on target <DVTiPhoneSimulator: 0x7fa7eeb0ede0> {
		SimDevice: iPhone 7 (7E09DA93-0546-4163-9F2F-94A78EEE432B, iOS 11.1, Booted)
} (11.1 (15B87))

Restarting after unexpected exit or crash in RNNCommandsHandlerTest/testAssertReadyForEachMethodThrowsExceptoins; summary will include totals from previous launches.

Test Suite 'Selected tests' started at 2017-11-18 16:34:00.463
Test Suite 'ReactNativeNavigationTests.xctest' started at 2017-11-18 16:34:00.464
Test Suite 'RNNCommandsHandlerTest' started at 2017-11-18 16:34:00.464
Test Case '-[RNNCommandsHandlerTest testDynamicStylesMergeWithStaticStyles]' started.
Test Case '-[RNNCommandsHandlerTest testDynamicStylesMergeWithStaticStyles]' passed (0.031 seconds).
Test Suite 'RNNCommandsHandlerTest' failed at 2017-11-18 16:34:00.496.
	 Executed 2 tests, with 1 failure (0 unexpected) in 0.031 (0.032) seconds
... etc...
[...]

PR #2210, opened 2017-11-23
Fix failing test

vovkasm:
Address issue  #2183

I still do not understand why original implementation works on CI (and failed in my environment), but think that this implementation slightly cleaner because use more high-level APIs.

Tested locally with `npm run test-unit-ios`, `npm run test-unit-ios -- --release` and from Xcode


vovkasm:
Oops, I rebased this request to latest v2 and now tests ok, but why it tried to release from pr? ))

DanielZlotin:
Thanks for the PR.
It didn't try to publish from PR, this is protected by an env check.

vovkasm:
Ah, ok then. I was misled by the fact that deploy stage runs unconditionally.

PR #2212, opened 2017-11-23
Prevent push spam clicking

developer239:
Inspired by [53fdbe6](https://github.com/wix/react-native-navigation/commit/53fdbe6315850a9167f5f2f06f68998395292a5f)

This was a major problem on slower devices. :)

### DIFF
--- a/src/Screen.js
+++ b/src/Screen.js
@@ -7,6 +7,8 @@ import {
 } from 'react-native';
 import platformSpecific from './deprecated/platformSpecificDeprecated';
 import Navigation from './Navigation';
+import _ from 'lodash';
+
 
 const NavigationSpecific = {
   push: platformSpecific.navigatorPush,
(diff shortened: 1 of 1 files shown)


yershalom:

filipef101:
works fine, but can you make it for other things than push?

developer239:
@filipef101 I can grant you access to the repository if you want. :) I can't get to this before Sunday. :/

filipef101:
I solved this by throtling my buttons, just commented as this doesnt fully solve the problem afaik
Also this won't be accepted as it is not a react native or react native navigation problem but multithreading ;)
At most I think it could be something optional. Also this library is mostly dead, and the issues... At least if one's doing a multiplatform app, so many pains in the arse

developer239:
@filipef101 The sad thing is that this is still one of the most useful react-native navigations today. 😄

filipef101:
yeah, sad indeed, but wix or wix devs are not to blame IMO
I think v2 will solve everything hopefully :)

src/Screen.js
```diff
@@ -177,6 +187,17 @@ class Navigator {
     }
   }
 
+  _checkLastAction(params) {
+    if (Date.now() - this._lastAction.timestamp < 800
```
no23reason: How did you choose this value? What if the pushed screen takes longer to render?

Semi related: is there a real scenario where pushing the same screen multiple times was actually desired? I mean, would it break something if you omitted the timestamp altogether?

PR #2213, opened 2017-11-23
Add iPhone X launch image

SabatinoMasala:
This will add the iPhone X launch screen if found in the LaunchImages. I wasn't sure about the **Default** naming convention though, can someone confirm that this would be `Default-2436h`?


jasonmerino:
Would be really cool if this could get merged so we can avoid that black screen in between the launch image and the JS code. 🙏🏼

yogevbd:
fixed in #2321

PR #2224, opened 2017-11-25
setStyle to style the according TabBar

danielang:
Use `navigator.setStyle({...}` to set the style of TabBar of the according screen at runtime for iOS and Android.
Good for theming purpose.

Exampe:
```javascript
this.props.navigator.setStyle({
        tabBarButtonColor: '#0d0d0d',
        tabBarSelectedButtonColor: '#ff6c33',
        tabBarBackgroundColor: '#ffffff',
        navigationBarColor: '#ffffff',
});
```

### DIFF
--- a/android/app/src/main/java/com/reactnativenavigation/views/BottomTabs.java
+++ b/android/app/src/main/java/com/reactnativenavigation/views/BottomTabs.java
@@ -88,7 +88,7 @@ public class BottomTabs extends AHBottomNavigation {
 
     private boolean hasTabsWithLabels() {
         for (int i = 0; i < getItemsCount(); i++) {
-            String title = getItem(0).getTitle(getContext());
+            String title = getItem(i).getTitle(getContext());
             if (!TextUtils.isEmpty(title)) {
                 return true;
             }
--- a/android/app/src/main/java/com/reactnativenavigation/views/BottomTabs.java
+++ b/android/app/src/main/java/com/reactnativenavigation/views/BottomTabs.java
@@ -26,7 +26,7 @@ public class BottomTabs extends AHBottomNavigation {
         setForceTint(true);
         setId(ViewUtils.generateViewId());
         createVisibilityAnimator();
-        setStyle();
+        setStyle(AppStyle.appStyle);
         setFontFamily();
     }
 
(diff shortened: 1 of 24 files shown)


android/app/src/main/java/com/reactnativenavigation/views/BottomTabs.java
```diff
@@ -132,23 +127,52 @@ private void createVisibilityAnimator() {
...
+        }
+        if (hasSelectedBottomTabsButtonColor(params)) {
+            if (getAccentColor() != params.selectedBottomTabsButtonColor.getColor()) {
+                setAccentColor(params.selectedBottomTabsButtonColor.getColor());
+            }
+        }
+    }
+
+    private boolean hasBadgeTextColor(StyleParams params) {
+        return params.bottomTabBadgeTextColor != null &&
```
guyca: We can avoid adding these convince method by initializing these style properties so they are never `null` [here](https://github.com/wix/react-native-navigation/blob/master/android/app/src/main/java/com/reactnativenavigation/params/StyleParams.java#L143). The `Color` class is meant to be used as a null object

danielang: Of cause. This was original coding, so I've adapted it to match this code style.
Which color will be the default one?

guyca: No need to specify a default color since in this case `hasColor()` will return false as expected.

danielang: so you will only check `hasColor()` in the `setStyle` function and remove the `has...`functions?

danielang:
any reaction? 😄

android/app/src/main/java/com/reactnativenavigation/views/BottomTabs.java
guyca: Yes exactly :)

guyca:
@danielang Hey sorry about the delay. If I don't respond you can always try pinging me on discord. 👍

witalobenicio:
Is this working on latest release?

Issue #1309, opened 2017-05-31
IconSize of FAB Button 


kkroth:
The icon size of the FAB Button is not compliant with the material design guide 

[FAB](https://material.io/guidelines/components/buttons-floating-action-button.html#buttons-floating-action-button-floating-action-button)

I´m using vector icons for the FAB Button but i cant adjust the icon size the layout of the FAB scales the icon to this:

![image](https://cloud.githubusercontent.com/assets/18683097/26635905/6f281f22-461b-11e7-963f-896e59fa9b42.png)

The dimension of the FAB should be:
 Interior icon: 24 x 24dp
Floating action button circle: 56 x 56dp (this works)

---
### Environment
* React Native Navigation version: 1.1
* React Native version: 44.0
* Platform(s) (iOS, Android, or both?): Adnroid
* Device info (Simulator/Device? OS version? Debug/Release?): Simulator and Device , Debug and Release

developer239:
+1

Fuhrmann:
There it is a pull request to solve this issue.

PR #2227, opened 2017-11-26
[Android] Ajust FAB icon size

Fuhrmann:
As of today, the floating action button, when provided an icon always set the size bigger then the specified  in the material design guideline:

**Wrong**
image

Even if we provide a icon 24x24 the fab icon will always have the same size (auto resize is happening). This changes set the scale type of the fab to CENTER. This way, when you provide the icon with the correct size (24) you should get this:

**Correct**
image

------

image

------

image

**An important thing to note**

Introducing this scale type will prevent the FAB to auto resize the icon. Let's say the user provide an icon 50x50. This should be the result (`which I think it's not a problem`):

image

I tested this in API 19, 22 and 25.

This fix the problem pointed out in the #1309.


jakewtaylor:
Thanks, this works great.

Any reason for not merging @guyca?

yershalom:

mths0x5f:
If I remember correctly, v1 PRs will be not processed anymore.

jakewtaylor:
This really doesn't need to be marked as stale. It's a simple PR with no conflicts that I can see.

Is v1 completely unsupported now?

PR #2239, opened 2017-11-28
Add unregisterForPreviewingWithContext

birkir:
- [x] Keep the preview controller context.
 - [x] Unregister for previewing on next screen push that is a preview.

3D Touch diagram

Sometimes when previewing different screens, they would "cache" and preview the correct one, but commit the old one.

This seems to fix that problem, also it is just a good idea to keep the context. It could be used later to allow users to unregister via an API.


rgoldiez:
@birkir - I applied this patch manually and I'm still seeing the same issue on previews.  Seems to happen when trying to preview the same component but with different props (and from the same screen).  In my case, I have a conversation feed and the preview functionality is to peak at a user's profile (or a tap / force'd press to go to their profile).  So on the same base `screen` you can preview the same component (userProfile) with different props for different users.

enahum:
@rgoldiez I have a screen that receives props and is being shown with peek, that same screen shows with different props for a list of different items in your case `userProfile` in my case `Channels` and it works perfectly

this is the code for our app if you want to take a look https://github.com/mattermost/mattermost-mobile/pull/1203

but without this PR it does show the last cached screen

enahum:
@birkir I think this branch needs another rebase

birkir:
Done. Also after looking at your PR for mattermost, I see that you are not committing the controller because it is just updating the UI rather than pushing a new screen.

What we could do is dispatch an event to the topViewController saying that the preview was deep pressed, so you could add an event listener and call `onSelectChannel`.

Just a thought to extend this even further.

enahum:
Yeah that might be pretty neat

guyca:
@birkir What's the status of this PR? is it ready or requires some more revisions?

birkir:
This is ready to be merged!

enahum:
is there anything else needed for this PR to get merged?

birkir:
Just waiting for a moderator to review and merge.

enahum:
friendly ping to @guyca ;)

PR #2242, opened 2017-11-29
Fix formatting in iOS installation documentation

floriangosse:


yershalom:

PR #2243, opened 2017-11-29
V2 view

Cool04ek:


lib/android/app/src/main/java/com/reactnativenavigation/react/coordinator/CoordinatorLayoutView.java
```diff
@@ -0,0 +1,54 @@
...
+import android.widget.FrameLayout;
+
+public class CoordinatorLayoutView extends CoordinatorLayout {
+
+	public CoordinatorLayoutView(Context context) {
+		super(context);
+		init();
+	}
+
+	public CoordinatorLayoutView(Context context, AttributeSet attrs) {
```
DanielZlotin: remove additional unused constructors, and inline the `init` method to be inside constructor

lib/android/app/src/main/java/com/reactnativenavigation/react/coordinator/CoordinatorLayoutView.java
```diff
@@ -0,0 +1,54 @@
...
+					MeasureSpec.makeMeasureSpec(getWidth(), MeasureSpec.EXACTLY),
+					MeasureSpec.makeMeasureSpec(getHeight(), MeasureSpec.EXACTLY));
+			layout(getLeft(), getTop(), getRight(), getBottom());
+		}
+	};
+
+	@Override
+	public void requestLayout() {
+		super.requestLayout();
+		post(measureAndLayout);
```
DanielZlotin: why is this needed? you already set MATCH_PARENT on yourself

lib/android/app/src/main/java/com/reactnativenavigation/react/coordinator/FloatingActionButtonManager.java
```diff
@@ -0,0 +1,105 @@
...
+			params.gravity = params.gravity | Gravity.RIGHT;
+			fab.setLayoutParams(params);
+		}
+	}
+
+	@ReactProp(name = "gravityLeft")
+	public void setGravityLeft(final FloatingActionButtonView fab, boolean gravityLeft) {
+		if (gravityLeft) {
+			CoordinatorLayout.LayoutParams params = (CoordinatorLayout.LayoutParams) fab.getLayoutParams();
+			if ((params.gravity & Gravity.RIGHT) == Gravity.RIGHT) {
```
DanielZlotin: I don't know what these are, but I'm guessing they could be refactored out to a helper function?

DanielZlotin: additionally, all this logic should probably be inside the view

lib/android/app/src/main/res/values/dimen.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <dimen name="margin">16dp</dimen>
```
DanielZlotin: why do we need it in xml? move it to a constant where it's used

package.json
```diff
@@ -45,7 +45,8 @@
     "react-native": "*"
   },
   "dependencies": {
-    "lodash": "4.x.x"
+    "lodash": "4.x.x",
+    "prop-types": "^15.6.0"
```
DanielZlotin: this should not be a dependency, it should go into devDependencies and peerDependencies: and with "version *"

playground/src/containers/LifecycleScreen.js
```diff
@@ -3,7 +3,7 @@ const { Component } = require('react');
 
 const { View, Text, Button } = require('react-native');
 
-const Navigation = require('react-native-navigation');
+const { Navigation } = require('react-native-navigation');
```
DanielZlotin: why the destructuring? this should stay a simple export

lib/src/index.js
```diff
@@ -1,4 +1,12 @@
...
+export {
+  Navigation,
+  FloatingActionButton,
+  CoordinatorLayout
+};
+
+export default Navigation;
 
-module.exports = Navigation;
-module.exports.default = Navigation;
```
DanielZlotin: undo this change. Navigation should be the only entry point. these 2 classes should be static properties of the internal Navigation class

lib/android/app/src/main/java/com/reactnativenavigation/react/coordinator/CoordinatorLayoutView.java
Cool04ek: It's some kind of bug from RN. It doesn't measure children properly without this 'hack'

lib/android/app/src/main/res/values/dimen.xml
Cool04ek: if it's constant, you need to handle converting dp to px. It's just easier with .xml

playground/src/containers/LifecycleScreen.js
Cool04ek: it's because of exporting not only Navigation. It can be changed after improving issue with export

lib/android/app/src/main/java/com/reactnativenavigation/react/coordinator/CoordinatorLayoutView.java
DanielZlotin: hmm. it's worth a comment then, explaining that. maybe RN50 solves it?..

lib/android/app/src/main/res/values/dimen.xml
DanielZlotin: I don't think it's a good idea to put such a general constant in the resources. which margin? where? also, how are the res being merged with RN apps? did you check?

Issue #1632, opened 2017-08-03
Drawer action not invoked when `dismissAllModels` used.


kusalshrestha:
### Issue Description

I'm currently using side menu Drawer in iOS platform. And i got a case where more than 2 modals are shown in succession, one upon the other; So, I've used `dismissAllModels` to dismiss all the presented modals at once. Problem is when i use `dismissAllModels`, side menu drawer stops working.

---
### Environment
* React Native Navigation version: 1.1.x
* React Native version: 0.44.0
* Platform(s) (iOS, Android, or both?): iOS
* Device info (Simulator/Device? OS version? Debug/Release?): Both in device & in simulator. And in both configuration

gran33:
What do u mean by "stops working"? Gestures aren't open the drawer? If u explicit open the drawer by action on button, is it working?

kusalshrestha:
Means side menu bar doesn't open @gran33

elliotstokes:
I think im seeing the same thing. The hamburger drawer menu stops working once you call dismissallmodals. The gesture still opens it but then any code that specifically opens/closes the drawer not through a gesture fails (appears to be ios only for me)

elliotstokes:
Dont suppose you came up with a workaround or a fix for this at all?

Issue #1550, opened 2017-07-21
dismissAllModals breaking toggleDrawer on ios


CoderProject:
### Issue Description

After I call `dismissAllModals()`, `toggleDrawer()` simply does not do anything on iOS.
Android works as expected. Is there a quickfix that may save my day?

### Steps to Reproduce / Code Snippets / Screenshots

- Open multiple modals recursively using `showModal()` (modal in modal, 2 is enough)
- Call `this.props.navigator.dismissAllModals();` from the latest modal. I also tried calling it from root screen via deep linking.
- All modals will disappear.
- Now try the following and you will see that the drawer will not toggle.
```
this.props.navigator.toggleDrawer({
	  side: 'left', // the side of the drawer since you can have two, 'left' / 'right'
	  animated: true, // does the toggle have transition animation or does it happen immediately (optional)
});
```

---
### Environment
* React Native Navigation version: 1.1.130
* React Native version: 0.45.1
* Platform(s) (iOS, Android, or both?): iOS
* Device info (Simulator/Device? OS version? Debug/Release?): Both, 10.X, both

ahm-hossein:
same problem here

agumack:
I have same problem

lxooxl:
use dismissModal  can solve this

CoderProject:
not even close.. I need all open modals to be closed, not just latest one.

kyle-ssg:
Yep, can replicate this also. Same thing happens if you use ```Navigation.dismissModal``` vs ```props.navigator.dismissModal```

lovbrand:
+1

henrikra:
I have replicated this too! After `dismissAllModals` is called the `this.props.navigator.toggleDrawer({ side: 'left' });` does not work anymore! Can you guys check this out @alinz @yomybaby @gran33 @guyca

elliotstokes:
We are also seeing something similar. The drawer will stop working after dismissAllModals is called. Seems to be the same issue as https://github.com/wix/react-native-navigation/issues/1632

MrVisitor:
Maybe someone could understand me and help. After using the Navigation.dismissAllModals() method, DrawerControllerIOS is unavailable in the componentsDic.

PR #2245, opened 2017-11-29
Resolved issues where Drawer did not work after dismissAllModals

mckn:
I changed so we only clean up viewcontrollers if it is a modal. This should prevent the Drawer from stop working after an dismissAllModals.

Should resolve:
#1632, #1550

Might resolve:
#1598, #418


gran33:
Fixed on 
https://github.com/wix/react-native-navigation/pull/2255

PR #2247, opened 2017-11-29
[iOS] Allow set view controllers to be called with multiple components

eliperkins:
The `UINavigationController` API allows for `setViewControllers:animated:` to be
called with multiple view controllers, for resetting entire stacks, rather than
just resetting to one view controller.

This allows for `navigator.resetTo()` to be called with an array of components,
rather than just one component, to let the navigation stack reset to multiple
view controllers.

This is backwards-compatible with the existing API.

An example of this would look like:

```js
this.props.navigator.resetTo({
  components: [
    {
      component: "ScreenA",
      passProps: {
        greeting: "Hello, World"
      },
      leftButtons: [
        // same as normal button setting for a screen
      ]
    },
    {
      component: "ScreenB",
      passProps: {
        greetingId: 1234
      },
      style: {
        // same as normal screen style
      }
    }
  ]
});
```

### TODO

* [ ] Add documentation

### DIFF
--- a/ios/RCCNavigationController.m
+++ b/ios/RCCNavigationController.m
@@ -257,31 +257,20 @@ NSString const *CALLBACK_ASSOCIATED_ID = @"RCCNavigationController.CALLBACK_ASSO
   {
     NSString *component = actionParams[@"component"];
     if (!component) return;
-    
+
     NSMutableDictionary *passProps = [actionParams[@"passProps"] mutableCopy];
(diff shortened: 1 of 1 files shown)


yogevbd:
@eliperkins  can you please fix conflicts?

guyca:
Also worth noting this api was added to v2 👍

eliperkins:
Rebased, so conflicts should be fixed!

PR #2255, opened 2017-11-30
Fix dismissAllModal API

gran33:
The root won't unregister when `dismissAllModals` been called


jasonmerino:
@gran33 thanks for this fix. Has this been released to version 1?

PR #2256, opened 2017-11-30
Android buttons

Johan-dutoit:
This is a good start for the Android buttons.  Has all the basic functionality.
Once this goes in, more people can build on top of this.


lib/android/app/src/main/java/com/reactnativenavigation/parse/NavigationOptions.java
```diff
@@ -39,6 +45,13 @@ public static NavigationOptions parse(JSONObject json) {
 		result.topBarHidden = BooleanOptions.parse(json.optString("topBarHidden"));
 		result.animateTopBarHide = BooleanOptions.parse(json.optString("animateTopBarHide"));
 
+		try {
+//			result.leftButtons = json.getJSONArray("leftButtons");
+			result.rightButtons =  Button.parseJsonArray(json.optJSONArray("rightButtons"));
+		} catch (JSONException e) {
```
Cool04ek: I'm not sure throwing an Exception is the best way. It better just to use `opt()` and check for null.

lib/android/app/src/main/java/com/reactnativenavigation/utils/ImageUtils.java
```diff
@@ -0,0 +1,39 @@
...
+import android.support.annotation.NonNull;
+
+import java.io.IOException;
+import java.net.URL;
+
+public class ImageUtils {
+    @NonNull
+    public static Drawable tryLoadIcon(Context currentContext, String uri) {
+
+        StrictMode.ThreadPolicy threadPolicy = StrictMode.getThreadPolicy();
```
Cool04ek: why do you need it?

lib/android/app/src/main/java/com/reactnativenavigation/views/TopBar.java
```diff
@@ -1,24 +1,32 @@
...
 	private final Toolbar titleBar;
+	private ContainerViewController.ContainerView containerView;
 
-	public TopBar(final Context context) {
+	public TopBar(final Context context, ContainerViewController.ContainerView containerView) {
 		super(context);
-		titleBar = new Toolbar(context);
+
+		this.titleBar = new Toolbar(context);
+		this.containerView = containerView;
```
Cool04ek: why do you need it here?

lib/android/app/src/main/java/com/reactnativenavigation/utils/ImageUtils.java
Johan-dutoit: I got this from V1.  Basically, it allows us to "download" the image synchronously instead of async.

lib/android/app/src/main/java/com/reactnativenavigation/parse/NavigationOptions.java
Johan-dutoit: Completely agree, was mostly like that whilst I was developing.

lib/android/app/src/main/java/com/reactnativenavigation/views/TopBar.java
Johan-dutoit: containerView has the emitter that is needed to send the event to React-Native

lib/android/app/src/main/java/com/reactnativenavigation/utils/ImageUtils.java
Cool04ek: I don't see any reasons doing it on `main` thread. It's rather better to do it async.

lib/android/app/src/main/java/com/reactnativenavigation/views/TopBar.java
Cool04ek: yeah, i got it. It's ok :)

AndroidE2E/app/src/androidTest/java/com/reactnativenavigation/e2e/androide2e/NavigationOptionsTest.java
```diff
@@ -32,4 +32,11 @@ public void testTopBarHidden() throws Exception {
 		elementByText("SHOW TOP BAR").click();
 		assertExists(By.text("Static Title"));
 	}
+	
+	@Test 
+	public void testRightButtons() throws Exception {
+		elementByText("PUSH OPTIONS SCREEN").click();
+		assertExists(By.text("ONE"));
+		elementByText("ONE").click();
```
Cool04ek: Test should always end with assert, not click

lib/android/app/src/main/java/com/reactnativenavigation/parse/NavigationOptions.java
```diff
@@ -32,15 +40,28 @@ public static NavigationOptions parse(JSONObject json, @NonNull NavigationOption
...
 	public BottomTabsOptions bottomTabsOptions = new BottomTabsOptions();
+	public ArrayList<Button> leftButtons;
+	public ArrayList<Button> rightButtons;
 
 	public void mergeWith(final NavigationOptions other) {
         topBarOptions.mergeWith(other.topBarOptions);
 		bottomTabsOptions.mergeWith(other.bottomTabsOptions);
+
+		if(other.leftButtons != null) {
+			leftButtons = other.leftButtons;
```
Cool04ek: you need to merge button options. Because now it's replaced. So, if you want just to change one property, you need to provide all another once more.

lib/android/app/src/main/java/com/reactnativenavigation/views/TitleBarButton.java
```diff
@@ -0,0 +1,163 @@
...
+
+		if (hasIcon()) {
+			setIcon(context, menuItem);
+		} else {
+			setTextColor();
+			setFontSize(menuItem);
+		}
+	}
+
+	public void setNavigationIcon(Context context) {
```
Cool04ek: please, rename to "applyNavigationIcon". Now it's kind of misleading, 'cause it looks like you need to provide icon, that should be set

lib/android/app/src/main/java/com/reactnativenavigation/views/TitleBarButton.java
```diff
@@ -0,0 +1,163 @@
...
+
+			@Override
+			public void onError(Throwable error) {
+				//TODO: handle
+				error.printStackTrace();
+			}
+		});
+	}
+
+	private void setIcon(Context context, final MenuItem menuItem) {
```
Cool04ek: please, rename to "applyIcon". Now it's kind of misleading, 'cause it looks like you need to provide icon, that should be set

lib/android/app/src/main/java/com/reactnativenavigation/utils/UiUtils.java
```diff
@@ -14,4 +19,12 @@ public boolean onPreDraw() {
 			}
 		});
 	}
+
+	public static void tintDrawable(Drawable drawable, int tint) {
+		drawable.setColorFilter(new PorterDuffColorFilter(tint, PorterDuff.Mode.SRC_IN));
+	}
+
+	public static void runOnMainThread(Runnable runnable) {
```
guyca: We can use `UiThread.post` 👍

Issue #1551, opened 2017-07-21
Add a changelog or a whats changed.


elliotstokes:
### Issue Description

Thanks for setting up the project - its been a real help all round. One thing were finding though is that updates are pushed up pretty regularly (we often get 60+ builds behind) Looking in the project though I cant find any documentation (Changelog or similar) as to whats been fixed/modified/removed/added within those changes. It would be really useful to get some indication as to whats changed, even if its just the headlines.

I'm currently doing a compare https://github.com/wix/react-native-navigation/compare/v1.1.88...v1.1.142

Thanks again

skovhus:
Would be really nice with a change log. Are you waiting for 2.x to be out before doing that?

guyca:
closing in favor of #2162

PR #2261, opened 2017-12-02
Fix rn 0.50 repeated calls to onMeasure.

chaitanya0bhagvan:
RN 0.50 introduced a new feature to dynamically measure ReactRootView. Since RNN uses RelativeLayout for Screen and BottomTabsLayout child views added to the layout may have additional rules. When the child views are measured they are measured with respect to the LayoutParams set during Screen creation and the Parent MeasureSpec. In RNN, ContentView extends ReactRootView and it calls onMeasure using the widthMeasureSpec and heightMeasureSpec passed to it. If these measureSpecs are EXACTLY then ReactRootView will measure itself to fit the exact size requested by the parent. 

If ContentView was added with LayoutParams.MATCH_PARENT for height and/or width with additional rules eg. BELOW or ABOVE the effective height of this View may not be compatible with MATCH_PARENT. RelativeLayout does two passes over its children, in the first pass each child is measured as per their LayoutParams, in the second pass LayoutParam rules are applied and each child is measured. The problem occurs here, if ReactRootView was asked to measure itself with MeasureSpec.EXACTLY but the measuredHeight was less than expected it will trigger a requestLayout() causing the entire layout to be remeasured. 

The fix is to call super.onMeasure() in ContentView with MeasureSpec.AT_MOST.

### DIFF
--- a/android/app/src/main/java/com/reactnativenavigation/views/ContentView.java
+++ b/android/app/src/main/java/com/reactnativenavigation/views/ContentView.java
@@ -66,9 +66,13 @@ public class ContentView extends ReactRootView {
 
     @Override
     protected void onMeasure(int widthMeasureSpec, int heightMeasureSpec) {
-        super.onMeasure(widthMeasureSpec, heightMeasureSpec);
-        int measuredHeight = viewMeasurer.getMeasuredHeight(heightMeasureSpec);
-        setMeasuredDimension(viewMeasurer.getMeasuredWidth(widthMeasureSpec), measuredHeight);
+        int widthSpec = MeasureSpec.makeMeasureSpec(MeasureSpec.getSize(widthMeasureSpec), MeasureSpec.AT_MOST);
+        int heightSpec = MeasureSpec.makeMeasureSpec(MeasureSpec.getSize(heightMeasureSpec), MeasureSpec.AT_MOST);
+
+        super.onMeasure(widthSpec, heightSpec);
+
+        int measuredHeight = viewMeasurer.getMeasuredHeight(heightSpec);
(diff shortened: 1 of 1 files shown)


enahum:
is there anything else needed for this PR to get merged?

bitcrumb:
Anything that is holding back this PR? If not, would like to see this merged as soon as possible :)
Thanks!

CherishIt:
Also waiting for this PR being merged to carry on with my work, thanks.

fforres:
Friendly Ping @guyca and/or @DanielZlotin  ? :)

ChristianAyala:
I can also confirm that this fixes the flickering issue in our app. Would love to see this merged in soon!

koansang:
When do you merge this PR? @guyca

ReggaePanda:
it works! thanks 🥇

Flavien:
Is that on NPM yet?

ReggaePanda:
@Flavien yes, I use yarn and with `yarn upgrade-interactive` I updated my project and `yarn start --reset-cache` it fixed this issue.

jasonmerino:
Looks like this fix was released with version `1.1.307`. 🙌🏼

ariona:
This fix the flickering view issue on my app :clinking_glasses:

skovhus:
@chaitanya0bhagvan is this backwards compatible? So we can use this with react-native < 0.50?

I really miss a change log here or at least which versions that depends on which react-native versions.

Related #1551

chaitanya0bhagvan:
@skovhus This is backward compatible change. As of now the docs and examples have not been updated to reflect RN 0.50 change where `index.android.js` and `index.ios.js` have been replaced with `index.js`. Let me know if you need any help with the RN upgrade.

PR #2264, opened 2017-12-03
V2 Current tab by container ID

yogevbd:


guyca:
@yogevbd What's the status of this PR?

playground/src/containers/TextScreen.js
```diff
@@ -6,6 +6,8 @@ const { View, Text, Button } = require('react-native');
 const Navigation = require('react-native-navigation');
 const testIDs = require('../testIDs');
 
+let globalFirstContainerID;
```
DanielZlotin: this should be removed once we have broadcasting mechanism https://github.com/wix/react-native-navigation/issues/2417

PR #2273, opened 2017-12-04
disabled back gesture support

yogevbd:


yogevbd:
@DanielZlotin changes done

DanielZlotin:
@yogevbd this means that default value will be `true` right?

yogevbd:
@DanielZlotin yes

DVLNarayana:
how and where to set this property from jscode?

yogevbd:
@DVLNarayana 
```
options: {
    popGesture: false // true by default
  }
```

Issue #1942, opened 2017-09-28
Side menu / drawer randomly doesn't work. Caused by 0.0 width on setSideMenuWidth


mobitar:
It's very difficult to replicate consistently, but sometimes, the side drawer just won't work. I thought I would ignore this issue, thinking it was just me, but all my users have reported it.

After fiddling around, the cause turns out to be in `SideMenu.java` in the `setSideMenuWidth` function:

```
lp.width = sideMenuView.getChildAt(0).getWidth();
```

In instances where this issue happens, this value is 0.0. 

The fix, as noted in [this issue](https://github.com/wix/react-native-navigation/issues/1583), is to explicitly set a width on the JS component:

```
import { Dimensions } from 'react-native';

...

<View style={{width: Dimensions.get('window') * 0.8}}>
```

One way that I've been able to force replicate this is by causing my app to crash by pressing back rapidly on a pushed screen many times, such that it pops too far back into a stack and crashes. Upon re-open, the drawer doesn't open.

#### Environment
React Native Navigation version: 1.1.231
React Native version: 0.45.1
Platform(s) (iOS, Android, or both?): Android
Device info (Simulator/Device? OS version? Debug/Release?): Simulator / Any device / Tested on Android 7.0

em222iv:
We have got it working consistently again with this combination:
"react-native": "0.48.4",
"react-native-navigation": "1.1.235",

sqow:
I'm seeing this issue reliably when going from:

```js
Navigation.startSingleScreenApp({
  screen: {
    screen: 'NoSideNavScreen',
    navigatorStyle: {
      navBarHidden: false
    },
    navigatorButtons: {}
  }
})
```

to:

```js
Navigation.startSingleScreenApp({
  screen: {
    screen: 'WithSideNavScreen',
    navigatorStyle: {
      navBarHidden: false
    },
    navigatorButtons: {
      leftButtons: [
        { id: 'sideMenu' }
      ]
    }
  },
  drawer: {
    left: {
      screen: 'TheDrawerComponent'
    }
  }
})
```

What might also be helpful is that **reversing** those (i.e. going from `startSingleScreenApp` where a drawer is configured to `startSingleScreenApp` without a drawer at all) and then **back** to the drawer configuration seems to work fairly reliably.

sqow:
Starting via `startSingleScreenApp` with a drawer configured in the params works as expected.

eeynard:
[...]

Issue #1863, opened 2017-09-14
Drawer is not working on Android


bdavid68:
### Issue Description

Drawer is not shown on Android

### Steps to Reproduce / Code Snippets / Screenshots

```
this.props.navigator.toggleDrawer({
      side: 'left',
      animated: true
    });
```

---
### Environment
* React Native Navigation version: FILL THIS OUT
* React Native version: FILL THIS OUT
* Platform(s) (iOS, Android, or both?): FILL THIS OUT
* Device info (Simulator/Device? OS version? Debug/Release?): FILL THIS OUT

nimerfarahty:
set width for the drawer screen , it will works

mathieuroblin:
I tried the drawer and am also encountering some issues on android even after setting up the width of the view.

I am able to show the drawer if I swipe my finger from the left of the screen to the center but not if I use navigator.toggleDrawer.

The code is running fine on iOS though.

React Native Navigation : 1.1.233
React Native version : 0.48
Platform : Android
Device info : Simulator, Android 7.1 & 8 (Haven't tried more versions) Debug

BIWhitfield:
Hey guys for the time being I've followed the suggestion of setting the width directly on my Sidemenu component and it seems to be working.

` render() {
    return (
      <View style={{ height: '100%', width: 280, backgroundColor: 'rgb(240,239,245)' }}>
    .......
}
`

SquaredData:
I encounter the same issue, even with the width-advise...

bdavid68:
Yeah, set width seems to be working.

SquaredData:
Hey bdavid, why did you close? I still have this problem...

fmussap:
https://github.com/SquaredData

I was having this problem when using RN 0.51.
I downgraded it to RN 0.49.5 and now it's working when you set width property on the Sidemenu component

SquaredData:
Thanks a lot!!! RN 0.49.5 works for me as well :)

jamesguan:
This is still an issue, even on react-native version 0.54.0

denieler:
I have this problem in RN 0.55.3 with "react-native-navigation": "^1.1.457".

I've set width for the Drawer component itself. But randomly after application installed and opened drawer doesn't open. I use `toggleDrawer` function to open it.

```
navigator.toggleDrawer({
  side: 'left',
  animated: true
})
```

After I close the app and open again it can start working. It's in release mode.
[...]

Issue #1851, opened 2017-09-13
Drawer is not working with nested "startSingleScreenApp" call.


kaltindas:
I have a Login/Register Page and on press on Login Button i want to start a new SingleScreenApp with Drawer. Drawer (Hamburger Btn) will rendered but without function. It happens nothing on click.. Same code (startSingleScreenApp) does work if i execute it on the root of App. 

`
export default class Splash extends Component {

    onPressed(){
        Navigation.startSingleScreenApp({
            screen: {
                screen: 'Home',
                title: 'Welcome',
                navigatorButtons: {
                    leftButtons: [
                        {
                            id: 'sideMenu',
                        }
                    ]
                }
            },
            drawer: {
                left: {
                    screen: 'Drawer',
                },
            },
            animationType: 'slide-down'
        });
    }

    render() {
        return(
            <View>
                <Button
                    onPress={ () => this.onPressed() }
                    title="Login And Open App"
                    color="#841584"
                />
            </View>
        )
    }
}` 

Whats wrong here?

chakrihacker:
Any update, I am having same problem

RiCoTeRoX:
@kaltindas hi! did you resolve this issue ?

isaachinman:
I am also experiencing this issue on an iOS 11 real device. Interestingly, it works fine in simulator.

Using react-native-navigation v1.1.407

@guyca Any idea what this is about?

nemat001:
Rather than opening a drawer with navigation.navigate(‘DrawerOpen’), you can now call navigation.openDrawer()

Issue #1583, opened 2017-07-27
toggleDrawer not working at production build when loads first time


Maydar:
### Issue Description
 toggleDrawer not working when application loads **first time after installation.**
If I reopen it, then it works. In Debug it works at majority times.

UPDATE
**downgrade to RN  0.45.1 resolved this issue** but it's still not applicable for me because I'm getting a few new bugs related to RN 0.45.1 =(

### Steps to Reproduce / Code Snippets / Screenshots
- build APK
- disable Drawer by setDrawerEnabled = false on first screen and enable it before calling resetTo to next screen. After that try to touch sideMenu icon and then drawer completely stop work until I  reopen app again.
- install it and **open it from installation menu**

### Environment
* React Native Navigation version: 1.162
* React Native version: 0.46.4
* Platform(s) (iOS, Android, or both?): Android  
* Device info (Simulator/Device? OS version? Debug/Release?): Device, Android 5-7, Release

matthewjamieson:
Im getting the same issue

Maydar:
@OssamaZ I already did this things. As I mentioned, this situation appers only when I open it first time.

aimbot:
Having the same issue on 46.4.

guyca:
@Maydar Can you try building the example app in release mode and see if it reproduces there?

Andrfas:
@Maydar could you please tell me did you find any solution?

phiphamuet:
Same issues with you. I finding some solution to solve it.

Strayer:
I have the same issue. It frequently occurs in release builds, but also often in debug builds. Interesting sidenote: When the issue occurs, setting the statusBarColor also doesn't seem to work in our project! Wasn't able to reproduce that though…

@guyca I uploaded an example app at https://github.com/Strayer/ReactNativeNavigationAndroidDrawerIssues that reliably reproduces the issue on my Moto Z (7.1.1)

Sometimes the drawer works, sometimes not. I recommend restarting the application a few times if it doesn't show the problem right away.

Strayer:
Okay, so… I *think* I'm on to something here. Since debugging release builds is more or less impossible I'm just fiddling around in the code of SideMenu, but when I add a specific width in setSideMenuWidth the drawer _does_ appear, albeit way too thin:

```java
private void setSideMenuWidth(final ContentView sideMenuView) {
        sideMenuView.setOnDisplayListener(new Screen.OnDisplayListener() {
            @Override
[...]

PR #2275, opened 2017-12-04
[Android] Added possibility to set drawer width based on screen size

mckn:
I have added the possibility to configure the the width of the drawer based on a percentage value of the screen size. 

This should resolve the following issues:
#1583, #1942, #1863, #1851


mckn:
Ah, sorry..I did not see that. I will revert that change. Thanks!

mckn:
@ItsNoHax Now it should be updated without the re-formatting.

mckn:
Closed due to the fixedWidth property that already is added.

PR #2297, opened 2017-12-06
V2 tabBarHidden

yogevbd:


DanielZlotin:
@yogevbd please rename all instances of `tabBar` to `bottomTabs`

yogevbd:
@DanielZlotin  It means that I have to change it on android as well right? or do you want the api to stay the same with "topBar"?

PR #2321, opened 2017-12-11
Adds launch image for iPhone X and iPad to showSplashScreen.

jasonmerino:
This resolves the black screen showing in between when a static launch image shows and the JS code renders your app on iOS.


shivawu:
Bump. This seems to be the correct fix

PR #2324, opened 2017-12-12
V2 side menu visibility ios (toggleDrawer)

yogevbd:


lib/ios/RNNSideMenuController.h
```diff
@@ -19,4 +19,6 @@
 
 -(instancetype)initWithControllers:(NSArray*)controllers;
 
+-(void)showSideMenu:(BOOL)show side:(MMDrawerSide)side animated:(BOOL)animated;
```
DanielZlotin: split this into 2 method: show, hide

Issue #2146, opened 2017-11-09
Android - Can't set TopTabs font family


fmorato:
### Issue Description

Can't set TopTabs font family.

### Steps to Reproduce / Code Snippets / Screenshots

There is no implementation for the feature.

How can I set the TopTabs font family? I tried to follow the code structure and copy other font definitions and created the method bellow in TopTabs.java. It is then called in TopBar.java's `setTopTabsStyle`

 ```
   void setTopTabsFontFamily(StyleParams style) {
        if (style.topTabFontFamily != null && style.topTabFontFamily.hasFont()) {
            ViewGroup vg = (ViewGroup) this.getChildAt(0);
            int tabsCount = this.getTabCount();
            for (int j = 0; j < tabsCount; j++) {
                ViewGroup vgTab = (ViewGroup) vg.getChildAt(j);
                int tabChildsCount = vgTab.getChildCount();
                for (int i = 0; i < tabChildsCount; i++) {
                    View tabViewChild = vgTab.getChildAt(i);
                    if (tabViewChild instanceof TextView) {
                        ((TextView) tabViewChild).setTypeface(style.topTabFontFamily.get());
                    }
                }
            }
        }
    }
```
The whole code executes, including `setTypeface` but nothing happens. What am I missing?

---
### Environment
* React Native Navigation version: 1.1.262
* React Native version: 0.50.3
* Platform(s) (iOS, Android, or both?): Android
* Device info (Simulator/Device? OS version? Debug/Release?): Device v 6.0.1

hoorsa:
+1

Issue #1784, opened 2017-09-02
change top tabs font family 


ahmedfouad2016:
change top tabs font family 
Is there any way to change top tabs font family in the topTabs or appStyle ? i used tabFontFamily but it change the font family in bottom bar
---
### Environment
* React Native Navigation version: 1.1.143
* React Native version: 0.46.4
* Platform(s) (iOS, Android, or both?): Android
* Device info (Simulator/Device? OS version? Debug/Release?): Device

ahmedfouad2016:
is there any update about this issue  ? @guyca

salisuabubakar:
Any update?

hoorsa:
+1

PR #2342, opened 2017-12-13
[Android] TopTabs font family

krystofcelba:
Added topTabTextFontFamily option so the font family of TopTabs can be set from JS now.

Resolves: #2146, #1784


hoorsa:
Thanks @krystofcelba

PR #2345, opened 2017-12-14
Tab bar items

yogevbd:


playground/src/containers/TextScreen.js
```diff
@@ -42,7 +42,7 @@ class TextScreen extends Component {
 
   onButtonPress() {
     Navigation.setOptions(this.props.containerId, {
-      bottomTabs: {
+      bottomTab: {
         tabBadge: `TeSt`
```
guyca: `badge` instead of `tabBadge` ?

PR #2410, opened 2017-12-25
Top tab font family

guyca:
API

```js
Navigation.push(this.props.containerId, {
  topTabs: [
          {
          name: 'navigation.playground.TopTabScreen',
          passProps: {
            title: 'Tab 2',
            text: 'This is top tab 2'
          },
          navigationOptions: {
            topTab: {
              title: 'Tab 2',
              titleFontFamily: 'Dosis-Regular'
            }
          }
        }
  ]
});
```


lib/android/app/src/main/java/com/reactnativenavigation/parse/TopTabOptions.java
```diff
@@ -1,15 +1,23 @@
...
+    @Nullable public Typeface fontFamily;
+    public int tabIndex;
 
-    public static TopTabOptions parse(JSONObject json) {
+    public static TopTabOptions parse(TypefaceLoader typefaceManager, JSONObject json) {
         TopTabOptions result = new TopTabOptions();
         if (json == null) return result;
 
         result.title = json.optString("title", NO_VALUE);
+        result.fontFamily = typefaceManager.getTypeFace(json.optString("titleFontFamily"));
```
Cool04ek: Why do you need to load typeface now. I mean it's better to have it consistent. Because topBar options load typeface in presenter, and topTabs in parser.

PR #2473, opened 2018-01-04
Add iPad landscape launch image

adamski:
Add the iPad landscape launch image. Without this you get an ugly stretched portrait image when loading in landscape orientation.


yershalom:

PR #2480, opened 2018-01-05
[WIP] [v2] [iOS] Refactor ReactNativeNavigation class into a developer-controllable class

eliperkins:
Some React Native libraries require access to the bridge, or perhaps a user wants to create a separate `RCTRootView` without `react-native-navigation`. This would require a second bridge (which is less than desirable for the library use-case), or access to the bridge that RNN creates.

By creating a class to manage application startup and hold on to the properties that we need, we can allow developers more control over how the app is launched, in a backwards-compatible way.

Additionally, this creates a more tree-like pattern for our model of ownership of objects, allowing for an easier control-flow and clearer source of truth.


DanielZlotin:
I think it's best to refactor RNNBridgeManager and inline it to be the ReactNativeNavigation object. Simpler and more direct. This change will remove RNN from being a singleton (which in itself is a good thing).
This way, you either create `[ReactNativeNavigation alloc init]`yourself or use the factory method of `bootstrap`

DanielZlotin:
A similar change should be introduced to NavigationApplication and NavigationActivity in android

PR #2510, opened 2018-01-11
V2 overlays

yogevbd:


yogevbd:
@DanielZlotin I added interactable just to give an example in our example project on how to implement overlays with animations (like in-app notifications)

DanielZlotin:
due to it being unmaintained and in an alpha stage I don't think it's a good idea, unless there is a specific use case which requires us to support it explicitly

playground/src/screens/OptionsScreen.js
```diff
@@ -190,6 +194,14 @@ class OptionsScreen extends Component {
     });
   }
 
+  onClickShowOverlay(passThroughTouches) {
+    Navigation.showOverlay({
+      component: {
+        name: 'navigation.playground.CustomDialog'
+      }
+    }, { passThroughTouches });
```
DanielZlotin: passThroughTouches should be part of `options` I think

yogevbd: done

DanielZlotin: and let's rename that to `interceptTouches`

DanielZlotin:
and let's rename that to interceptTouches

lib/src/commands/Commands.js
```diff
@@ -62,22 +62,20 @@ class Commands {
...
+  showOverlay(componentData) {
+    const input = _.cloneDeep(componentData);
+    OptionsProcessor.processOptions(input);
+
+    const layout = this.layoutTreeParser.parse(input);
+    this.layoutTreeCrawler.crawl(layout);
+
+    const promise = this.nativeCommandsSender.showOverlay(layout);
+
+    return promise;
```
anonrig: Why not `return this.nativeCommandsSender.showOverlay(layout)`?

PR #2529, opened 2018-01-15
Adding preferredContentSize

DavidCMurphy:
So while building an iPad application using this navigation library i have come across the need to set the size on controllers presented through modalPresentationStyle = formSheet.

Though currently we don't have any prop that allows you to do so. Obviously this is iOS specific but i think some other people would like some flexibility in this regard.

example:

```javascript
Navigation.showModal({
            screen: action$.screen,
            title: action$.title,
            animationType: 'slide-up',
            navigatorStyle: {
                modalPresentationStyle: 'formSheet',
                preferredContentSize: { width: 790, height: 560 }
            },
            passProps: action$.props
        })
```


guyca:
Thanks for the pr @DavidCMurphy 
Could you please update the [docs](https://github.com/wix/react-native-navigation/blob/master/docs/top-level-api.md) with the available modalPresentation styles as well?

DavidCMurphy:
Hey @guyca, this should work, let me know if i'm missing anything.

PR #2539, opened 2018-01-16
SetStyle fix

yogevbd:
@danielang Removing changes in RCCNavigationController fixed the issue we had. Does this make sense?

### DIFF
--- a/android/app/src/main/java/com/reactnativenavigation/layouts/BottomTabsLayout.java
+++ b/android/app/src/main/java/com/reactnativenavigation/layouts/BottomTabsLayout.java
@@ -16,6 +16,7 @@ import com.facebook.react.bridge.Callback;
 import com.facebook.react.bridge.Promise;
 import com.facebook.react.bridge.WritableMap;
 import com.reactnativenavigation.NavigationApplication;
+import com.reactnativenavigation.events.Event;
 import com.reactnativenavigation.events.EventBus;
 import com.reactnativenavigation.events.ScreenChangedEvent;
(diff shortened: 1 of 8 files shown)


danielang:
I've add this changes because if you don't update all `viewControllers` in the stack of the screen they won't update their styles when you pop them: the tabbar stays in their previous color, etc.

yogevbd:
@danielang The problem is that all viewControllers in UINavigationController stack will call `updateStyle` on `setStyle`. It causes styles that affect UINavigationBar to be applied multiple times (like `navBarHidden`).
I think moving tabBarItem styles to `setStyleOnInit` will solve this.

danielang:
@yogevbd: by the naming convention of the method `setStyleOnInit` I thought that this method would only called once while the viewController is initializing.
So how the viewController will be notified when the style of the tabbar has been changed?

danielang:
@yogevbd any reaction? 🙂

guyca:
The feature your PR implemented is updating BottomTabs (tabBar) style by calling `setStyle` - @danielang can you confirm this PR covers the use cases you had in mind?

danielang:
@guyca: Yes, of cause.
But it seems that @yogevbd  has discovered an 'bug' in iOS that the naviationbar will be updated too often while transitioning between screens; so it produces some lags.
I couldn't verfiy it in my simulator. And we've began a discussion about a better implementation. And now I am waiting for a response.

yogevbd:
@danielang What I had in mind is to implement another function that handled all tabItem styles and call it every time just like setStyleOnInit.

PR #2584, opened 2018-01-23
Check whether the sharedBridge is valid

gsteyn:
Checking whether the `sharedBridge` is valid instead of only checking whether it is not null. This allows us to initialise a new bridge if the bridge was invalidated. This is useful when creating an app that loads other React Native apps, and the user wants to close the loaded app manually to try and load another.


yershalom:

PR #2112, opened 2017-11-03
[iOS] Large title text attributes

eliperkins:
This adds support for changing `largeTitleTextAttributes` on navigation controllers, including custom font sizes, font families, etc.

See Apple docs here: https://developer.apple.com/documentation/uikit/uinavigationbar/2919946-largetitletextattributes 

Supports #2090 

Related to #1643

PR #2593, opened 2018-01-25
Add support for prefersLargeTitles in iOS 11

### LINKED ISSUE #1643: [Feature request] iOS 11 Style Navbar Titles

hammadj:
Fixes #1643


oferRounds:
Cool!

I think if we’re adding that, to make it complete, we should also add support to customize the appearance of the large title, such as fontFamily, fontSize (i.e., textAttributes).

eliperkins:
@oferRounds support for appearance of large titles is in #2112

PR #2595, opened 2018-01-25
fixed failing unit test

andresesfm:
Simple e2e Test fix


guyca:
Thanks @andresesfm 
We're still not running Android e2e on CI since we'd like to migrate to Detox first.. that's how this one slipped by.
Thanks for fixing and adding the missing test 👍

PR #2617, opened 2018-01-31
[Fix] Add iOS NavText font size and family support

RTsGIT:
Current RNN codebase does not adhere to the Docs for supporting navBarTextFontFamily attribute on iOS platform.

Custom font support for Android was added in https://github.com/wix/react-native-navigation/commit/86f397b782e23d91c8592270484cbb1106ad1f2d

This PR aims to support navBarTextFontFamily and navBarTitleFontSize on iOS. If navBarTextFontFamily or navBarTitleFontSize are not passed, default font attributes are chosen.


yershalom:

PR #2626, opened 2018-02-02
 Loading local resources into CollapsingTopBarBackground doesn't work in android in release mode

chubakueno:
Local resources are not being correctly loaded into collapsing top bar background  when building in release mode for android because of a missing translation step, keeping compatibility with react ^0.29 in this regard.


yershalom:

chubakueno:
Several fixes more have been commited, I would like to have this PR reviewed so that I can propose them in another PR or, if possible, accept them in this PR so that there is no additional delay.

PR #2661, opened 2018-02-06
[ios] Turn off Lightbox animation

nbomberger:
This provides a flag to allow the user to turn off lightbox animation.


yershalom:

PR #2680, opened 2018-02-08
V2 fab

Cool04ek:


lib/android/app/src/main/java/com/reactnativenavigation/views/StackLayout.java
```diff
@@ -34,6 +41,10 @@ public void onPress(String buttonId) {
 
     public void applyOptions(Options options, ReactComponent component) {
         new OptionsPresenter(topBar, component).applyOptions(options);
+        if (fabOptionsPresenter == null) {
```
guyca: Our code should be concise and deterministic. if `fabOptionsPresenter` is a property of the class, it should not be initialized conditionally on demand, instead create in ,like all fields, in constructor and inject the `component` in `applyOptions` method.

lib/android/app/src/main/java/com/reactnativenavigation/views/StackLayout.java
```diff
@@ -15,6 +20,8 @@
 public class StackLayout extends RelativeLayout implements TitleBarButton.OnClickListener {
 
     private final TopBar topBar;
+    private FabOptionsPresenter fabOptionsPresenter;
```
guyca: Should the stack contain `FabOptionsPresenter` ?
I think every node can contain a fab, so `FabOptionsPresenter` can be a property of `ViewController` and `fabOptionsPresenter.present(...)` can be called from `ViewController.applyOptions(Options options)`

playground/src/screens/OptionsScreen.js
```diff
@@ -63,9 +87,8 @@ class OptionsScreen extends Component {
         <Button title='Top Bar Transparent' onPress={this.onClickTopBarTransparent} />
         <Button title='Top Bar Opaque' onPress={this.onClickTopBarOpaque} />
         <Button title='scrollView Screen' testID={testIDs.SCROLLVIEW_SCREEN_BUTTON} onPress={this.onClickScrollViewScreen} />
-        <Button title='Custom Transition' testID={testIDs.CUSTOM_TRANSITION_BUTTON} onPress={this.onClickCustomTranstition} />
-        <Button title='Show custom alert' testID={testIDs.SHOW_CUSTOM_ALERT_BUTTON} onPress={this.onClickAlert} />
-        <Button title='Show snackbar' testID={testIDs.SHOW_SNACKBAR_BUTTON} onPress={this.onClickSnackbar} />
+        <Button title='Custom Transition' onPress={this.onClickCustomTranstition} />
+        <Button title='Hide fab' testID={testIDs.HIDE_FAB} onPress={this.onClickFab} />
```
guyca: Can we show this button only on Android?

lib/android/app/src/main/java/com/reactnativenavigation/views/FabMenu.java
```diff
@@ -0,0 +1,145 @@
...
+import static android.view.ViewGroup.LayoutParams.WRAP_CONTENT;
+import static android.widget.RelativeLayout.ALIGN_PARENT_BOTTOM;
+import static android.widget.RelativeLayout.ALIGN_PARENT_LEFT;
+import static android.widget.RelativeLayout.ALIGN_PARENT_RIGHT;
+import static android.widget.RelativeLayout.ALIGN_PARENT_TOP;
+import static com.reactnativenavigation.parse.Options.BooleanOptions.False;
+import static com.reactnativenavigation.parse.Options.BooleanOptions.True;
+
+
+public class FabMenu extends FloatingActionMenu implements FabAnimator {
```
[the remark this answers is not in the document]
Cool04ek: Unfortunately, no. It's a restriction, provided by the lib we use for Fab

playground/src/screens/OptionsScreen.js
```diff
@@ -180,19 +205,11 @@ class OptionsScreen extends Component {
...
-      text: 'Test!',
-      // textColor: 'red',
-      // backgroundColor: 'green',
-      duration: 'long',
-      button: {
-        text: 'Action',
-        textColor: 'blue'
+  onClickFab() {
+    Navigation.setOptions(this.props.componentId, {
+      fabMenu: {
```
guyca:
```js
Navigation.setOptions(this.props.componentId, {
  fab: {
    id: 'myFAB',
    backgroundColor: 'blue',
    actions: [
      {
         // an action which is displayed in a "menu", same api as main fab
      }
    ]
  }
});
```

This way we avoid having two entities, the actions (menu items) are a property of the fab.

Issue #2685, opened 2018-02-09
Stack doesn't work on Android


billyohgren:
### Issue Description

I'm trying to run an Android app with a StackViewController but all I get is a white screen with a topBar. Running this in iOS works though.

### Steps to Reproduce / Code Snippets / Screenshots
This won't work on Android (but works in iOS):
```Navigation.setRoot(
stack: {
    name: 'Test',
    children: [{
      component: {
        name: 'testScreen',
        passProps: {
        },
      },
    },
    ],
  },
)
```

This will display my testScreen on Android and iOS (but won't let me push because it's not a stack):
```
component: {
        name: 'swadd.testScreen',
        passProps: {
        },
     },
```
---
### Environment
* React Native Navigation version: 2.0.2119
* React Native version: 0.50.3
* Platform(s) (iOS, Android, or both?): Android
* Device info (Simulator/Device? OS version? Debug/Release?): Samsung S8. 7.0. Debug.

billyohgren:
After countless error of debugging I solved this by changing:
```getView().addView(enteringView);```
to
```getView().addView(enteringView, LayoutParams.MATCH_PARENT, LayoutParams.MATCH_PARENT);```
in StackController.java in both ```public void push(ViewController child, final Promise promise)``` and ```public void animatePush(final ViewController child, final Promise promise)```.

Since I'm no Android developer it would be nice to get some feedback on this fix, I'll create a PR and submit later today.

**Comment by RubenSandwich** [comment]
This is also required for the screen to resize back to full screen after a keyboard dismisses.

**Comment by guyca** [comment]
Hey @billyohgren 

The following works in the playground app
```js
    Navigation.setRoot({
      stack: {
        children: [
          {
            component: {
              name: 'navigation.playground.WelcomeScreen'
            }
          }
        ]
      }
    });
```

Can you please paste the complete api call?

viotti:
Similar problem here, but with `bottomTabs`. I even posted a question on Stack Overflow.

https://stackoverflow.com/q/48724828/2619107

I get a white screen with the bottom bar on Android.

dviluk:
same problem here, with stack based. I get a white screen and the navbar.

JohnSmith0602:
Same issue here with v2.0.2121, ios works fine, while android renders white screen and the navbar.

The api call is its simplest form:

javascript
```
Navigation.events().onAppLaunched(() => {
  Navigation.setRoot({
    stack: {
      children: [{
        component: {
[...]

PR #2686, opened 2018-02-09
Add match_parent layoutparam in stack controller

billyohgren:
This solves #2685


guyca:
Thanks @billyohgren , wil push a fix soon

PR #2691, opened 2018-02-09
Update v2 documentation 

billyohgren:
Some basic changes to the documentation, because starting with RNN today could be quite confusing.

Container is now Component.
Import statement is import { Navigation } from 'react-native-navigation'.
Added use case for stack.
I can't find any documentation on how to compile the documentation :P I tried to compile with Jekyll without any luck.

### DIFF
--- a/docs/_sidebar.md
+++ b/docs/_sidebar.md
@@ -6,7 +6,7 @@
  - [Top Level](/docs/Navigation)
 - Params
  - [Root](/docs/Root)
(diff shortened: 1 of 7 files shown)


guyca:
Hey there, thanks for taking the time to update the docs! The reason they are now obsolete is because the  docs scripts broke when we switched to TypeScript. @DanielZlotin Is working on a solution

Issue #1407, opened 2017-06-23
[TabBar] Specify font size


hose314:
Cannot find any mention on this property in the docs or it's not even exist?

---
### Environment
* React Native Navigation version: 1.1.91
* React Native version: 0.43.4
* Platform(s) (iOS, Android, or both?): Android
* Device info (Simulator/Device? OS version? Debug/Release?): all

ziyafenn:
Any luck finding an answer to control fontSize for tabBar?

bieooo:
If you guys didn't found yet, here is the solution

`
const Nav = TabNavigator ({
//Screens Here
}, { 
    tabBarOptions:{
        labelStyles:{
            fontSize:10
        }
}
});
`

@hose314 @ziyafenn

DaniyarJakupov:
@bieooo you're confusing libraries, this is not react-navigation.
@hose314 @ziyafenn on iOS this works: 
```
tabsStyle: { 
    tabBarTextFontSize: 18,
}
```

Coudn't find solution for android tho

alexey-belous:
I've added ability to specify fontSize by following properties of appStyle: tabFontSize and selectedTabFontSize.

thisisjaiswal:
tabBarOptions: {
      **labelStyle**: { fontSize: 12 },
}

works

buddhika-ranasinghe:
Thanks @thisisjaiswal 
**labelStyle:{ fontSize:25 }** worked for me

weikee94:
@thisisjaiswal @buddhika-ranasinghe  may i know what version u guys referencing ? v1 or v2 thanks.

umairkhan1997:
tabBarOptions: {
labelStyle: { fontSize: 12 },
}

this helps .

gerardramospear:
@thisisjaiswal that worked as well as for font family, was this documented anywhere? was wondering how you knew this because I couldn't find this until i saw your post

PR #2754, opened 2018-02-19
Added ability to specify font size for tab bar labels on Android

alexey-belous:
Solution for issue #1407


android/app/src/main/java/com/reactnativenavigation/params/parsers/StyleParamsParser.java
```diff
@@ -125,6 +127,8 @@ private StyleParams createDefaultStyleParams() {
         result.titleBarHideOnScroll = false;
         result.orientation = Orientation.auto;
         result.bottomTabFontFamily = new StyleParams.Font();
+        result.bottomTabFontSize = 10;
```
guyca: We should avoid hardcoding a default font size. Can we set font size only if specified by the user?

Issue #2435, opened 2017-12-28
V2 Animation Framework


bogobogo:
Create an animation framework that allows the user to define animations in the js side.
Current proposed API:

## API

```js
Navigation.push(this.props.componentId, {
  name: 'navigation.playground.CustomTransitionDestination',
  options: {
    animations: {
      push: {
        topBar: {
          id: 'STACK_ID' // Id of the stack containing the TopBar to animate
        },
        bottomTabs: {
          id: 'BOTTOM_TABS_ID'
        },
        content: {
          y: {
            from: 0,
            to: 200,
            duration: 150,
            interpolation: 'decelerate',
            startDelay: 100
          },
          alpha: {
            from: 0,
            to: 1,
            duration: 150
          }
        }
      },
      pop: {},
      showModal: {},
      dismissModal: {}
    }
  }
})
```

## Android
- [x] navigationOptions should be accessible through a ViewController #2440 
- [x] merge animations into navigationOptions
- [x] y
- [x] x
- [x] alpha
- [x] duration
- [x] interpolation
- [x] startDelay
- [x] Animate topBar and bottomTabs

* Support custom animations for the following commands:
- [x] push #2436 
- [x] pop #2437 
- [x] startApp #2438 
- [x] show/hide Modal #2790

* Support for default app animation style in `setDefaultOptions`

## iOS
- [x] navigationOptions should be accessible through a ViewController #2440 
- [x] merge animations into navigationOptions
- [x] y
- [x] x
- [x] alpha
- [x] duration
- [x] interpolation
- [x] startDelay
- [x] Animate topBar and bottomTabs

* Support custom animations for the following commands:
- [x] push
- [x] pop
- [x] ~~startApp~~ https://github.com/wix/react-native-navigation/issues/2901
- [x] showModal
- [x] dismissModal

* Support for default app animation style in `setDefaultOptions`

brunolemos:
iOS too I suppose?

kyle-ssg:
Quick question about this, I see that you can define push / pop / openModal etc animations from pushing to a page, this makes sense (especially defining a pop animation for a push).

More often than not though I could see the need to define the animation for just the screen I am transitioning to case e.g.

```Navigation.openModal(this.props.componentId, {
  name: 'navigation.playground.MyFadeModal',
..animationProps
})
```


Will this be possible too ?

guyca:
[...]

PR #2772, opened 2018-02-22
V2 anim

Cool04ek:
#2435


lib/android/app/src/main/java/com/reactnativenavigation/anim/NavigationAnimator.java
```diff
@@ -30,10 +39,12 @@ public NavigationAnimator(Context context) {
 
     public void animatePush(final View view, @Nullable final NavigationAnimationListener animationListener) {
         view.setVisibility(View.INVISIBLE);
-        ObjectAnimator alpha = ObjectAnimator.ofFloat(view, View.ALPHA, 0, 1);
-        alpha.setInterpolator(DECELERATE_INTERPOLATOR);
-
-        AnimatorSet set = new AnimatorSet();
+        AnimatorSet set;
+        if (options != null && !options.push.isEmpty()) {
```
guyca: Lets try to avoid null as much as possible. We can init `options` and `push`, effectively utilising the null object pattern which contributes to a more predictable code.

guyca: Furthermore, this logic should be extracted to `NavigationOptions`. Something like `options.getPushAnimation(view)`

lib/android/app/src/main/java/com/reactnativenavigation/anim/NavigationAnimator.java
```diff
@@ -47,19 +58,30 @@ public void onAnimationEnd(Animator animation) {
...
+        return set;
     }
 
     public void animatePop(View view, @Nullable final NavigationAnimationListener animationListener) {
-        ObjectAnimator alpha = ObjectAnimator.ofFloat(view, View.ALPHA, 1, 0);
-        ObjectAnimator translationY = ObjectAnimator.ofFloat(view, View.TRANSLATION_Y, 0, this.translationY);
-
-        AnimatorSet set = new AnimatorSet();
+        AnimatorSet set;
+        if (options != null && !options.pop.isEmpty()) {
```
guyca: Same here, animation creation logic should be part of the animator.

lib/android/app/src/main/java/com/reactnativenavigation/anim/NavigationAnimator.java
```diff
@@ -68,9 +90,81 @@ public void onAnimationEnd(Animator animation) {
...
-        set.setInterpolator(ACCELERATE_INTERPOLATOR);
         set.start();
     }
+
+    @NonNull
+    private AnimatorSet getDefaultPopAnimation(View view) {
+        AnimatorSet set = new AnimatorSet();
+
+        ObjectAnimator alpha = ObjectAnimator.ofFloat(view, View.ALPHA, 1, 0);
+        alpha.setInterpolator(ACCELERATE_INTERPOLATOR);
```
guyca: Is there a reason we don't set shared properties once on the animator set?

Cool04ek: Not sure I understand the question

lib/android/app/src/main/java/com/reactnativenavigation/parse/AnimationOptions.java
```diff
@@ -0,0 +1,68 @@
...
+        options.rotationY = ValueAnimationOptions.parse(json.optJSONObject("rotationY"));
+        options.rotation = ValueAnimationOptions.parse(json.optJSONObject("rotation"));
+        options.alpha = ValueAnimationOptions.parse(json.optJSONObject("alpha"));
+        options.scaleX = ValueAnimationOptions.parse(json.optJSONObject("scaleX"));
+        options.scaleY = ValueAnimationOptions.parse(json.optJSONObject("scaleY"));
+
+        return options;
+    }
+
+    private boolean empty = true;
```
guyca: Can we use positive (hasValue) instead of negative (isEmpty)? This help avoid double negation conditional which are very confusing.

lib/android/app/src/main/java/com/reactnativenavigation/anim/NavigationAnimator.java
```diff
@@ -68,9 +90,81 @@ public void onAnimationEnd(Animator animation) {
...
+        alpha.setDuration(DURATION);
+        set.playTogether(translationY, alpha);
+        return set;
+    }
+
+    public void setOptions(AnimationsOptions options) {
+        this.options = options;
+    }
+
+    private AnimatorSet createAnimation(AnimationOptions options, View view) {
```
guyca: Seems like there's a lot of duplication here.
Could have been cleaner if `AnimationsOptions` exposed a `toList()` function which returned all animations as list, then we would be able to iterate over them and create the appropriate animation with some factory method.
This is totally optional :)

Issue #2263, opened 2017-12-02
Linker error when installing for iOS


msloan21:
### Issue Description

I'm attempting to install react-native-navigation for iOS and I'm getting Linker Error when I perform a clean/build in xCode.

### Steps to Reproduce / Code Snippets / Screenshots

 - I've confirmed I have linked the navigation library
 - I've confirmed I have the correct path entered for 'Header Search Paths' and it's set to recursive (if I change the path to something else, I get the cant find RCCManager.h error)
 - I have updated AppDelegate.m
 - I have removed node_modules and run npm install
 - I can run the example without any issues

<img width="277" alt="screen shot 2017-12-02 at 3 32 18 pm" src="https://user-images.githubusercontent.com/991128/33519763-a904b24a-d77b-11e7-8659-cd198a87d610.png">

---
### Environment
* React Native Navigation version: 1.1.298
* React Native version: 0.50.4
* Platform(s) (iOS, Android, or both?): iOS
* Device info (Simulator/Device? OS version? Debug/Release?): Simulator

psegalen:
Hi, I encountered an error like this one and was able to fix it by adding libReactNativeNavigation.a to the "Linked Frameworks and Libraries" list in "General" tab of the target settings. Maybe it should be added as a step of the iOS installation?

psegalen:
I submitted PR #2793 to add this.

PR #2793, opened 2018-02-28
Added a missing iOS installation step

psegalen:
Hi, this should avoir anyone to encounter issue #2263.
I ran into the same problem and it seems that the additional step I added (add the lib to `Linked Frameworks and Libraries` is now mandatory.


yershalom:

PR #2797, opened 2018-03-01
Orientation android

guyca:


filipef101:
@guyca Does this fix the app before splashscreen starting horizontally if device rotated then going to vertical on android?

guyca:
Hey @filipef101 
This pr implements orientation style options in v2. If you found a bug in v2, please open an issue as I'm not sure I fully understand the issue you've described.

PR #2830, opened 2018-03-07
Removed instruction not more valuable

realtebo:
Actually, the reference AppDelegate.m is already using @index, so there is no more needs to suggest changes


yershalom:

Issue #676, opened 2017-01-21
Hide labels in the nav bar?


pjb3:
Is there a way to hide the labels in the nav bar, similar to how Facebook and Instagram do? I can set them to an empty string, but there are 2 problems:

1. When I set the title in one of the screens in that tab bar, it then sets the label
2. The spacing is off. The spot where the labels are supposed to be still seems to be there, pushing the icons up a little bit.

guyca:
1. refers only to iOS, right?
2. Is this relevant to both platforms?

pjb3:
Right, just iOS

Colmea:
FYI, the spacing issue is also on Android.

indivisable:
In your main app.js:

 ```
 tabsStyle: {
      tabBarShowLabels: 'hidden',
      tabBarButtonColor: '#999999',
      tabBarSelectedButtonColor: '#3FA2F7',
      //tabBarBackgroundColor: '#FFFFFF'
    },
```

Then in RCCTabBarController.m (Note the addition of the 'tabBarShowLabels') 

```
 NSDictionary *tabsStyle = props[@"style"];
  if (tabsStyle)
  {
    NSString *tabBarButtonColor = tabsStyle[@"tabBarButtonColor"];
    tabBarShowLabels = tabsStyle[@"tabBarShowLabels"];
    if (tabBarButtonColor)
    {
      UIColor *color = tabBarButtonColor != (id)[NSNull null] ? [RCTConvert UIColor:tabBarButtonColor] : nil;
      self.tabBar.tintColor = color;
      buttonColor = color;
      selectedButtonColor = color;
    }
    
    NSString *tabBarSelectedButtonColor = tabsStyle[@"tabBarSelectedButtonColor"];
    if (tabBarSelectedButtonColor)
    {
      UIColor *color = tabBarSelectedButtonColor != (id)[NSNull null] ? [RCTConvert UIColor:tabBarSelectedButtonColor] : nil;
      self.tabBar.tintColor = color;
      selectedButtonColor = color;
    }
    
    NSString *tabBarBackgroundColor = tabsStyle[@"tabBarBackgroundColor"];
    if (tabBarBackgroundColor)
    {
      UIColor *color = tabBarBackgroundColor != (id)[NSNull null] ? [RCTConvert UIColor:tabBarBackgroundColor] : nil;
      self.tabBar.barTintColor = color;
    }
  }
  else{
    tabBarShowLabels = @"shown";
  }

[...]

Issue #420, opened 2016-11-09
Tabbar iOS: center icon if no title/label


Crash--:
ATM, if we set a title & label to null, icons are not centered in the tabbar on iOS. 

I've made the change on my project, is it something you can be interested in? If yes, I can submit a PR 

before: 
![capture d ecran 2016-11-09 a 10 06 52](https://cloud.githubusercontent.com/assets/1107936/20132660/46f51098-a664-11e6-9f56-3961cd6d5a49.png)

after
![capture d ecran 2016-11-09 a 10 04 43](https://cloud.githubusercontent.com/assets/1107936/20132663/4a27a4b0-a664-11e6-929c-4ebc20fbdf03.png)

DesignMonkey:
👍  for PR

But shouldn't it just be label that affect the tabbar. Title is used as the topheader in the navigation and not in the tabbar

guyca:
@Crash-- Thanks a lot! Please submit a PR 👍

perrosnk:
related to this #311

PR #2857, opened 2018-03-12
When no label is passed in, the icon is centered

KabirKang:
When you don't pass in a label to a tab item, the UI acts as if an empty label is there, so the icons look as if they are positioned incorrectly.

screen shot 2018-03-12 at 4 17 05 pm

The fix checks if a label was provided to the tab, and if it was not provided, it positions the icon in a more natural position:

screen shot 2018-03-12 at 4 17 01 pm

When labels are passed in, the display looks as usual.

screen shot 2018-03-12 at 4 19 42 pm

Related Issues:
https://github.com/wix/react-native-navigation/issues/420
https://github.com/wix/react-native-navigation/issues/676

### DIFF
--- a/ios/RCCTabBarController.m
+++ b/ios/RCCTabBarController.m
@@ -170,9 +170,19 @@
       iconImageSelected = [RCTConvert UIImage:icon];
     }
     
-    viewController.tabBarItem = [[UITabBarItem alloc] initWithTitle:title image:iconImage tag:0];
-    viewController.tabBarItem.accessibilityIdentifier = tabItemLayout[@"props"][@"testID"];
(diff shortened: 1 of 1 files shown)


guyca:
Hey @KabirKang 
Thanks for the PR. I believe we discussed this matter internally a while back and decided against such solution as it relies on magic numbers and might break on new OS versions.

You can use `iconInsets` [property](https://wix.github.io/react-native-navigation/#/top-level-api?id=starttabbasedappparams) for now.

KabirKang:
@guyca what if I modified it so that the numbers are not magic but configurable?

PR #2869, opened 2018-03-15
V2 titlebar changes

frankenthumbs:
since views are defautly white/black.. there is no way to make customNavigationTitleView and customNavigationBarView transparent. If we make the views clear, this allows for react-native screen to style its parent component to any opacity in javascript.


frankenthumbs:
sorry guys.. i mean to do this in the nfl fork.. will put up a PR for wix

Issue #546, opened 2016-12-08
Documentation issue: FB Navigator comparison no longer exists?


osman-masood:
Hello,

The Facebook page which apparently said that `NavigatorIOS` is deprecated, no longer exists: 
https://facebook.github.io/react-native/docs/navigator-comparison.html#content

Additionally, there is no information on the `NavigatorIOS` documentation to indicate that it is deprecated, as the README here suggests: https://facebook.github.io/react-native/docs/navigatorios.html

Perhaps the README should be clarified to include up-to-date information as to whether `NavigatorIOS` is in fact deprecated? Because it looks odd if Facebook does not claim it is deprecated, yet the authors of this library do.

PR #2890, opened 2018-03-19
Updating README.md

lborgav:
The following text is removed: "For example, this package replaces the native NavigatorIOS that has been abandoned in favor of JS-based solutions that are easier to maintain.", because it's not accurate.


DanielZlotin:
Aside from the broken link (see https://github.com/wix/react-native-navigation/issues/546), why is this not accurate?

lborgav:
Sorry, the broken link is the only problem.

PR #2907, opened 2018-03-23
Fix README.md (v2)

bramus:
The tables had some issues: the header for contributor column wasn't there (with the contributors still in place). I've re-added the headers where needed.

_(Other fix could be to remove the contributors if that's. Don't know which style you are following.)_


guyca:
Hey @bramus 
Thanks a lot for the contribution! Actually I'd like to remove the contributors column. I just removed the column's header, wasn't aware it's messed up on GitHub, looked fine in MacDown

PR #2917, opened 2018-03-26
Stack id

Cool04ek:


lib/android/app/src/main/java/com/reactnativenavigation/anim/TopBarAnimator.java
```diff
@@ -21,15 +23,23 @@
...
+    public TopBarAnimator(TopBar topBar, @Nullable String stackId) {
+        this.topBar = topBar;
+        this.stackId = stackId;
+    }
+
     public void show(AnimationOptions options) {
-        if (options.hasValue()) {
+        if (options.hasValue() && options.id.hasValue() && options.id.get().equals(stackId)) {
+            showAnimator = options.getAnimation(topBar);
+        } else if (options.hasValue() && !options.id.hasValue()) {
```
guyca: Isn't `options.id.hasValue()` always `false` here?

lib/android/app/src/main/java/com/reactnativenavigation/anim/TopBarAnimator.java
```diff
@@ -21,15 +23,23 @@
...
 
+    public TopBarAnimator(TopBar topBar, @Nullable String stackId) {
+        this.topBar = topBar;
+        this.stackId = stackId;
+    }
+
     public void show(AnimationOptions options) {
-        if (options.hasValue()) {
+        if (options.hasValue() && options.id.hasValue() && options.id.get().equals(stackId)) {
+            showAnimator = options.getAnimation(topBar);
```
guyca: Line 41 equals line 43... is this conditional needed?

lib/android/app/src/main/java/com/reactnativenavigation/anim/TopBarAnimator.java
```diff
@@ -62,7 +72,9 @@ private AnimatorSet getDefaultShowAnimator(float startTranslation, TimeInterpola
     }
 
     public void hide(AnimationOptions options, AnimationListener listener) {
-        if (options.hasValue()) {
+        if (options.hasValue() && options.id.hasValue() && options.id.get().equals(stackId)) {
+            hideAnimator = options.getAnimation(topBar);
```
guyca: line 76 is the same as line 78, is this conditional required?

lib/android/app/src/main/java/com/reactnativenavigation/anim/TopBarAnimator.java
Cool04ek: No. The difference is, that first condition is true, if you have provided id and it's the same. And second condition true if you didn't provide id at all

lib/android/app/src/main/java/com/reactnativenavigation/anim/TopBarAnimator.java
Cool04ek: Same here

PR #2922, opened 2018-03-27
Fixed build error: undefined identifier RCTBridgeWillReloadNotificati…

### LINKED ISSUE #2913: use of undefined identifier RCTBridgeWillReloadNotification in RCCManager.m
### Issue Description

Fails to build:

```
~/react_apps/core/node_modules/react-native-navigation/ios/RCCManager.m:39:100: Use of undeclared identifier 'RCTBridgeWillReloadNotification'
```

Can get around this by defining the string in RCCManager.m:

```
NSString *const RCTBridgeWillReloadNotification = @"RCTBridgeWillReloadNotification";
```

Build using current master, I do get the problem going back to 1409

---
### Environment
* React Native Navigation version: `1.1.409` and up
* React Native version: `0.47.1`
* Platform(s) (iOS, Android, or both?): `iOS`
* Device info (Simulator/Device? OS version? Debug/Release?): `Simulator and device 11.2`

bjacog:
…on in RCCManager.m

Fixes #2913 

Added the NSString const that line 41 requires. Not sure why it was missing, but the line in question was commented out until recently.

I'm not 100% familiar with `NSNotificationCenter` observers, so if this should really be commented out you are welcome to reject it.

### DIFF
--- a/ios/RCCManager.m
+++ b/ios/RCCManager.m
@@ -12,6 +12,8 @@
 @property (nonatomic, strong, readwrite) NSDictionary *globalAppStyle;
 @end
 
+NSString *const RCTBridgeWillReloadNotification = @"RCTBridgeWillReloadNotification";
+    
 @implementation RCCManager
 
 + (instancetype)sharedInstance {


yogevbd:
It look like your'e using an old react-native version.
We officially support  react-native >= 0.48
This patch will cause debugging issues but shouldn't affect your app in release mode.

bjacog:
@yogevbd thanks, now I know what that was being used for. I will update my RN version and see if the issue is still there.

witalobenicio:
@yogevbd my RN version is > 0.48 and I'm getting the same error.

PR #2959, opened 2018-04-02
Reference to function confusing and unnecessary

switch13:
Removed reference to - (BOOL)application:(UIApplication *)application... The way it's worded makes it seem like you only need to overwrite what's in the function, not the entire file's code.


switch13:
Sorry, not sure how to properly request this.

PR #2975, opened 2018-04-04
V1 [iOS/Android] Feature/async dismiss modal

Crash--:
ATM, when we call dismissAllModal we can not be sure when all modals are really dismissed. 

I've a very simple use case in my app : 
- I've several opened modals (2) 
- When a condition is reached, I need to dismiss these 2 modals and display a new one

Before this patch, I managed this use case with a simple timeout() but it was not pretty safe and bug prone. Now, I just have to `await dismissAllModal()` before runing `showModal()` again.

This feature is done for both Android and iOS.

iOS : On the native side, async stuff was already done. Just edited the JS function to use async / await. 

Android: Nothing was done for that. I added some promise stuff, it works pretty well in my app but don't know if it's the best way to achieve that.


guyca:
Hey @Crash-- Can you please revert indentation changes from the PR and keep only relevant code? Will help a lot getting this in.

Issue #2695, opened 2018-02-10
NPE in NavigationReactEventEmitter.java


varungupta85:
I am observing multiple reports for below crash in the Google Play console for my app.

```java
java.lang.NullPointerException: 
  at com.reactnativenavigation.bridge.NavigationReactEventEmitter.<init> (NavigationReactEventEmitter.java:17)
  at com.reactnativenavigation.react.NavigationReactGateway.onReactContextInitialized (NavigationReactGateway.java:114)
  at com.reactnativenavigation.react.NavigationReactGateway.access$000 (NavigationReactGateway.java:24)
  at com.reactnativenavigation.react.NavigationReactGateway$ReactNativeHostImpl.onReactContextInitialized (NavigationReactGateway.java:174)
  at com.facebook.react.ReactInstanceManager$6.run (ReactInstanceManager.java:1017)
  at android.os.Handler.handleCallback (Handler.java:742)
  at android.os.Handler.dispatchMessage (Handler.java:95)
  at android.os.Looper.loop (Looper.java:154)
  at android.app.ActivityThread.main (ActivityThread.java:5527)
  at java.lang.reflect.Method.invoke (Native Method)
  at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run (ZygoteInit.java:739)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:629)
```

I don't know what are the steps to reproduce. Below is the line in code on which the abort is occurring

```java
public NavigationReactEventEmitter(ReactContext reactContext) {
        this.eventEmitter = reactContext.getJSModule(RCTDeviceEventEmitter.class);
}
```

So, it looks like the `reactContext` is null but I am not sure why would it be null.

---
### Environment
* React Native Navigation version: 1.1.365
* React Native version: 0.52.4
* Platform(s) (iOS, Android, or both?): Android
* Device info (Simulator/Device? OS version? Debug/Release?): Device / Release

guyca:
hmm that's really strange. I checked our crash logs and couldn't find a similar crash.
The event emitter is instantiated in `onReactContextInitialized`, one would assume react context won't be null at this point.

varungupta85:
Yes. That confused me also. It is visible in the call stack also that the event emitter in the `onReactContextInitialized`, so it shouldn't be null which is why I didn't want to just add a null check there.

marcmo:
I experience the same issue with RN 0.50.3 and RNN 1.1.389
```
[...]

PR #2976, opened 2018-04-04
Fixing NPE found on issue #2695

dvxme:
There's a roundabout manner of obtaining the react context on initialisation that is null on some instances (https://github.com/wix/react-native-navigation/issues/2695), including errors on startup that prevents debugging on console. (For instance, throwing an error on index before starting the app)

It is solved by sending the same context given on method onReactContextInitialized, instead of discarding it and calling getReactContext.


guyca:
Thanks, I'll merge after the weekend 👍

guyca:
It's already published, each commit to master is published

PR #2982, opened 2018-04-05
Fix custom title bar issues

frankenthumbs:
(1) prevent `customNavigationTitleView` / `customNavigationBarView` / `customNavigationComponentBackground` from being mounted twice when `viewWillAppear` is being called in `RNNRootViewController`
(2) `prevent customNavigationTitleView` / `customNavigationBarView` / `customNavigationComponentBackground` from being remounted every time the new will appear again.


yogevbd:
Moved applying options to the viewController's `viewDidLoad` function which make more sense.

PR #2984, opened 2018-04-06
Improved the clarity of the top-level API doc

johngoren:
I clarified some of the language here so that it is easier for the beginner to understand. (It was a little bit awkward before)


guyca:
Thanks @johngoren 👍

PR #2991, opened 2018-04-07
Migrate to react v16.3 lifecycle

nicastelo:
- Migrate to react v16.3 lifecycle 
- Add backward compatibility for older React versions


lib/src/components/ComponentWrapper.tsx
```diff
@@ -73,6 +74,8 @@ export class ComponentWrapper {
       }
     }
 
+    ReactLifecyclesCompat.polyfill(WrappedComponent);
```
DanielZlotin: Is this the only way to support backwards compatibility? I would really like not having another dependency, and polyfilling seems like overkill..

lib/src/components/ComponentWrapper.tsx
```diff
@@ -15,11 +22,11 @@ export class ComponentWrapper {
         this._saveComponentRef = this._saveComponentRef.bind(this);
         this.state = {
           componentId: props.componentId,
-          allProps: _.merge({}, props, store.getPropsForId(props.componentId))
```
DanielZlotin: I believe we need `props` and `id` here as well, but I guess a unit test will need to prove it..

lib/src/components/ComponentWrapper.tsx
nicastelo: This is the way proposed by the react team. They talk a bit about it in this post https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html#open-source-project-maintainers.

The other thing that comes to my mind is to have two different WrappedComponent, one with the new lifecycle and another with the old one. Then we use the proper one based on the React version.

lib/src/components/ComponentWrapper.tsx
DanielZlotin: 👍

lib/src/components/ComponentWrapper.tsx
DanielZlotin: 👍

DanielZlotin:
There was a problem with the build machines, it's fixed now

Issue #1928, opened 2017-09-25
rightButtons on Android attempt to set key "enabled" on an immutable/frozen object


sqow:
### Issue Description

Attempting to set `rightButtons` on the `navigatorButtons` object of the `screen` object within, at least, a `startSingleScreenApp` call on Android (6.0.1+)  results in the following error:

```
You attempted to set the key `enabled` with the value `true` on an object that is meant to be immutable and has been frozen.
```

### Steps to Reproduce / Code Snippets / Screenshots

```js
const navigatorStyle = 

Navigation.startSingleScreenApp({
  screen: {
    screen: 'com.example.MyScreen',
    navigatorStyle: {
      statusBarColor: '#831d19',
      navigationBarColor: '#339999',
      navBarBackgroundColor: '#339999',
      navBarTextColor: '#ffffff',
      navBarButtonColor: '#ffffff',
      statusBarTextColorScheme: 'light',
      navBarHidden: false,
      tabBarButtonColor: 'red',
      tabBarSelectedButtonColor: 'red',
      tabBarBackgroundColor: 'red'
    },
    navigatorButtons: {
      rightButtons: [
        {
          id: 'sideMenu'
        }
      ]
    }
  },
  drawer: {
    right: {
      screen: 'com.example.Drawer'
    }
  }
})
```

This relates to [this line](https://github.com/wix/react-native-navigation/blob/master/src/deprecated/platformSpecificDeprecated.android.js#L524) being used in the array of `rightButtons` (plural), while the [left button (singular) block](https://github.com/wix/react-native-navigation/blob/master/src/deprecated/platformSpecificDeprecated.android.js#L537-L548) has no comparable setting of an `enabled` property.

---
### Environment
* React Native Navigation version: 1.1.228
* React Native version: 0.47.0
* Platform(s) (iOS, Android, or both?): Android, not tested on iOS
* Device info (Simulator/Device? OS version? Debug/Release?): Samsung Galaxy S6 running 6.0.1, debug

HoogsterInc:
I have this too. Works on iOS but not android

tschoffelen:
This seems pretty major, but doesn't seem to get much attention. I'm facing the same issue, any solutions found yet?

cqlanus:
I'm experiencing this, as well. The error cites [this line](https://github.com/wix/react-native-navigation/blob/master/src/deprecated/platformSpecificDeprecated.android.js#L565) as an issue

[...]

PR #3032, opened 2018-04-12
Fix/v1 rightbuttons immutable error

adrienthiery:
See #1928


tschoffelen:
This should be merged! Still a huge issue on Android.

adrienthiery:
@guyca any feedback on this? It got closed automatically but is that not relevant?

PR #3051, opened 2018-04-15
Add Android support for rootBackgroundImageName

ywongweb:
Add Android support for rootBackgroundImageName


FRizzonelli:
Hi :)

Just a question, why the name of the property is now 'rootBackgroundImageName'? I've pulled master but it's not working.. Am I missing something?

kayzenkayzen:
its screenBackgroundImageName supported on android ? or just only rootBackgroundImageName?

PR #3072, opened 2018-04-18
fix IndexOutOfBoundsException

andresesfm:
IOOBE occurs when selected index is -1 i.e. nothing should be selected by default. This adds a simple check before setting the value


andresesfm:
simplified in https://github.com/wix/react-native-navigation/pull/3082

PR #3115, opened 2018-04-27
Avoid NPE in SET reported in #2576

varungupta85:


guyca:
Thanks @varungupta85 👍

PR #3130, opened 2018-04-30
WIP - allow tabBar to set selected index

frankenthumbs:
api is up for discussion. some more work needs to be done around `setRoot` and other apis passing back a componentId.

i'm not sure how this would work for android.
Basically the code on the client APIs would be something like this
```
const componentId = await Navigation.showModal({
    bottomTabs: {...}
});

// show the 3rd tab here
Navigation.setTabIndex(componentId, 2);
```

@andresesfm can you take a look and see if that sounds like something we can do on android?
and can somebody from wix let me know if i'm going down the correct rabbit hole?


guyca:
Hey @frankenthumbs 
Selecting tabs programatically should already be possible using `currentTabIndex` [options](https://github.com/wix/react-native-navigation/blob/v2/playground/src/screens/TextScreen.js#L67). Is this enough for you use case?

frankenthumbs:
@guyca hrmm.. good call.. didn't think of that.. lemme see if that works.

guyca:
I assume it did work, Frank ping me if it doesn't.

PR #3132, opened 2018-04-30
Fixed crash caused by async operation

andresesfm:
Removed redundant checks for `testId.hasValue()` and made sure the view is still available by the time the async process for getting the image is done


guyca:
Hey @andresesfm 
Thanks for the PR. What exactly is the async operation causing the crash? Are you able to reproduce it?
I don't mind merging this, but does this actually fix the problem? It seems that instead of crashing, the view testId won't be set on the tab item - am I correct?

PR #3128, opened 2018-04-30
Fixes NPE when initializing twice

andresesfm:
When refreshing multiple times, the asynchronous process that creates React Navigation causes a crash. Looking at the code, there was a lint warning as the context that was being used might be null (as it's being recreated). This PR addresses the issue by passing the freshly created process and checking for flag before emitting event

PR #3135, opened 2018-05-01
Revert "Fixes NPE when initializing twice"

guyca:
Reverts wix/react-native-navigation#3128

`ApplicationLifecycleTest.test.js` broke after merging this


guyca:
@andresesfm I think we didn't pick this regressing in CI as it runs e2e in release mode - which reload scenarios are not relevant to.

PR #3247, opened 2018-05-22
[v2] UISplitViewController

birkir:
This PR implements a new LayoutType that enables UISplitViewController.

 - [x] Implement `splitView` LayoutType
 - [x] Implement `RNNSplitViewController`
 - [x] Implement `RNNSplitViewOptions`
 - [x] Support all layout types in master and detail view.
 - [x] Support `Navigation.setStackRoot` on both master and detail to replace the stack.
 - [x] Support `Navigation.mergeOptions` for changing it's options.
 - [x] Playground app: Set `Universal` flag in XCode.
 - [x] Playground app: Add splitView demo to home screen
 - [x] JS Coverage
 - [x] Documentation

Future questions and thoughts:
 - Android doesn't have anything called Split View. It can be achieved with Fragments but this is not an API that is standard. See https://github.com/codepath/android-master-detail-demo
 - Write ObjC tests. This can be done in a separate pull-request once merged and stabilized.

Screenshot of iPad


guyca:
Hey @birkir, Thanks for the PR ❤️🎉🎊 

What are your thought on the following api
```js
const splitView = {
  master: {

  },
  detail: {
    stack: {
      id: 'DETAILS_ID',
      children: [
      ]
    }
  },
  options: {

  }
}
```

In order to **replace** the layout presented in `details`, we'll need to support a new api
```js
Navigation.replace('DETAILS_ID', {
  component: {

  }
});
```

The `SplitView` should be generic and support all layout types as `master` and `details` views (as long as they are supported by the OS)

birkir:
Of course, that makes sense, thanks for the feedback. I didn't research the terms for UISplitViewController enough to come up with master/detail.

I'll try making this work without changing the internals a lot.

birkir:
Ok, ended up using `Navigation.mergeOptions` for that.

So we can now do this to dynamically show/hide the master.

```js
Navigation.mergeOptions('SPLITVIEW_ID', {
  displayMode: 'hidden',
});
```

birkir:
For some reason, a test totally unrelated to this PR is failing. I removed it to check and see if the rest of the PR is checking out.

In theory is `Navigation.dismissModal()` dismissing all other modals than the current one...

guyca:
Hey @birkir 
I fixed the test today. regarding replace, I agree we can use stacks for now and add the missing api later.
cc @yogevbd

guyca:
@birkir 
That's some detox flakyness we need to address. I think an unrelated system alert popped up on the screen which caused the app to lose focus, as a result detox timed out. Rerunning the build now..

birkir:
Yes, the failing test looks like is consistently now.

master is also failing but with different error

guyca:
I'll look into this issue today, it happens locally as well from time to time.

PR #3251, opened 2018-05-23
Smarter set root

yogevbd:


guyca:
@ebouJ [The new api is used in the playground app](https://github.com/wix/react-native-navigation/blob/v2/playground/src/app.js#L137)

PR #3273, opened 2018-05-29
[v2] Preview peek and pop (iOS)

birkir:
Work is complete.

 - [x] preview.elementId has to be set for preview to work
 - [x] preview.height set's the height of preview
 - [x] preview.width should set the width of preview (zoom behaviour)
 - [x] preview.actions sets actions available in preview
 - [x] Send "previewContext" event to `onNavigationEvent` when previewing
 - [x] Send "previewCommit" event to `onNavigationEvent` when committing if preview.commit = 1
 - [x] Push new view controller without animation when committing if preview.commit = 1
 - [x] Send "previewDismissed" event  to `onNavigationEvent` when committing if preview.commit = 0


birkir:
List of things:
 - [x] preview.elementId has to be set for preview to work.
 - [x] preview.height set's the height of preview.
 - [ ] preview.width should set the width?
 - [x] Send "preview" event to `onNavigationEvent` when previewing.
 - [x] Send "previewCommit" event to `onNavigationEvent` when committing preview if preview.commit = true.
 - [x] Send "previewDismissed" event  to `onNavigationEvent` when dismssing preview if preview.commit = false.
 - [x] preview.actions sets actions available in preview.

lib/ios/RNNRootViewController.m
```diff
@@ -237,6 +237,58 @@ - (void)orientationDidChange:(NSNotification*)notification {
 	}
 }
 
+- (UIViewController *)previewingContext:(id<UIViewControllerPreviewing>)previewingContext viewControllerForLocation:(CGPoint)location{
+	return self.previewController;
```
ptmt: Implemented something like this PR myself (only without Navigation.Element). Looks good! 

The only question I have is about cancelling touches after previewing is initiated in `onPressIn`. E.g. try to add `onPress` to `Button`. I managed to do a workaround by subscribing to the global event emitter and cancelling touches on js side, but this required additional event `previewControllerAppears` or something like this fired from here. This is kinda messy. 

The other approach would be handling this on native side here. Any ideas?

birkir: Right, we have to be able to support onPress and 3D Press simultaniously.

We have all cases covered but cancelling a preview, we could access that through UIPreviewInteractionDelegate, but reading from StackOverflow, this API is useless after iOS 11.

So what I have in mind is to simply steal the responder lock of the button when previewing.

I'll come up with something, but I am not sure if we should ship the code by default, unless it gets really messy.

ptmt: Thanks! I have my solution worked through handling js events, just needed to add one more event (preview controller  appeared) and add `fromComponentId` for the view controller that is responsible for showing preview controller: 

image

Pretty sure it might be overhead, but seems that an additional event could add some value.

birkir: Allright and then you can easily create a boolean state of the preview.

```js
class Foo extends React.Component {

  state = {
    previewComponentId: null,
    disabled: false,
  };

  listeners = [];

  componentWillMount() {
    this.listeners.push(
      Navigation.events().registerNativeEventListener((name, params) => {
        if (name === 'previewContext') {
          const { previewComponentId } = params;
          this.setState({ previewComponentId });
        }
      }),
      Navigation.events().registerComponentDidAppearListener((componentId, componentName) => {
        if (this.state.previewComponentId === componentId) {
          this.setState({ disabled: true });
        }
      }),
      Navigation.events().registerComponentDidDisappearListener((componentId, componentName) => {
        if (this.state.previewComponentId === componentId) {
          this.setState({ disabled: false, previewComponentId: null });
        }
      })
    );
  }

  componentWillUnmount() {
    this.listeners.forEach(listener => listener.remove && listener.remove());
  }

  onPress() {
    if (this.state.disabled) return;
    // Normal push
  }

  onPressIn() {
    // Navigation preview
  }

  render() {
    return (
      <Button onPress={this.onPress} onPressIn={this.onPressIn} title="Push screen" />
    );
  }

```


Any idea how we can wrap this nicely into an touchable package or should we just leave it up to the users to implement something like this? I think the later because you may want to disable ScrollView scroll, do some animations etc.

birkir:
Work is complete.

New events:
 - previewContext (componentId, previewComponentId)
 - previewCommit (componentId, previewComponentId)
 - previewDismissed (componentId, previewComponentId)

lib/ios/RNNRootViewController.m
ptmt: Nice! I think it can be done in userspace now, as you just did (mine impl is similar, minor diffs). After more real-world usage someone may come up with nicer API, or we may solve this in native realm.

PR #3296, opened 2018-06-04
V2 add additional ios options for tabbar icons

cshen4:
added selectedIcon, disableIconTint, disableSelectedIconTint styling props


guyca:
@cshen4 Thanks for the PR. We're in the middle of migrating the Wix app to v2, it might take us a few days to process the PR. Thanks for the understanding 👍

Issue #3171, opened 2018-05-07
[v2] Error in pushing component


heroic:
### Issue Description

I am quite sure I haven't made any changes to my code, but a recent npm install, with all versions remaining same, Navigation.push is throwing a warning:

> push should be called from a stack child component

I am not even sure what this means! Here's the layout I am using right now:

  ```js
LoginStack: {
    stack: {
      children: [{
        component: {
          name: 'Login'
        }
      }]
    }
  }

```

And this is what I am using to push:

```js
Navigation.push(this.props.componentId, {
  component: {
    name: 'OTP',
    passProps: {phone: this.state.phone}
  }
})
```
### Steps to Reproduce / Code Snippets / Screenshots

![img_681c89cb4aa6-1](https://user-images.githubusercontent.com/523335/39719647-76c78dc8-5257-11e8-8b3a-2637a2f284ba.jpeg)

---
### Environment
* React Native Navigation version: 2.0.2274
* React Native version: 0.55.3
* Platform(s) (iOS, Android, or both?): iOS
* Device info (Simulator/Device? OS version? Debug/Release?): Device

guyca:
You're calling `Navigation.push` with a componentId which is not in a stack. You can assign a predefined ID to the stack and use it instead of `this.props.componentId` from props.

heroic:
That is not possible, as the line I am getting this error from is in the "Login" component, which as you can see in the Layout above, is a child in the stack!

guyca:
@heroic Can you please PR a failing e2e test to the v2 branch? If you're not comfortable writing a Detox test, just PR the problematic layout to the playground app so I can look into this issue.

ramonallday:
Im getting the same error

`Navigation.setRoot({
      root: {
        stack: {
          id: 'MyStack',
          children: [
            {
              component: {
                name: 'SIGNUP',
              },
            },
            {
              component: {
                name: 'LOGIN',
              },
            },
          ],
        },
      },
    })`

Then after async action executing

guyca:
@ramonallday If you believe this is a bug in RNN, Please PR a failing e2e test.

ramonallday:
@guyca https://github.com/wix/react-native-navigation/pull/3310 I hope this makes sense what I am trying to accomplish

guyca:
@yogevbd [Here's a test reproducing the issue](https://github.com/wix/react-native-navigation/blob/v2/e2e/ScreenStack.test.js#L102)

PR #3310, opened 2018-06-06
Push screen after async action

ramonallday:
Mentioned problem in here https://github.com/wix/react-native-navigation/issues/3171

I am not comfortable writing Detox tests, I am just seeking an advice how to push screen after async action.  

I would like to push new screen to stack after user logs in, but right now I am getting **push should be called from a stack child component**


guyca:
@ramonallday 
Thanks for the PR, I've added a [failing test](https://github.com/wix/react-native-navigation/blob/v2/e2e/ScreenStack.test.js#L102), once we fix it on iOS your issue will be fixed as well 👍

Issue #2331, opened 2017-12-12
Restarting app via instanceManager.recreateReactContextInBackground leads to crash


ruslan-bikkinin:
### Issue Description

We are working on [react-native-code-push](https://github.com/avishayil/react-native-code-push) module for CodePush and facing troubles investigating [this](https://github.com/Microsoft/react-native-code-push/issues/1088).
Briefly, we are providing API method for "soft" app restart (e.g. without killing application) using RN context restarting mechanism based on [`instanceManager.recreateReactContextInBackground();` method of `ReactInstanceManager` class](https://github.com/avishayil/react-native-restart/blob/master/android/src/main/java/com/avishayil/rnrestart/ReactNativeRestart.java#L55) and for some RNN scenarios this approach leads to the app crashes.
I am asking for assistance from you guys to help us to figure out what could be the source of problem here because restarting breaks only for `react-native-navigation` apps and works like a charm with usual RN app.

### Steps to Reproduce / Code Snippets / Screenshots

`App.js`:

```javascript
import { AsyncStorage, AppRegistry } from 'react-native';
import { Navigation } from 'react-native-navigation';
import React from 'react';
import { Text } from 'react-native';
import RNRestart from "react-native-restart";

const App = () =>
	<Text onPress={() => RNRestart.Restart()}>
		Touch to run RNRestart.Restart and see the crash
	</Text>
;

const screenName = Math.random().toString();

Navigation.registerComponent(screenName, () => App);
//check what components was registered
console.log("Registered components: " + AppRegistry.getAppKeys());
Navigation.startSingleScreenApp({
	screen: {
		screen: screenName,
	},
	passProps: { },
});
```

1. Download [demo project](https://github.com/ruslan-bikkinin/rnnCrashApp) and run `npm i`. It uses [react-native-restart](https://github.com/avishayil/react-native-restart) package that contains pure restart code API extracted from `react-native-code-push`.
2. Build it using `./gradlew assembleDebug` from `android` directory
3. Install apk on simulator
4. Ensure there is something like this in Logcat logs:

![image](https://user-images.githubusercontent.com/30265843/33899558-ee52fd1c-df7c-11e7-9d9e-7c7d50197721.png)

5. Click on button within app.
6. Observe the crash and ensure there is something like this in Logcat logs:

![image](https://user-images.githubusercontent.com/30265843/33899587-01fc2622-df7d-11e7-8b78-a9e5898e9167.png)

**Expected Behavior:**
[...]

PR #3318, opened 2018-06-07
Destroy navigation activity layouts on catalyst instance destroy

ericketts:
The changes in this PR are to address #2331 & Microsoft/react-native-code-push#1144.

I've tried it on a few devices and it seems to consistently prevent the crash from occurring; that said, I'm a bit wary that the `runOnUiThread` is introducing a race condition that may fail to prevent the bug under the right circumstances. Any thoughts on tackling this (potential) issue?

Edit: I should also mention I'm aware that V1 PRs are not being accepted at this time, but wanted to submit this at least for others that are struggling with this bug to implement the solution on their forks.

### DIFF
--- a/android/app/src/main/java/com/reactnativenavigation/bridge/NavigationReactModule.java
+++ b/android/app/src/main/java/com/reactnativenavigation/bridge/NavigationReactModule.java
@@ -9,6 +9,7 @@ import com.facebook.react.bridge.ReactContextBaseJavaModule;
 import com.facebook.react.bridge.ReactMethod;
 import com.facebook.react.bridge.ReadableArray;
 import com.facebook.react.bridge.ReadableMap;
+import com.reactnativenavigation.controllers.NavigationActivity;
 import com.reactnativenavigation.controllers.NavigationCommandsHandler;
 import com.reactnativenavigation.params.ContextualMenuParams;
 import com.reactnativenavigation.params.FabParams;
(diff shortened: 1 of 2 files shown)


ericketts:
awesome to see this merged guys, thanks a bunch for taking the time to review!

unfortunately V2 seems to suffer from the same issue but this solution will not work for it (afaik V2 doesn't keep a statically accessible ref to the activity); given our use case I'll need to fix that eventually as well and will PR it too whenever I get time to work on it.

levynir:
Thanks @guyca & @ericketts , drinks on me when any of you come to Hertzelia, Israel anytime soon .

guyca:
@ericketts Thank you. If you could write a failing detox test for v2, we'll fix this asap. @levynir drinks on me if you manage to come to our [meetup](https://www.meetup.com/React-Native-TLV/events/255511586/) 👍

ericketts:
@levynir no plans on the horizon but certainly somewhere I've wanted to visit for a while now, thanks much for the offer its very generous. 

@guyca to your point about a failing detox test, I'm not sure what the best way to do this would be, since as far as I can tell we'd need to either add the restart lib or a command to RNN that does what the restart lib does (ie calling `ReactInstaceManager#recreateReactContextInBackground()`) to properly recreate the crash scenario. do either of these seem like viable options, or any ideas for other options I may have missed?

as far as a fix though, I've adapted this fix to V2 and so far in my testing I have been crash free so I may have something for that soon.

PR #3321, opened 2018-06-08
Don't break buttons that are already set when setting only one side.

donataswix:


donataswix:
I honestly did not not what I was doing except when writing integration test, so please merge and fix this code :) Did not check android.

Issue #3319, opened 2018-06-07
[V2] Default background color doesn't work as specified in the documentation


emiyake:
### Issue Description

When configuring a default background  (`Navigation.setDefaultOptions`) color as specified in the documentation, nothing happens. 

```
 {
  layout: {
   backgroundColor: 'red'
  }
}
```

Reference: https://wix.github.io/react-native-navigation/v2/#/docs/styling?id=common-options

It turns out it works when adding, instead:

```
{
  screenBackgroundColor: 'red'
}
```

 Setting the color using `Navigation.setDefaultOptions` only works on Android. But it works on iOS when defining this property per screen.

* React Native Navigation version: 2.0.2305
* React Native version: 0.54.4
* Platform(s) (iOS, Android, or both?): Both
* Device info (Simulator/Device? OS version? Debug/Release?): iOS 10, 11 - Android API 25

PR #3325, opened 2018-06-08
[V2] Screen background option fixed in iOS (#3319)

emiyake:
This fixes issue #3319


yogevbd:
Merged: #3332 
Thanks!

Issue #3317, opened 2018-06-07
[V2] [Android] cannot find symbol import com.facebook.react.uimanager.MeasureSpecProvider


Peeeep:
### Issue Description

Since 4c8d48b1b596c314ff3242a2b619c3f893d4b5e1 Android builds fail on ReactNative >= 0.55.4 as both MeasureSpecProvider and SizeMonitoringFrameLayout moved to `uimanager/common` https://github.com/facebook/react-native/commit/6b45fb2cb1ca44fa7375bc7696bf90a68a85df3c

### Steps to Reproduce / Code Snippets / Screenshots

`react-native run-android`

```
error: cannot find symbol import com.facebook.react.uimanager.MeasureSpecProvider
error: cannot find symbol import com.facebook.react.uimanager.SizeMonitoringFrameLayout
```

### Temp patch 

`lib/android/app/src/main/java/com/reactnativenavigation/react/SyncUiImplementation.java`
```
- import com.facebook.react.uimanager.MeasureSpecProvider;
+ import com.facebook.react.uimanager.common.MeasureSpecProvider;

- import com.facebook.react.uimanager.SizeMonitoringFrameLayout;
+ import com.facebook.react.uimanager.common.SizeMonitoringFrameLayout;
```

No idea how to properly fix this, my Java skills are pretty much non-existent ;-)

---
### Environment
* React Native Navigation version: 2.0.2329
* React Native version: >= 0.55.4
* Platform(s) (iOS, Android, or both?): Android

nathantn:
I got the same problem with React Native Navigation 2.0.2330 version

croshim:
Same problem with 2.0.2333 version.

gabrielvcbessa:
Still getting this when trying to build a Release version, using 
```
./gradlew assembleRelease
```
I am using RNN 2.0.2377 with react-native 0.55.4

sytolk:
Add variantFilter fix this issue for me: https://github.com/wix/react-native-navigation/issues/3388#issuecomment-399071604

The problem is that with `./gradlew assembleRelease` all RNN lib flavours (reactNative51, reactNative55) are build no matter that we have `missingDimensionStrategy` defined.

gabrielvcbessa:
@sytolk Thanks, this works!

masterit3000:
@sytolk  thanks! wonderful

PR #3338, opened 2018-06-11
Fix Android compilation errors.

Mayzie:
As it is, trying to compile the current package will produce the
following compilation errors:

```
error: cannot find symbol class MeasureSpecProvider
error: cannot find symbol class SizeMonitoringFrameLayout
```

This is due to these two classes not existing in the *.uimanager
package, but rather the *.uimanager.common package.

See the directory structure at: https://github.com/facebook/react-native/tree/master/ReactAndroid/src/main/java/com/facebook/react/uimanager/common

This commit corrects the above mistake introduced in commit 4c8d48b


guyca:
This will be fixed by #3317 in a backwards compatible manner, hopefully this week.

PR #3339, opened 2018-06-11
Expose public method to find VC

birkir:
Maybe there is a better way of doing this?

I need an public method to retrieve a VC by componentId.

The use case I have, is to be able to find current VC's navigationController and push external component **as-is** into it. Using the external component feature does not work in this case, and finding current presented VC with `[UIApplication sharedApplication]` is not working.

For example:
```objc
RCT_EXPORT_METHOD(openURL:(NSString *)url componentId:(NSString *)componentId) {

  // Get UIViewController from ReactNativeNavigation by componentId
  UIViewController *vc = [ReactNativeNavigation findViewController:componentId];

  SFSafariViewController *safariVC = [[SFSafariViewController alloc] initWithURL:[[NSURL alloc] initWithString:url]];
  [vc.navigationController presentViewController:safariViewController animated:YES];
}
```


yershalom:
@birkir Please don't forget to change this line also
https://github.com/wix/react-native-navigation/blob/v2/lib/src/Navigation.ts#L18

birkir:
Nice catch, thanks!

PR #3343, opened 2018-06-11
Migrate to jest

yershalom:
Move all tests from mocha to jest for supporting multiple iOS simulators for e2e tests.


e2e/init.js
```diff
@@ -1,16 +1,25 @@
 const detox = require('detox');
 const config = require('../package.json').detox;
 const exec = require('shell-utils').exec;
+jest.setTimeout(480000);
```
guyca: Is such a long timeout really needed?

package.json
```diff
@@ -116,9 +117,8 @@
     }
   },
   "detox": {
-    "test-runner": "mocha",
+    "test-runner": "jest",
     "specs": "e2e",
-    "runner-config": "e2e/mocha.opts",
```
guyca: Delete `e2e/mocha.opts`?

guyca: Can we remove mocha from devDependencies now?

e2e/init.js
yershalom: We can lower this, nothing smart behind this number..

package.json
yershalom: I'll remove it.

yershalom: And with jest we use e2e/config.json

PR #3390, opened 2018-06-19
[v2][ios] add onSearchBarCancelPressed Event

benhiller:
Summary: This adds a new event `onSearchBarCancelPressed` that fires when the user presses the cancel button on the search bar in the navigation bar. `onSearchBarUpdated` does indirectly fire when the user presses cancel, as the query will be reset to an empty string, but it is not possible to distinguish those events from the user simply resetting the query. My use case for this is to have a view that appears on top of the normal content of the view controller when the search bar is focused.

This is my first pull request, so let me know if I'm doing anything wrong! I tried to adhere to the style in files as I found it.

Test Plan: ran `npm run test-js` and confirmed added JS unit tests passed.

I also tested this with my app, and confirmed the `onSearchBarCancelPressed` method was called when I pressed the cancel
button.

### DIFF
--- a/docs/docs/events.md
+++ b/docs/docs/events.md
@@ -103,3 +103,13 @@ class MyComponent extends Component {
   }
 }
 ```
+
+## onSearchBarCancelPressed (iOS 11+ only)
+Called when the cancel button on the SearchBar from NavigationBar gets pressed.
+```js
+class MyComponent extends Component {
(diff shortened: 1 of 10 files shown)


birkir:
Great! I was using the keyboard event handlers to detect it previously, it was very fragile.

Issue #3250, opened 2018-05-23
Iphone X tabBar flickers


lovebehl:
### Issue Description

While implementing FlatList i found a unexpected bug which occurs only in iphone X, i guess because of padding issue of tabbar it is happening.
Whenever a screen is pushed on tabbar and in that tabbar screen if you used FlatList, while coming back to the tabbar screen it starts flickering.

### Steps to Reproduce / Code Snippets / Screenshots

[Video showing flickering](https://expirebox.com/download/45b161bf441ccc5c371fa4f0658baeb3.html)

---
### Environment
* React Native Navigation version: 1.1.458
* React Native version: 0.55.2
* Platform(s) (iOS, Android, or both?): iOS only on iphone X
* Device info (Simulator/Device? OS version? Debug/Release?): All

Agontuk:
Facing the same problem, though I only tested it with RN 55. I only used the simulator since I don't have the device. Does this also occur in previous RN versions ?

rf1804:
@Agontuk i haven't checked with previous versions as i noticed this bug only now while testing on iphone X

omar-dev:
I am having the same problem with **V2**, however I suspect that it has to do the bottom tabs as I am pushing a new screen with tabs hidden when I go back it flickers the ScrollView

Agontuk:
@omar-dev, my problem exactly. But I'm using V1. Another question, does safearea work with `tabBarHidden: true` in V2 ?

rf1804:
@omar-dev i haven't hidden the tabs, it is normal tabs only and i'm using V1

a6051529:
Facing the same problem

a6051529:
Update!
---------------------------------------
I found A solution for this problem.

Use `{height: '100%'}` replace `{flex: 1}` at root React Tag

**example**:

![image](https://user-images.githubusercontent.com/7309118/40599169-6cc9278c-627e-11e8-962a-a88063451e06.png)

But I don't know the reason, Who can tell me? :heart::heart::heart:

danieljvdm:
has anyone gotten anywhere with this? It looks really really bad on the iPhone x

steverob:
Facing the same issue. :(

nikechanhk:
I have a "dirty" way to solve this.

You can use a View to wrap the Scrollview or Flatlist and hardcode the height of the view to 641
`<View style={{height: 641}}>`
            `   <FlatList style={{flex: 1}}`
           `    />`
[...]

Issue #2849, opened 2018-03-11
Height error tabBar when we pop


Tekosawa:
### Issue Description

When we launch app with startTabBasedApp, and I push new screen without tabBar, and when we go back to a screen with TabBar, the height of the TabBar is wrong

Here, if you want see the result : 
https://youtu.be/70fwvdQ-2F4

### Steps to Reproduce / Code Snippets / Screenshots

Start application with startTabBasedApp, push a new screen without tabBar 

```
this.props.navigator.push({
    screen: 'example.newScreen',
    passProps: {},
    animated: true,
    navigatorStyle: {
    	navBarHidden: true,
    	tabBarHidden: true,
    	navBarTransparent: true,
    	navBarTranslucent: Platform.OS === 'ios',
    },
    animationType: 'none',
});

```
And pop on previous screen with TabBar.

### Environment
* React Native Navigation version: 1.1.386
* React Native version:  0.51
* Platform(s) (iOS, Android, or both?): iOS only
* Device info (Simulator/Device? OS version? Debug/Release?): Device, Release (Sometime on Simulator, but hard to reproduce) - iOS 11.2

Tekosawa:
When we pop too fast just after the push, controller recreate area above of tabbar on simulator.

Create startTabBasedApp and on screen pushed, write this code : 

https://gist.github.com/Tekosawa/4d4127a548518e84377267aaa3c56869

Push the screen with parameter in my first post and after 5-10 retry (2-3 second)) you will enable to recreate the bug.

margox:
+1

manuhook:
+1

felipemartim:
Same thing here. This has been happening for a long time now.

harshalbid271:
Anyone found a fix for this?

steverob:
Facing this issue on iPhone X after upgrade to `55.4` from `52.3` :(

simonmitchell:
Done a bit of digging into this, it seems like it's the `RCTRootContentView` that has the incorrect height. `RCTRootView` adds it as a subview and set's it's frame in `layoutSubviews` like so:
```
- (void)layoutSubviews
{
  [super layoutSubviews];
  _contentView.frame = self.bounds;
  _loadingView.center = (CGPoint){
    CGRectGetMidX(self.bounds),
    CGRectGetMidY(self.bounds)
  };
}
```

[...]

PR #3405, opened 2018-06-21
Fix/rootview bounds iphone x

simonmitchell:
Replacing `self.view` on a `UIViewController` should not be done in this way, if at all... It can lead to unexpected behaviour, which has been shown in the recent issues submitted where UI glitches were seen on an iPhone X:

https://github.com/wix/react-native-navigation/issues/2849
https://github.com/wix/react-native-navigation/issues/3250

which this pull request fixes!


steverob:
@yogevbd  Can this be reviewed please.. Thanks! :)

steverob:
I updated our project to use this version and our custom NavBar just went away.

I pulled the example project and I only see the Navigator Buttons (right and left). From the code I can see that there is a button in the center of the navbar but it's nowhere to be seen in the app.

Can you confirm this issue @simonmitchell @yogevbd 

8a894b0f-48fa-4b93-b0e3-cf440b9b12de

simonmitchell:
@yogevbd @guyca I have tested this using the latest code on this branch and it seems to work fine in the example project. I think the problem was that I created the merge request prematurely and then shortly after fixed some issues I'd caused, but the PR had already been merged by this point (Impressively quick work from you guys!)

Issue #3392, opened 2018-06-20
[v2][Android] NPE


sytolk:
This can be reproduced if the app starts with the red screen like this:
`error: bundling failed: Error: We don't know what to do with this node type. We were previously a Statement but we can't fit in here?
    at NodePath.insertBefore (/home/stanimir/structmedia/fuelapp/FuelApp/node_modules/@babel/traverse/lib/path/modification.js:55:11)`

`"description":"/home/stanimir/App/node_modules/react-native-modal/src/index.js: We don't know what to do with this node type. We were previously a Statement but we can't fit in here?`
                                                                       
with dependencies: {
    "react": "16.3.2",
    "react-native": "0.56.0-rc.2",
    "react-native-navigation": "^2.0.2359",
    "react-native-modal": "^6.1.0",
}
"devDependencies": {
    "babel-preset-react-native": "^5.0.2",  **-> this red screen not appears with "babel-preset-react-native": "5.0.0"**
}
```
06-20 09:08:33.444 25288-25288/? E/AndroidRuntime: FATAL EXCEPTION: main
                                                   Process: com.fuelapp.debug, PID: 25288
                                                   java.lang.RuntimeException: Unable to resume activity {com.fuelapp.debug/com.fuelapp.MainActivity}: java.lang.NullPointerException: Attempt to invoke virtual method 'com.facebook.react.bridge.JavaScriptModule com.facebook.react.bridge.ReactContext.getJSModule(java.lang.Class)' on a null object reference
                                                       at android.app.ActivityThread.performResumeActivity(ActivityThread.java:3659)
                                                       at android.app.ActivityThread.handleResumeActivity(ActivityThread.java:3699)
                                                       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1655)
                                                       at android.os.Handler.dispatchMessage(Handler.java:105)
                                                       at android.os.Looper.loop(Looper.java:172)
                                                       at android.app.ActivityThread.main(ActivityThread.java:6637)
                                                       at java.lang.reflect.Method.invoke(Native Method)
                                                       at com.android.internal.os.Zygote$MethodAndArgsCaller.run(Zygote.java:240)
                                                       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:767)
[...]

PR #3475, opened 2018-07-02
Fix https://github.com/wix/react-native-navigation/issues/3392

sytolk:


guyca:
Hey @sytolk 
Thanks for taking a stab at this. I don't think only a null check is the correct way to handle this.
We should probably check if `waitingForAppLaunchEvent == true` in `onReactContextInitialized` and emit the `appLaunched` event in that case.
I'm closing and will tackle this soon. Thanks for taking the time to look into it 👍

PR #3497, opened 2018-07-04
[BREAKING] Event registry rework

DanielZlotin:
- Events will NOT be automatically called on components. Instead, all events must be registered (and later `.remove()`) through `Navigation.events()...`
- ~Combined all component lifecycle events into `registerComponentLifecycleListener`~
- ~TODO: add componentDidMount, componentWillUnmount events~
- For ease of use you can call `this.subscription = Navigation.events().bindComponent(this)` from inside your component's constructor and all methods will be automatically triggered just for this component
- ~Don't forget to `this.subscription.remove()` on componentWillUnmount~ this is optional now. RNN cleans the listeners whenever the component is unmounted (you can still call `remove` yourself when needed)


DanielZlotin:
All of this is sadly due to how HOC work in react.. There's no way (that I found) to call overridden methods on children of react components..

DanielZlotin:
All is done. Once the e2e are stable we can merge

juddey:
This is fantastic, thank you :clap:

PR #3503, opened 2018-07-05
[v2] LargeTitle: Fix scroll position after `mergeOptions` is called

ptmt:
Before this fix each `mergeOptions` call messes with a scroll position which leads to smaller titles (you can scroll LargeTitle back but it's not convenient). 

```js
class App extends Component<Props> {
  static get options() {
    return {
      topBar: {
        title: {
          text: 'My Screen'
        },
        largeTitle: {
          visible: true
        },
        searchBar: true,
        translucent: true,
      },
    };
  }
  
  componentDidMount() {
    setTimeout(() => {
        topBar: {
          title: {
            text: "My Screen"
          }
        }
      })
    })
  }
  render() {
    return (
          <FlatList
            data={ITEMS}
            contentContainerStyle={styles.container}
            renderItem={({ item }) => (
              <View style={styles.row}>
                <Text style={styles.rowText}>
                  <Text>{item.key}</Text>
                </Text>
              </View>
            )}
          />
    );
  }
}
```
screen shot 2018-07-05 at 16 51 34

after fix:

screen shot 2018-07-05 at 16 54 46

### DIFF
--- a/lib/ios/RNNLargeTitleOptions.m
+++ b/lib/ios/RNNLargeTitleOptions.m
@@ -8,13 +8,13 @@
 
 	if (@available(iOS 11.0, *)) {
 		
-		viewController.navigationItem.largeTitleDisplayMode = UINavigationItemLargeTitleDisplayModeNever;
-	
+		
 		if ([self.visible boolValue]){
 			viewController.navigationController.navigationBar.prefersLargeTitles = YES;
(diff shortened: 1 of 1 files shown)


ptmt:
Hmm, actually no, it doesn't work under different conditions, so the reason must be different. Closing for now.

Issue #1977, opened 2017-10-05
Need RTL Support for Tab Navigation


coder4affine:
### Issue Description

As per my believe there is no support for RTL for tab navigation. 

### Steps to Reproduce / Code Snippets / Screenshots

I am using I18N for switch the language b/w arabic and english. but I did not find any way for RTL support for navbar and tab navigation.

---
### Environment
* React Native Navigation version: 1.1.235
* React Native version: 0.48
* Platform(s) (iOS, Android, or both?): both
* Device info (Simulator/Device? OS version? Debug/Release?): all

jadidian:
Any updates on this? I really like `react-native-navigation` because of its smooth transitions / animations but I had to use `react-navigation` with all its glitchy transitions in my latest project only because it supports RTL; I think this is an important feature for lots of people speaking Hebrew, Arabic, Persian, Azeri, Urdu etc. and deserves more attention from the development team.

blackhair:
Need to RTL `tabNavigator` items, please fix it

RamanDEEP225:
is there any way to support RTL by using react-native-navigation npm in IOS, I did this in android but not in IOS, is it possible?

PR #3508, opened 2018-07-06
RTL Layout

### LINKED ISSUE #3485: [V2] RTL Layout Support
### Issue Description

In many cases our applications need to support __RTL__ languages, __React__ allows to _RTL_ via `I18nManager` API, but __RNN__ does not support any options to deal with layout directions ( Only loves _LTR_ 😢 ).

~~Android Platform:~~
~~In RNN v1 when system language was a RTL  language (like: Persian or Arabic) or simply call `I18nManager.forceRTL(true)`, Everything goes fine, back button shown in right.~~
~~But in RNN v2 Nothing, always left and always LTR, no way to done it.~~

1. Change system language to > `Persian/فارسی` `Arabic`
2. Restart _RNN_ App
3. Push a component to navigator  and __look at navigation__

  > _If works, back button should displayed at right either title._

#### UPDATE: I described workarounds for both _Android_ and _iOS_ platforms in comments.
---
### Environment
* React Native Navigation version: 2.0.2394
* React Native version: 0.55.4
* Platform(s) (iOS, Android, or both?): Both
* Device info (Simulator/Device? OS version? Debug/Release?): Simulator

hadimostafapour:
Fixes #3485 

```javascript

// Set direction to RTL layout
      layout: {
          direction: 'rtl'
      }
});

Navigation.setRoot({});
```

```xml
<application
      android:name=".MainApplication"
+     android:supportsRtl="true"
      ...
      android:theme="@style/AppTheme">
```

### DIFF
--- a/lib/ios/RNNCommandsHandler.m
+++ b/lib/ios/RNNCommandsHandler.m
@@ -7,6 +7,7 @@
 #import "RNNSplitViewController.h"
 #import "RNNElementFinder.h"
 #import "React/RCTUIManager.h"
+#import "React/RCTI18nUtil.h"
 
 
 static NSString* const setRoot	= @"setRoot";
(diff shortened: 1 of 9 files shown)


hadimostafapour:
And also fixes #1977

lib/ios/RNNCommandsHandler.m
```diff
@@ -48,6 +49,20 @@ -(instancetype) initWithStore:(RNNStore*)store controllerFactory:(RNNControllerF
 
 -(void) setRoot:(NSDictionary*)layout completion:(RNNTransitionCompletionBlock)completion {
 	[self assertReady];
+
+	if (@available(iOS 9, *)) {
+        if ([_controllerFactory.defaultOptionsDict[@"layout"][@"direction"] isEqualToString:@"rtl"]) {
+            [[RCTI18nUtil sharedInstance] allowRTL:YES];
+            [[RCTI18nUtil sharedInstance] forceRTL:YES];
```
omar-dev: Forcing RTL with React native will break Apps that implement RTL first approach, I think the best solution is to leave it to the user to force RTL or not in js.

hadimostafapour: Right, but it's required to correctly apply RTL layout + RTL content.
If We want a RTL layout means we needs RTL content or LTR layout needs LTR content, 
Could you describe some cases that needs RTL layout with LTR content?

hadimostafapour: @omar-dev In new version of RN marginRight is always right, they are introduced  marginStart, marginEnd instead of making marginRight reverse on RTL.

henrikra:
@hadimostafapour Are you going to fix those conflicts?

RubenSandwich:
What is the status on this pull request? I'm depending on RTL support for an upcoming app.

hadimostafapour:
I've created a new pull request for this, check out #4575

PR #3506, opened 2018-07-05
[v2] making topbar icon configuration compatible on Android and iOS

emiyake:
Top bar's icon color configuration were different between iOS and Android.

On iOS
`{topBar: { backButton: {color: 'red' } } }` wasn't working.
`{topBar: { leftButtons: [{color: 'red'}] }` wasn't working either.

On the other hand: 
`{ topBar: { buttonColor: 'red'} }` was ignored on Android.

It works on both platforms now.

PR #3509, opened 2018-07-06
[Docs] Update styling

hadimostafapour:
`largeTitle: true` is not valid and throws error


hadimostafapour:
@guyca Seems `topBar.backButton` is android specific but not mentioned in docs. Am I right?

emiyake:
@hadimostafapour I create a PR making Topbar configuration common to Android and iOS
https://github.com/wix/react-native-navigation/pull/3506

PR #3519, opened 2018-07-09
[v2] Support UIBarStyle for topBar and bottomTabs

birkir:
Add support for `barStyle` property in TopBar and BottomTabs.


birkir:
Wow the tests are super flaky. Is this detox or just the nature of the emulators and would apply to all graybox/blackbox testing?

PR #3547, opened 2018-07-13
Feature/avoid double routing

DavitVosk:
Avoid double routing


tayfunyasar:
Will you guys merge this commit?

birkir:
**Quick review**

1. Lots of noise in your PR, like custom border color etc.
2. This doesn't work. You are blocking the same screen component to be pushed twice in a row, which is expected in RNN. (Think of a list of items, each item shows title and group, if you press the group you push the same controller with another list of items, etc. etc.)

PR #3441, opened 2018-06-26
Revert "Fix/rootview bounds iphone x"

yogevbd:
Reverts wix/react-native-navigation#3405

PR #3548, opened 2018-07-13
Fix app crash bug when reload [cmd + r]

NaoyaKurahashi:
Sorry, this pull request I created yesterday and merged same day is not sufficient...
https://github.com/wix/react-native-navigation/pull/3441

```
if (self.view != nil){
-        RCTRootView *rootView = self.view;
-        NSString *screenName = [rootView moduleName];
```
This condition is always true. So, when we reload simulator (cmd + r) app reset and back to home of simulator.

I modify my commit, so please check.


developer239:
❤️

ziyafenn:
Looking forward to this merge

PR #3578, opened 2018-07-18
[V2] Allow swipeable components to be used on iOS when drawers are used

### LINKED ISSUE #3575: [v2] cannot use a swipeable element on iOS when using side drawer
### Issue Description

I am using a swipable list view (from native-base) in a tab view layout that has a **left** side drawer. I want to swipe left from the **right**, but it does not work on iOS. it works on android. on iOS, I can drag it only for about 10 pixels.

screen shot 2018-07-17 at 4 05 08 pm
screen shot 2018-07-17 at 4 06 56 pm

[FILL THIS OUT - It will be extremely helpful]

---
### Environment
* React Native Navigation version: 2.0.2.412
* Device info (Simulator/Device? OS version? Debug/Release?): both

ujwal-setlur:
Hi,

This resolves #3575, where I could not use a swipeable component like Swipeable ListView on iOS when I was using a side drawer. The openDrawer gesture was taking over from the swiping action. This change disables the drawer swipe open gesture from the center view. I think this is acceptable since drawers can still be dragged open from the edges, and this seems to be standard behavior on iOS anyway.

If there is a better way to do this, please let me know. Thanks!


ujwal-setlur:
Can someone review this? Thanks.

ujwal-setlur:
Closing to reopen a fresh one.

PR #3604, opened 2018-07-22
[v2] Add badgeColor option to bottomTab

SCadilhac:
This commit adds the badgeColor option, to customize the color of the badge.


mths0x5f:
Look good, have you tried to customize the elevation of BottomTabs also? (Android-only)

lib/android/app/src/main/java/com/reactnativenavigation/views/BottomTabs.java
```diff
@@ -29,6 +30,11 @@ public void setBadge(int bottomTabIndex, String badge) {
         setNotification(badge, bottomTabIndex);
     }
 
+    public void setBadgeColor(@ColorInt Integer color) {
+        if (color == null) return;
```
guyca: You should also check if the current badge color equals `color` and return if true. Otherwise the BottomTabs view will be recreated unnecessarily which results in noticeable artifacts.

guyca: @SCadilhac You're right, merging :)

PR #3610, opened 2018-07-23
[v2][ios] Do not set largeTitleDisplayMode to prevent largeTitle jumping

ptmt:
It's probably a temporary workaround, until we figure out why it gets re-rendered incorrectly. 
Use case: Two tabs with scrollviews and large titles inside. After switching tabs, a large title gets rendered as a small title. 
Seems like a minor change with the much better and consistent UX as an outcome, but maybe I'm missing something.

### DIFF
--- a/lib/ios/RNNLargeTitleOptions.m
+++ b/lib/ios/RNNLargeTitleOptions.m
@@ -8,12 +8,20 @@
 
 	if (@available(iOS 11.0, *)) {
 		
+		BOOL prefersLargeTitles = viewController.navigationController.navigationBar.prefersLargeTitles;
+		BOOL previousValue = prefersLargeTitles;
+		long displayMode = viewController.navigationItem.largeTitleDisplayMode;
 		if ([self.visible boolValue]){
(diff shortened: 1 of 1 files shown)


ptmt:
> his works for me but only if I've already scroll the view once and make it large

So something wrong happens when some properties get touched more times than needed while options applying. I've updated the PR, could you please try this without `mergeOptions` to see if it helps?

nicolaslopezj:
It does the same that with the other method.

But, without any of those changes, the ```mergeOptions``` "co-solution" doesn't work

nicolaslopezj:
I liked more the first solution. It’s more simple and it will be merged easier

ptmt:
It seems all of these are semi-solutions, working half of the times. To my use case, `mergeOptions` doesn't help either. (BottomTabs + stacks). Will try to investigate more on this + will share a reproducible example.

ericlewis:
Still an issue

ericlewis:
Using the mergeOptions hack helps if you have scrolled the item down once, but the first tab in a tab bar doesn't work with this hack.

PR #3620, opened 2018-07-24
Refactor peek and pop

birkir:
A couple of reasons for this refactor:

1. The event system was changed in RNNv2 so the preview events were commented out, hence not working.
2. `<Navigation.Element />` could not be nested, so that solution didn't work for many use cases (even though it was a cool idea).
3. Handling touches for determining if we should do a screen push or a peek is a very hard problem that I feel like RNN should ship a solution with.
4. Integrating the system into external View Controllers wasn't possible.

### TODO

- [x] Documentation
- [x] previewCompleted event
- [x] Component to handle touches and prevent bad things to happen
- [x] moved the previewContext logic into a callback that is attached to the VC

### Breaking changes
 - `preview.elementId` is now `preview.reactTag`
 - `<Navigation.Element />` is now `<Navigation.TouchablePreview />` for handling onPress and onPressIn.
 - `previewContext` event is depricated and I merged `previewCommit + previewDismissed` into one event: `previewCompleted({ previewComponentId })`

### Complete demo

This demo shows all the features this PR implements. Disables scroll on peek etc.

```jsx
class MyComponent extends Component {

  constructor(props) {
    super(props);
    this.state = { scrollEnabled: true };
    Navigation.events().bindComponent(this);
  }

  previewCompleted = ({ componentId, previewComponentId }) => {
    console.log('Component from %o to %o', componentId, previewComponentID);
  }

  onPress = ({ reactTag }) => {
    Navigation.push(this.props.componentId, {
      component: {
        name: 'second.screen',
        options: {
          preview: {
            reactTag,
            height: 300,
            commit: true,
            actions: [/* ... */],
          },
        },
      },
    });
  }

  render() {
    return (
      <ScrollView
        scrollEnabled={this.state.scrollEnabled}
      >

        <Navigation.TouchablePreview
          touchableComponent={TouchableHighlight}
          onPress={this.onPress}
          onPressIn={this.onPress}
          onPeekIn={() => this.setState({ scrollEnabled: false })}
          onPeekOut={() => this.setState({ scrollEnabled: true })}
        >
          <Text>My button</Text>
        </Navigation.TouchablePreview>

      </ScrollView>
    );
  }

}
```

### DIFF
--- a/docs/docs/animations.md
+++ b/docs/docs/animations.md
@@ -33,3 +33,35 @@ Navigation.push(this.props.componentId, {
   }
 });
 ```
+
+## Peek and Pop (iOS 11.4+)
+
+react-native-navigation supports the [Peek and pop](
+https://developer.apple.com/library/content/documentation/UserExperience/Conceptual/Adopting3DTouchOniPhone/#//apple_ref/doc/uid/TP40016543-CH1-SW3) feature in iOS 11.4 and newer.
+
+This works by passing a ref a componentent you would want to transform into a peek view. We have included a handly component to handle all the touches and ref for you.
+
+```jsx
+const handlePress ({ reactTag }) => {
+  Navigation.push(this.props.componentId, {
+    component {
+      name: 'previewed.screen',
+      options: {
+        preview: {
+          reactTag,
+        },
+      },
+    },
+  });
+};
+
+const Button = (
+  <Navigation.TouchablePreview
+    touchableComponent={TouchableHighlight}
+    onPress={handlePress}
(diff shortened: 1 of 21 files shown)


yogevbd:
Hey @birkir, for some reason tests started the fail so I had to comment out `TouchableOpacity` in `Navigation.ts`.
Can you please check? Thanks

birkir:
I'm on it.

stoica94:
This example is not working. Tried this exact example and got 
`undefined is not an object (evaluating '_ref.reactTag')`

PR #3632, opened 2018-07-26
[v2]Shared element transition android

guyca:


daedaesplayground1911:
Yes I like this combo that you have come up with. I'm going to have to practice on this stuff but I'm going to try to make it work. But thanks bro.

millersdev:
Hi, we would like to pick this up, for now its a bit buggy on iOS but not at all working for Android, any ideas or status why its not working and what's needed to make it work?

Issue #3627, opened 2018-07-25
[V2] showModal with {topBar: {drawBehind: true}} not displaying the TopBar on Android


emiyake:
Top bar is not shown in `showModal` when adding the following option:
```
  { 
     topBar: { 
        drawBehind: true,
        elevation: 0,
     }
  }
```

It works fine when `elevation: 0` is removed.

This might have been introduced after this merge: https://github.com/wix/react-native-navigation/pull/3601/files
@guyca 

---
### Environment
* React Native Navigation version: 2.0.2428
* React Native version: 0.56.0
* Platform(s) (iOS, Android, or both?): Android only
* Device info (Simulator/Device? OS version? Debug/Release?): Api 26

edmundojr:
Same happening here, when setting `elevation` to `0` and `drawBehind` to `true` the topBar is hidden.

PR #3633, opened 2018-07-26
[V2] fixes when elevation: 0 and drawBehind: true on Android

emiyake:
Fixes issue https://github.com/wix/react-native-navigation/issues/3627


mths0x5f:
have you understood why it happens?

PR #3645, opened 2018-07-28
Instructions for create-react-native-app

bbalan:
Apps created with create-react-native-app will have a single index.js entry point. References to index.ios.js will throw an error.

### DIFF
--- a/docs/docs/Installing.md
+++ b/docs/docs/Installing.md
@@ -42,6 +42,8 @@
 
 	@end
 	```
+	
+For apps created with `create-react-native-app`, change `jsBundleURLForBundleRoot:@"index.ios"` to `jsBundleURLForBundleRoot:@"index"`.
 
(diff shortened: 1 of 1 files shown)


henrikra:
This advice is not specific for `create-react-native-app`. `index.android` and `index.ios` was combined to one file in some RN version

bbalan:
You're right, looks like it was [v0.49](https://github.com/facebook/react-native/releases/tag/v0.49.0).

birkir:
Same applies for android:

In MainApplication.java
```java
    @Override
    protected ReactGateway createReactGateway() {
        ReactNativeHost host = new NavigationReactNativeHost(this, isDebug(), createAdditionalReactPackages()) {
            @Override
                return "index";
            }
        };
        return new ReactGateway(this, isDebug(), host);
    }
```

Issue #1642, opened 2017-08-04
Plan for redux support in v2? (HOC support)


gtchance:
I'm looking to switch from v1 to v2, but the first thing I noticed was that when registering containers, you can no longer pass in the store and Provider. Is this in the roadmap?

DanielZlotin:
I don't think it's necessary to add any specific support for redux or any other data flow library, all you need to do is wrap your screens in a higher order component `<Provider store={store}/>`.

I automatically pass all props through to the inner implementation of `Container` so it should just work.

Open an issue if you see a specific problem with redux

em222iv:
@DanielZlotin Would you please post an example of how to accomplish this with v2? I'm having a hard time implementing redux with the new v2 structure of registerContainer

DanielZlotin:
You can look at the integration test with redux here
https://github.com/wix/react-native-navigation/tree/v2/integration/redux it
provides a example.
If you encounter a specific problem please report

jleed:
Is there example how to register containers with redux?

I tried to do this

`
			Navigation.registerContainer(item.name, () => {
				const Screen = item.screen;
				return (props) => {
					return (
						<Provider store={store}>
							<Screen {...props}/>
						</Provider>
					)
				}
			});
`

but it fails

Cannot read property 'didDisappear' of undefined
Cannot read property 'didAppear' of undefined

vovkasm:
This should work:
```
  Navigation.registerContainer('FitHelper.ProgramList', () => {
    return class ProgramList extends React.Component {
      render () {
        return <Provider store={store}><ProgramListScene /></Provider>
      }
    }
  })
```
Under the hood, Navigation's component registry wraps that you return from creator to another component that gets ref from original component, so you need full subclass of React.Component here or refs will not work.

pentarex:
@vovkasm it works, but is there a nicer, cleaner solution for this. I have like 11 containers. And passing it for every looks like overkill... or I am not getting it

vovkasm:
You always can create HOC... something like this (not tested):
```jsx
function sceneCreator (sceneComp, store) {
  return () => {
    class SceneWrapper extends React.Component {
      render () {
        return <Provider store={store}>{React.createElement(sceneComp)}</Provider>
// may be this also works: return <Provider store={store}><sceneComp/></Provider>
      }
    }
  }
}
```
And use as:
```javascript
Navigation.registerContainer('screen.Test', sceneCreator(TestScene, store))
```

[...]

PR #3675, opened 2018-08-02
Official redux support

DanielZlotin:
The people have spoken.

### DIFF
--- a/lib/src/Navigation.ts
+++ b/lib/src/Navigation.ts
@@ -57,6 +57,14 @@ export class Navigation {
   }
 
   /**
+   * Utility helper function like registerComponent,
+   * wraps the provided component with a react-redux Provider with the passed redux store
+   */
+  public registerComponentWithRedux(componentName: string, getComponentClassFunc: ComponentProvider, reduxStore: any): ComponentType<any> {
+    return this.componentRegistry.registerComponent(componentName, getComponentClassFunc, reduxStore);
+  }
+
+  /**
    * Reset the app to a new layout
    */
   public setRoot(layout): Promise<any> {
--- a/lib/src/Navigation.ts
+++ b/lib/src/Navigation.ts
@@ -60,8 +60,8 @@ export class Navigation {
    * Utility helper function like registerComponent,
    * wraps the provided component with a react-redux Provider with the passed redux store
    */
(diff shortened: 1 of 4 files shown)


rfdc:
I have tried few option from #1642 and did not work. 
What is the solution? Is there an example showing the whole process? 
Thanks

tiempham:
@DanielZlotin  Can you show an example? I don't understand function parameters. Thanks

drorbiran:
Hi, Here is an example of a super simple project with Redux
https://github.com/drorbiran/react-native-simple-blog/blob/redux/src/screens.js

DanielZlotin:
Updated with example

spacesuitdiver:
@DanielZlotin sorry if this is contained in some other discussion. Was there a technical reason we chose to explicitly provide a method to register components with redux instead of just having everyone figure this out with HOCs (or some other solution)?

mammad2c:
componentDidAppear not working in v2.0.2472
Is there any other way to connect redux ?

DanielZlotin:
Nothing wrong with doing your own HOCs. Just that lots of people had problems implementing it, so we added this as a "quality of life" feature.

spencewine:
@DanielZlotin I'm getting this error: `Cannot read property 'componentId` of null when using `registerComponentWithRedux`. Any idea why this would happen?
screen shot 2018-08-17 at 11 43 06 am

ywongweb:
@spencewine Did you managed to find a solution?

wbercx:
@DanielZlotin Just wanted to provide some feedback on this. I recently overheard someone suggest the latest version of RNN only supports Redux.

The implementation isn't limited to Redux and, for example, works fine with `mobx-react` as well. The current naming does not suggest that though.

I think it may make more sense to have this be called `registerComponentWithProvider`. It seems that captures what this essentially does - it let's you use *a* Provider.

DanielZlotin:
Ah I knew this would happen :facepalm: that's why I was reluctant to add this in the first place.
Thanks for the feedback.

wbercx:
Nah, it's a fantastic addition, it's just the naming that's is a bit more specific than it needs to be :grin:

DanielZlotin:
The store argument is kinda redux specific though.. Please open a separate issue with a better api suggestion and we'll deprecate the current api if needed.
Thanks

alzalabany:
Does this work with new context api providers ?

barjas:
any step by step plz ?

PR #3678, opened 2018-08-02
V2

victorpettengill:
Fix backButton tint color for iOS


birkir:
And now we know what it looks like 😂

Issue #3243, opened 2018-05-22
[Android] App stuck when resuming after being in the background for a long time


barhbie:
### Issue Description

Android app gets stuck on the splash screen after putting it in the background with the home button and then resuming it after several hours.

### Steps to Reproduce / Code Snippets / Screenshots

The problem is most easily reproduced this way:

Create a new app with `react-native init app` and install react-native-navigation following the instructions at https://wix.github.io/react-native-navigation/#/installation-android

Implement a simple test app:
```
import React, { Component } from 'react';
import {
  Text,
  View
} from 'react-native';
import { Navigation } from 'react-native-navigation';

class TestScreen extends Component {
    render() {
        return (
            <View>
                <Text>
                    Welcome to React Native!
          </Text>
            </View>
        );
    }
}

Navigation.registerComponent('testScreen', () => TestScreen);

Navigation.startSingleScreenApp({
    screen: {
      screen: 'testScreen',
      title: 'Welcome',
    },
    animationType: 'slide-down'
  });
```

Start the app with `react-native run-android`. When the app has started shut it down and then start it again by pressing the launcher icon. Press the home button to put the app in the background and then wait several hours (it's difficult to make a good estimate unfortunately) and then resume the app, this will result in a permanent blank screen (or splash screen if such is implemented). Putting the app in the background and resuming it again solves the problem.

The "wait several hours" part makes it hard to debug, but I've also managed to recreate the error by starting the app and putting it in the background as described above and then going to `Settings -> Developer Options -> Running services -> Show cached processes -> native-navigation-app-name -> Stop` and then resuming the app.

It is worth noting that resuming the app in any other situation than the two described above seems to work fine.

The issue is RNN-specific. I've tried running a non-RNN app and this problem does not occur.

---
### Environment
* React Native Navigation version:  I've tested 1.1.348, 1.1.457 and 1.1.462, problem occurs on all these versions.
* React Native version: I've tested 0.51.0, 0.55.3 and 0.55.4, same problem on all versions
* Platform(s) (iOS, Android, or both?): Android
* Device info (Simulator/Device? OS version? Debug/Release?): The problems occur both on device (Samsung Galaxy S9+, Android version 8.0.0) and emulator (Android version 7.1.1), and both in debug and release versions of the app.

jslaunch:
I am also facing the same issue.
@barhbie You found any solutions?

barhbie:
Yes, I should've posted the solution and closed this issue but I forgot.

[...]

Issue #1022, opened 2017-04-10
[Android critical] App is restarted when activity is killed


henrikra:
### Issue Description
Background:
Before we were using NavigationExperimental which is JS based navigation. It is now deprecated and that is why we refactored navigation to your library.

Problem:
When `Do not keep activities` setting is ON from the Developer Settings the app restarts itself. This happens when home button is pressed to put the app on background. At this point the activity is destroyed since `Do not keep activities` is ON. When you go back to the app it starts it from start again and initial state is used.

Also looks like ALL the JS code is run again when activity is destroyed. Redux stores, all files. This should not happen in my opinion. Looks like if I remove [this line](https://github.com/wix/react-native-navigation/blob/43b1edcc1c48df564437586b51c3c72068651557/android/app/src/main/java/com/reactnativenavigation/controllers/NavigationActivity.java#L148) from this library it does not kill JS which is good. Only problem now is that the navigation state is destroyed and I have to always start from home screen.

When we were using NavigationExperimental this did not happen even when `Do not keep activities` was ON. It always remembered where the app was even when activity was destroyed.

So my guess is that react-native-navigation does not resume the state back up when activity is recreated. It always starts everything from scratch when activity is destroyed.

### Why is this important?
Because Android OS can destroy any activities any time it wants to save memory. We tried this library in production and several users experienced this. So in their case they have used some other memory intensive app which caused OS to kill our app activity.

### Steps to Reproduce
1. Go to `Android Settings` -> `Developer options` -> `Do not keep activities`
2. Build and run [react-native-navigation example](https://github.com/wix/react-native-navigation/tree/master/example)
3. Tap example `Push Plain Screen` to push new screen
4. Press home button to put example app to the background
5. Go back to example app and notice that the app resets itself and starts from the home screen

---
### Environment
* React Native Navigation version: v2.0.0-experimental.271
* React Native version: 0.42.3
* Platform: Android
* Device info: Debug, real device

Villux:
Yes, this is very critical problem. Should we somehow keep the state always stored in JS side? So when app's activity is redrawn we restore everything in JS side.

guyca:
Hey guys
This a a very interesting observation. Currently, Js execution is indeed coupled to the Activity which is why Js context is destroyed along with the Activity. We already changed that in V2. I think this change was introduced in RN when HeadlessJs was added, but I might be wrong.
I assume your redux state is already persisted, if that's the case this shouldn't be a major issue.

[...]

PR #3688, opened 2018-08-04
Hack to fix when app is stuck on the splash screen.

BasitAli:
When the navigation activity is killed/restarted by the system, onCreate launches the SplashActivity and kills itself. The new SplashActivity starts context initialization but that is broken in between by the onDestroy of the NavigationActivity we just killed. This fix adds a boolean to disable destruction of js in this case.

### DIFF
--- a/android/app/src/main/java/com/reactnativenavigation/controllers/NavigationActivity.java
+++ b/android/app/src/main/java/com/reactnativenavigation/controllers/NavigationActivity.java
@@ -63,12 +63,15 @@ public class NavigationActivity extends AppCompatActivity implements DefaultHard
     @Nullable
     private PermissionListener mPermissionListener;
 
+    boolean killedBySystem = false;
+
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         if (!NavigationApplication.instance.getReactGateway().hasStartedCreatingContext() ||
                 getIntent() == null ||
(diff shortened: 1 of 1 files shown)


binbinla:
hello, same issue here, can this change fix the problem for you?

BasitAli:
@binbinla It did. Using this branch on the app that I'm working on right now.

irvingchenn:
@BasitAli LGTM

badrange:
Hi @guyca I'd love to see your feedback on this PR. 

Upon committing #1838 you added a comment about a similar issue and how to prevent it, which you repeated the following day in https://github.com/wix/react-native-navigation/issues/1022#issuecomment-329432373

This doesn't work anymore - but @barhbie came up with a fix in https://github.com/wix/react-native-navigation/issues/3243#issuecomment-406208403

I have tested both methods, and they both work for our use case. @BasitAli 's solution is the simplest one, is there anything that prevents this from being merged?

skovhus:
This works for us. We avoid killing the navigation state every time the app opens from the background and we can now assume the app although the activity is killed. 👍

tigo0:
Would be awesome to have this merged soon. Thanks.

kevinkevw:
Can anyone review this one and perform a merge? 

The current Navigation.isAppLaunched() way is not very ideal and many users reported it blocks the navigator to proceed, all navigator functions not able to run. ( not sure the reasons yet but got reports from various Android OS )

asolas:
same as home button, a new screen created each time.
test

PR #3743, opened 2018-08-10
Add link to V2 Documentation

thomasMinshull:
Added a header to the sidebar to indicate that this is the V1.1.X documentation and added a link to the V2 docs.

I know of at least one other person who waisted a few hours because they didn't realize they were looking at the wrong docs, so I thought that this header and link might be of some use.


guyca:
Thanks!

PR #3758, opened 2018-08-13
Update Installing.md - adding code closing section

franciscofsales:
Adding missing code closing tag that causes it to look broken on website


franciscofsales:
Realized its duplicated

PR #3813, opened 2018-08-21
support badge 0

whhotw:
unifying the functionality with Android


henrikra:
Hmm what is the use case that we want to show badge with zero?

whhotw:
For now

```
this.props.navigator.setTabBadge({
  badge: 0,
});
```

screen shot 2018-08-21 at 22 54 31

Support is mean, when set badge 0 will be remove badge on iOS

whhotw:
@henrikra yes, if set `null` will be disappear, but i want each platform behavior same

henrikra:
Ahaa so the platforms behave in different way.

whhotw:
yap, on Android if set 0 can disappear, but different with iOS

henrikra:
Okay now I get it! Please update the description of this PR because it should tell that you are unifying the functionality with Android

henrikra:
I think they fixed it now https://github.com/wix/react-native-navigation/commit/62125f9a8f1dbfcb38fb50456117b5bcb8490d7a

whhotw:
i don't think so, because 62125f9 just resolve empty string, but i'm want set number 0

guyca:
@whhotw This seems like Js issues. We had a PR which introduced types to all options objects - we had to revert it but it will be restored soon (I hope).
Once it's back, we can use it to enforce `string?` type for badge which will prevent users from passing numbers.
Regardless, Thanks a lot for the effort 💪

henrikra:
TS types help a lot yeah! I would love to get those types in!

shagun121:
'' this is will be equal to no  batch  and '0' this will be equal to 0 batch

PR #3836, opened 2018-08-23
[v1][Android] react-native v.0.57 compatibility / remove deprecated UIImplementationProvider

johakr:
`com.facebook.react.uimanager.UIImplementationProvider` has been removed as of react-native@0.57 ([commit](https://github.com/facebook/react-native/commit/506f92083806e91a90d9a213bcdd05ab3b2ba888)), breaking react-native-navigation v1 on android.

Hence removing the usage of UIImplementationProvider in react-native-navigation.

I know, all your efforts are on v2 already, but it would be great not to see v1 compatibility with react-native break too soon. :)


varungupta85:
@guyca Can we revert this merge now that we can provide the `UIImplementationProvider` again?

rotarsebastian:
Is there any way I could use the last version of react-native with the v1 react-native-navigation?

Issue #2766, opened 2018-02-21
FR: Allow disable some tabs


brunolemos:
My use case is that when the user haven't completed 100% of the signup (or his profile has been rejected/changes requested), he has access to 2 tabs and the other 2 should be grayed out and disabled (no press handler). We don't want to hide the tabs, only disable some, because we want them to transition to not-gray after he completes the signup.

I don't see any way to disable some specific tabs on react-native-navigation.
I see it's possible to change their icon, but not disable the press handler.

guyca:
@DanielZlotin What do you think about adding ability to handle BottomTab click in Js?

```js
options: {
  bottomTab: {
    title: 'Tab 2',
    icon: require('../images/two.png'),
    onClick: (index) => console.log(`tab ${index} clicked`)
  }
}
```

brunolemos:
Do you see any other use case for a custom press handler?
Personally I think a `disabled: true` would be enough.

DanielZlotin:
That was part of the initial roadmap but we postponed it.. 
Then if the onClick is provided we do nothing in native?
If we gonna add this now, we should probably add the onClick on any buttons (topbar, toptabs etc) as a function as well. (while still supporting the other methods)

guyca:
Right, if `onClick` is provided we don't handle the click in native. But in this case, should we rename it to `clickHandler`? To better reflect that this method handles the click and is not just a regular listener/callback.

DanielZlotin:
How about `onClickHandler`

guyca:
Sounds good

brunolemos:
I would personally vote for:

- Add a `disabled` prop
- Add `disabledIcon` and other props to provide styling for disabled tabs
- [optional] Add a `callback` (or `onPress`, ...) prop 
  - Native will change tab no matter if you pass this prop or not
  - Native will not change tab only if you pass `disabled: true`

guyca:
I'm a bit hesitant to support `enabled` property for bottom tabs as this is pretty much a UX anti pattern. For a reason known only to you (and your product manager), a bottom tab is disabled. The user clicks on it without getting any feedback (it's disabled dah..) and this kind of sucks.

I couldn't find anything about this in the [design guidelines on Android](https://material.io/guidelines/components/bottom-navigation.html#).

Honestly, I think if a user should be logged in and a tab that required login is visible - show an empty state encouraging the user to login.

brunolemos:
Oks the other solution should be enough

[...]

PR #3857, opened 2018-08-26
Allow the ability to disable the native tab transition event.

TomAtterton:
Add prop `disableNativeTransition` to bottomTabOptions.

Relating to issue #2766 

The prop it self is used to stop the tabNavigator from changing viewControllers, but still emitting the bottomTabSelected event to allow us to customise the transition. This matches the same way we handle the topButton events, so i see no issue resolving the issue like this.

* Only for iOS, if you approve i can start working on the android side.


marcorm:
@TomAtterton i really appreciate what you're doing... Are you going on with the Android side?

jayeshanandani:
@TomAtterton are you going to add android version as well?

PR #3895, opened 2018-08-31
Add sensorLandscape orientation option for Android to allow landscape…

adam-aerobotics:
… and reverseLandscape.


adam-aerobotics:
Apologies if I'm doing the pull request incorrectly. This is just a small fix we're using to allow the app to show in landscape orientation in both landscape configurations of a device, and I thought it might be useful for other people.

PR #3906, opened 2018-09-03
Reduce package size

retyui:
Update .npmignore


rgoldiez:
@retyui - I think they want to version control these files.

PR #3926, opened 2018-09-05
update hoist-non-react-statics to 3.x.x

ywongweb:
Fix crash when using the new `registerComponentWithRedux` API
https://github.com/wix/react-native-navigation/pull/3675#issuecomment-413923307

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
     "react-native": "*"
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


henrikra:
Btw looks like all dependencies in `package.json` are not pin pointed and this is very dangerous even with lock file. Are you guys thinking to pin point exact versions so there isn't such a big risk for breaking?

PR #3931, opened 2018-09-05
[v2] (restore) Typings for styling options and layouts

birkir:
This is an updated version of the typings.

Is there any way we can test this on Wix's own app repo before merging, or does it have to be published?

Also the ts error log would have been helpful too


yogevbd:
Amazing PR! I will personally test it in the wix app.
Thanks @birkir!

PR #3936, opened 2018-09-06
Remove less-than-correct v2 readiness claim from README

getaaron:
After sinking a ton of time into trying to get v2 working for a few different apps, I ran into [many known issues](https://github.com/wix/react-native-navigation/issues?q=is%3Aissue+is%3Aopen+label%3Av2+sort%3Acreated-desc).

I think it's fair to say we shouldn't recommend new users start with v2 quite yet, so this PR removes that suggestion for now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,8 +11,9 @@
 
 ## Important
 We are currently in late stages of development of [v2](https://github.com/wix/react-native-navigation/tree/v2), which is published to npm under `alpha` tag.
-* New to RNN? We recommend you start with [v2](https://github.com/wix/react-native-navigation/tree/v2).
-* Already using v1? now is the time to migrate your code base to [v2](https://github.com/wix/react-native-navigation/tree/v2), and you can easily do so with a single line of code using the [v1-v2 adapter](https://github.com/wix-playground/react-native-navigation-v1-v2-adapter).
+* ~~New to RNN? We recommend you start with [v2](https://github.com/wix/react-native-navigation/tree/v2).~~
+* ~~Already using v1? now is the time to migrate your code base to [v2](https://github.com/wix/react-native-navigation/tree/v2), and you can easily do so with a single line of code using the [v1-v2 adapter](https://github.com/wix-playground/react-native-navigation-v1-v2-adapter).~~
+* Do not use v2 yet since it is not close to ready yet and doesn't work for many use cases.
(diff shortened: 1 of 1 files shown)


sijad:
are you saying v2 is not production ready?
can you please let me know what was the issues you've encountered? I'm planning to upgrade to v2 too.

bgerhards:
@sijad if you check out the V2 documentation, this is one of the first things you'll see:

As we are in stage alpha, expect breaking API changes or use a specific version (for example "2.0.1234")

I would highly recommend not using V2, or any Alpha product, into production until it's out of Alpha, at least.

I like this pull request, however, I would recommend just removing all of the bulleted items. Let the user click into V2 and get all the information from that branch. Make it less confusing than it is in master today.

sijad:
@bgerhards v1 README stated:

> * New to RNN? We recommend you start with v2.
> * Already using v1? now is the time to migrate your code base to v2, and you can easily do so with a single line of code using the v1-v2 adapter.

> ⚠️Since we're focusing our efforts on v2, we are not accepting pull requests for v1 and are not addresing issues for v1.

last part is my biggest concern. if v1 is not maintained so v2 seems a wiser choice.

guyca:
Hey guys

I’m sorry to hear you encountered issues while integrating with RNN. We’re fully aware of the issues reported by the community and indeed we didn’t have much time to address them.
The reason for that is that we launched the new Wix app with RNN v2 (two weeks ago on Android, last week on iOS) and had to be available for the Wix app team.

Now that we’ve released the app on both platforms we should have more time to dedicate issues brought up by the community.

In regards to readiness of v2. While it does have issues which have not been addressed, v1 is simply not supported and we wanted to get feedback on v2. The input and contributions we got from the community has improved v2 a lot. FWIW, [v1 has a lot of issues](https://github.com/wix/react-native-navigation/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+label%3Av1) which can’t/won’t be addressed.

I believe the solution to your frustration is to become more involved in the project. Working on OSS is a huge challenge. Everyday I need to balance Wix’s needs and demands with the community’s. There’s no point sugar coating this - either you pick up java and objc so you can PR the project and fix the issues you encounter - or have some patience.

I’m closing the PR as I still stand behind what’s written in the front page - we do recommend using v2.

PR #3947, opened 2018-09-11
Allow also number to be passed

henrikra:
I have added number to be valid screen name too because then you can easily use enums to define your screen. Like this
```
export enum Screen {
  Login = 1,
  Home,
  Settings,
}
```

This way I just add new enum name and it will generate new id (number) for it. This reduces boilerplate when defining screen names


guyca:
Wow this a a very nice approach! Worth adding it to the docs as well 🙇

henrikra:
@guyca Yeah please add to the docs so you can try it out yourself too :)

Issue #4020, opened 2018-09-24
[V2] Broken links in README


jaden-young:
### Issue Description

The "Contribute" links in the V2 readme are broken. They point to https://github.com/wix/react-native-navigation/blob/v2/docs/WorkingLocally.md, but as the docs have been separated into `api` and `docs` folders, the correct link should be https://github.com/wix/react-native-navigation/blob/v2/docs/docs/WorkingLocally.md.

### Steps to Reproduce

Follow any of the "Contribute" links in the V2 readme, get GitHub's 404 page.

PR #4070, opened 2018-10-02
Fixed broken contributing link

bashbers:
#4020 

Fixed small issue where the contributing link would return a 404 page instead of redirecting the user to the contributing.md


yershalom:
V2 was merged into master. Thanks

PR #4105, opened 2018-10-08
Options resolving refactor

yogevbd:
- Bind RNNLayoutProtocol to presenters
- Merge options will now merge to the specific layout provided (component, stack, sideMenu)
- Options will be applied by the correct presenter
- Options will bubble up to all parents on leaf viewWillAppear
- Added unit tests for presenters


omar-dev:
This change breaks screens with large titles.
screen shot 2018-10-08 at 8 11 10 pm

rgoldiez:
@yogevbd - Couple of questions/issues with this PR:
- Looks like some of the `sideMenu` options recently added in a PR (by me) have been removed in these changes.  No longer does `RNNSideMenuSideOptions` accept the `shouldStretchDrawer` and `animationVelocity` props.  `animationType` (which is still there) doesn't seem to do anything.
- I have a custom `topBar` button component so that the standard 3-bar hamburger icon for a sideMenu can have a badge ... I'm doing this as a custom react native component button.... but this PR causes a crash when the `topBar` is passed a custom button component.

I would view the above as new bugs, preventing me from upgrading.  Do you agree that these are unintended side effects of this PR?

guyca:
@rgoldiez These are definitely regressions. Can you list all missing SideMenu options?
cc @yogevbd

rgoldiez:
@guyca @yogevbd - Best I can tell, `shouldStretchDrawer` and `animationVelocity` are missing now, and `animationType` is defined but not applied to the `sideMenu`

rgoldiez:
@guyca @yogevbd - It looks like `shouldStretchDrawer` and `animationVelocity` have been elevated to `sideMenuOptions` (left or right controlled), not `sideMenuSideOptions` (both left and right get the same value).  Is this the new method for settings these?  

`animationType` still does not seem wired in, but maybe I'm missing it.

Passing a component as a `topBar` button causes a crash (NOT GOOD!).

guyca:
@rgoldiez We'll look into SideMenu options tomorrow. Could you please open new issues instead of commenting in this PR? it makes it easier for us to prioritise. Thanks and sorry for the inconvenience.

PR #4129, opened 2018-10-12
Update documentation to avoid confusion on component layout type

adrienthiery:
After a while using v1, I finally got to try v2, but I had been stuck once with not being able to run the app on iOS (not sure where I was wrong).

Second try, I managed to run it but I stayed stuck on my one screen for an hour, trying to navigate to another, seeing no error, but seeing the screen was mounted but nowhere to be found.

To avoid that to happen to someone else, I propose adding these two sentences in the docs (which are a bit repetitive, but hopefully can save some time to the next person who wants to give it a try)


adrienthiery:
Not adrienharnay, but my pleasure 🤗

guyca:
Close enough? 🤷‍♂️

Issue #4076, opened 2018-10-03
[v2] [iOS] Bottom tabs visible when using layout bottomTabs visible: false


rendomnet:
### Issue Description

Can't hide bottomTabs using layout bottomTabs
It is visible on first tab and when I switch tab only then it will disappear.

```
  Navigation.setRoot({
    root: {
      bottomTabs: {
        children: [
...          
        ],
        options: {
          bottomTabs: {
            visible: false,
            animate: false,
            drawBehind: true
          }
        }

      }
    }
  });
```

### Steps to Reproduce / Code Snippets / Screenshots

Use  
```
options: {
          bottomTabs: {
            visible: false,
            animate: false,
            drawBehind: true
          }
        }
```
inside bottomTabs layout

---
### Environment
* React Native Navigation version: 2.0.2570
* React Native version: 0.56
* Platform(s) (iOS, Android, or both?): iOS
* Device info (Simulator/Device? OS version? Debug/Release?): both

rendomnet:
Any updates on this issue?

michellmorso:
im experiencing this as well

PR #4144, opened 2018-10-16
[V2][iOS] Apply bottom tabs visibility

gabiseabra:
This PR fixes an issue when using the `bottomTabs` layout with `options.bottomTabs.visible` set to `false` on iOS. It initially renders a visible `UITabBar`, which only disappears after navigating to another screen and going back. It should fix https://github.com/wix/react-native-navigation/issues/4076


stevenvdr:
The BottomTabsOptions have been refactored, so this fix can no longer be applied

gabiseabra:
Yeah, this fix is not necessary as of v2.0.2607. I'm closing this PR, thanks.

PR #4147, opened 2018-10-16
[V2][Android] destroy views on catalyst instance destroy to fix android crash

ericketts:
this fixes redscreen/crash when `ReactInstaceManager#recreateReactContextInBackground()` is called. This is a V2 adaptation of the previous fix for the same issue on V1, namely #3318


ericketts:
@guyca sure that works for me, I'll alter the code to just do a null check rather than a `try`/`catch`. 

As far as the `ClassCastException`, that was a defensive handling of the possibility that the user has screwed up their activity definition/inheritance, and [the cast](https://github.com/wix/react-native-navigation/pull/4147/files#diff-f76c5ae88668421418c4648039689a2bR181) of `ReactContextBaseJavaModule#getCurrentActivity()` to `NavigationActivity` fails. Almost certainly unnecessary, but like I said, a preemptive defensive handling.

ericketts:
@guyca ok I believe that I've addressed all your comments, please let me know if I've missed anything!

Issue #4148, opened 2018-10-17
bottomTabs backgroundColor not working


jp928:
Reproduce:

use style as below:

```js
bottomTabs: {
        visible: true,
        animate: false,
        // currentTabIndex: 0,
        // currentTabId: 'currentTabId',
        testID: 'bottomTabsTestID',
        drawBehind: false,
        backgroundColor: 'white',
        fontSize: 10,
        titleDisplayMode: 'alwaysShow'
}
```

### Environment
* React Native Navigation version: 2.0.2585
* React Native version: 0.57.1
* Platform(s) (iOS, Android, or both?): iOS
* Device info (Simulator/Device? OS version? Debug/Release?): simulator
![screen shot 2018-10-17 at 5 36 42 pm](https://user-images.githubusercontent.com/659612/47066899-47871400-d233-11e8-84be-f4d8ac4018d4.png)

guyca:
Are you sure it doesn't work on bot platforms?

jp928:
@guyca I am terrible sorry. I had update my issue report.

PR #4157, opened 2018-10-18
[V2][iOS] fix: Apply initial bottomTabs options

meznaric:
I had the same issue as the other user in #4148  -  `bottomTabs.backgroundColor` not working on iOS. At least not if set through `Navigation.setDefaultOptions()`.

The issue seems to be the fact that we never set values on the tab bar in `applyOptions` which I think is called when initialising.

Conveniently enough mergeOptions does exactly what we want and it seems to fix the issue for me.


meznaric:
Closing as it seem to be resolved by another merged PR. See commit: https://github.com/wix/react-native-navigation/commit/d6a4314280e23529989d00396fc9c5147d85ae12

Issue #3272, opened 2018-05-29
[v2] Modal screenBackgroundColor transparent bug


yusufyildirim:
### Issue Description
Hi,

I need a modal with transparent background. I set my modal's background to transparent but there is a bug I think. It comes transparent but turns to white.
### Steps to Reproduce / Code Snippets / Screenshots
Here is the gif:
![2018-05-29 18 59 59](https://user-images.githubusercontent.com/22980987/40670676-86d7f668-6372-11e8-83f2-a319cb66f623.gif)

---
### Environment
* React Native Navigation version: v2@alpha
* React Native version: 0.55.4
* Platform(s) (iOS, Android, or both?): iOS (Didn't test android)
* Device info (Simulator/Device? OS version? Debug/Release?): Simulator

guyca:
When a modal is displayed, the previous controller is removed from hierarchy. If you'd to display a transparent Modal over the previous controller, you need to add `modalPresentationStyle: 'overCurrentContext' option to you modal.

yusufyildirim:
I tried it before you said but nothing happened. I tried again and still same.

guyca:
Interesting. can you include some code snippets?

yusufyildirim:
Sure. This my my code for testing purpose. I just realized, this might be about stack i don't know. I'm gonna try without stack.

```
const tagModal = {
        stack: {
          children: [{
            component: {
              name: 'TagModal',
              passProps: { tag, onSuccess: this.onTagAdded },
              options: {
                screenBackgroundColor: 'transparent',
                modalPresentationStyle: 'overCurrentContext',
                topBar: {
                  visible: false,
                  animate: true,
                },
              },
            },
          }],
        },
      };
  
      navigation.showModal(tagModal);
```

yusufyildirim:
Yupp this is about stack. Working well without stack layout. I mean, like this:

```
const tagModal = {
        component: {
          name: 'TagModal',
          passProps: { tag, onSuccess: this.onTagAdded },
          options: {
            screenBackgroundColor: 'transparent',
            modalPresentationStyle: 'overCurrentContext',
            topBar: {
              visible: false,
              animate: true,
            },
          },
        },
      };
  
      navigation.showModal(tagModal);
```

guyca:
You added `modalPresentationStyle` to the wrong node. You need to add it to the root component presented in the modal, which in your case is the stack.

yusufyildirim:
[...]

PR #4159, opened 2018-10-19
[V2] Set UIModalPresentationStyle before presenting a modal view controlle…

muhammadsr:
Fixes issue with modal screen transparent background. Fixes this issue https://github.com/wix/react-native-navigation/issues/3272


yogevbd:
Fixed in latest version. thanks

PR #4188, opened 2018-10-22
[V2] topBar title options not applied when also using subtitle

rgoldiez:


pkyeck:
Had the same problem - but for me it had nothing to do with the subtitle. Was happening with or without subtitle. The `RNNTitleOptions.fontAttributes` was null.

But just looking at the two methods (`setupTitle` and `setupSubtitle`) lets me wonder why setting up the label is done so differently in both of them ...?

Cleaning everything up a bit lead to this:

```
-(UILabel*)setupTitle {
	CGRect titleFrame = self.titleView.frame;
	if (self.subtitle) {
		titleFrame.size.height /= 2;
	}
	UILabel *titleLabel = [[UILabel alloc] initWithFrame:titleFrame];
	titleLabel.textAlignment = NSTextAlignmentCenter;
	titleLabel.backgroundColor = [UIColor clearColor];
	titleLabel.autoresizingMask = self.titleView.autoresizingMask;
	
	NSDictionary* fontAttributes = [RNNFontAttributesCreator createFontAttributesWithFontFamily:[_titleOptions.fontFamily getWithDefaultValue:nil] fontSize:[_titleOptions.fontSize getWithDefaultValue:nil] color:[_titleOptions.color getWithDefaultValue:nil]];
	
	[titleLabel setAttributedText:[[NSAttributedString alloc] initWithString:self.title attributes:fontAttributes]];
	
	CGSize labelSize = [titleLabel.text sizeWithAttributes:fontAttributes];
	CGRect labelframe = titleLabel.frame;
	labelframe.size = labelSize;
	titleLabel.frame = labelframe;
	
	if (!self.subtitle) {
		titleLabel.center = self.titleView.center;
	}
	
	[titleLabel sizeToFit];
	[self.titleView addSubview:titleLabel];
	
	return titleLabel;
}
```

Where the `setupSubtitle` one looked almost identical.

And while we're at it, also remove the `@property (nonatomic, strong) NSDictionary* fontAttributes;` from `RNNTitleOptions.h` - it's never used and lead to this bug in the first place.

pkyeck:
can a core-member please take a look at this, should be pretty straight forward

PR #4191, opened 2018-10-22
[V2][Android] Auto blur on unmount option

ericketts:
This adds an Android-only flag `blurOnUnmount`, that if `true` when calling `push`, `pop`, `showModal`, `dismisModal`, as well as when `true` on the component's `static options(passProps) { ... }` return value, will cause RNN to check for any focused views when the component the flag was passed for is being unmounted, and blur said view if it exists, causing not just the view to lose focus, but also the soft keyboard to be dismissed.

We added this to our fork of RNN to tackle issues stemming from an inability to dismiss the soft keyboard on unmounting components, due to the `disappear`/`unmount` lifecycle events of the component coming too late for our keyboard dismissal calls to make it across the native<->js bridge before the component was unmounted. 

As this functionality is hidden behind a flag and does not alter base functionality, and the issue it seeks to solve might have been encountered by other users of RNN, I figured I might as well submit a PR to see if there is interest in merging this back into the base.


lib/src/interfaces/Options.ts
```diff
@@ -753,4 +753,10 @@ setRoot: {
    * #### (iOS specific)
    */
   rootBackgroundImage?: ImageRequireSource;
+  /**
+   * Enable or disable automatically blurring focused input, dismissing keyboard on unmount
+   * #### (Android specific)
+   * @default false
```
Collin3: Just curious, is there a reason why this was defaulted to false? I ran into the Android issue where the keyboard would stay up after navigation events and was wondering if that is ever expected behavior? It would definitely make my life easier if this was defaulted to true and I didn't have to override it on all of the navigation actions.

ericketts: yeah I defaulted it to false because in the sense of expected behavior its technically a breaking change, so I wanted to have it only be opt in. However if you'd like to have it be the default behavior for your project, I'm fairly sure that you can configure that to be the case [with existing APIs provided](https://wix.github.io/react-native-navigation/#/docs/styling?id=enabling-persistent-styling-properties) (note I didn't actually verify this but it seems logically sound 😄)

Collin3: Ahh interesting. I figured it was a bug because it was Android specific and we were only seeing the keyboard stay open after navigation events when we had a side menu layout. When we changed the layout to be a stack based app, then we started seeing the keyboard being dismissed on navigation events.

Either way, I think the documentation you referenced looks like it should work. Thanks a lot!

Issue #3892, opened 2018-08-30
[V2] Overlay closes and opens keyboard on iOS


pribeh:
On iOS, calling an overlay while the keyboard is open closes the keyboard. After dismissing the overlay, the keyboard opens again. Is there a reason why this is? Or can this be considered a bug?

## STEPS ##

- Navigation.push screen with textinput
- Focus textinput to load keyboard
- Navigation.showOverlay component (in our case a notification type component with smallish size)

## RESULT ##

As soon as the overlay loads, the keyboard and textinput collapse. When the overlay is dismissed, the keyboard and textinput expand again. 

We're using overlays for in-app notifications so this is quite bothersome for the end user in this instance. 

---
### Environment
* React Native Navigation version: ^2.0.2513
* React Native version: 0.55
* Platform(s) (iOS, Android, or both?): iOS
* Device info (Simulator/Device? OS version? Debug/Release?): iOS 10/11

pribeh:
Is anybody else experiencing this?

nikolas7892:
+1

trqson:
+1

tomhicks:
@pribeh or anyone else can you confirm this is fixed? I'm still getting this issue on iOS, version 2.12.0.

FRizzonelli:
@tomhicks @guyca I think the fix is in this line of the commit

`- [overlayWindow makeKeyAndVisible];`
`+ [overlayWindow setHidden:NO];`

The problem is that isn't included in master or in 2.12.0. Maybe something went wrong with merge/other commits?

pribeh:
@tomhicks yeah, it doesn't seem to be fixed and @FRizzonelli is correct that the fix looks to be missing.

FRizzonelli:
#4778 PR from @tomhicks, @guyca can you approve and merge it?

pribeh:
Looks like [overlayWindow makeKeyAndVisible] was added to make sure OS dialogs appear overtop overlays. The side effect is that keyboards or keyboard related textinputs are hidden when overlays are displayed. Anybody have any ideas on a way around this?

marf:
Same problem here

geroale:
Same :(

pribeh:
This issue is now resolved as of https://github.com/wix/react-native-navigation/commit/e9ca247a524e474daf3ccf56989289ce679fc063 by @yogevbd

[...]

PR #4211, opened 2018-10-24
Fixes overlay closes keyboard - #3892

yogevbd:


brandons:
This breaks any overlays that need keyboard input. 😐

PR #4223, opened 2018-10-25
Make push with passProps type safe

henrikra:
You have C screen where you can go to from two different screens: A and B. You also have to pass some props for screen C when you push that screen. So you do pass props from A and B as you should. Later you add some other prop for screen C that is required. Before this PR you can easily forget to add the new prop for all the places where you are pushing screen C. Now with the generic you cannot forget since compiler gives you error!

Tell me what your guys think. I think this would increase type safety a lot!

Example code

```typescript
// ---- With this PR ----

interface SomeProps {
  someStringProp: string;
}

Navigation.push<SomeProps>(this.props.componentId, {
  component: {
    name: Screen.SomeScreen,
    passProps: { someStringProp: 1 }, // compiler error
  },
});

Navigation.push<SomeProps>(this.props.componentId, {
  component: {
    name: Screen.SomeScreen,
    passProps: { someStringProp: 'lol' }, // everything is fine
  },
});


// ---- Without this PR ----

Navigation.push(this.props.componentId, {
  component: {
    name: Screen.SomeScreen,
    passProps: { someStringProp: 1 }, // no compilation errors :(
  },
});
```


henrikra:
@guyca can we try this out? Or do you have some doubts 😄

guyca:
Oh no doubts, was just a bit busy. Thanks @henrikra, really liking your pr's 💪

Issue #3691, opened 2018-08-05
[V2] Add support for navigate back event


noamco:
### Issue Description

Add support for navigate back event triggered by user pressing the back/up button, physical back button or swiping left.

wbercx:
+1. Would be great to be able to intercept back button presses in v2, for showing confirmation modals etc. This is the only loss in functionality I've run into so far when moving from v1 to v2.

angelos3lex:
Is this planned to be implemented sometime?

jarrx:
Would really need this funktion as well.

Issue #3480, opened 2018-07-02
Emit nativeEvent when a screen is popped with back button


guyca:

ItsNoHax:
I need this! I was wondering why navigationButtonPressed wasn't catching my back button press.

PR #4226, opened 2018-10-25
[V2][Android] fire onNavigationButtonPressed for android back button

ShaMan123:
V2
-------
Fire onNavigationButtonPressed for android back button
---------

`onNavigationButtonPressed` doesn't fire when pressing android's top bar back button.
This makes it difficult/impossible to handle the navigation state of a stack and causes the implementation of the native back button to be a pain.
This PR fixes this issue on android.

Use:
```
Navigation.events().registerNavigationButtonPressedListener(({ buttonId }) => {
            if (buttonId === 'RNN.back') {
                   // User pressed topBar backButton
                   // Navigation has popped the screen => update your code if needed
            }
        });
```

I didn't code on iOS yet so I don't know if there's a need for a fix there too.
See #3480, #3691 & I recall at least another issue about this.


guyca:
Hey there, I'm not sure this change is desirable.
1. What motivated you to make this change?
2. This change breaks parity with iOS ⚠️ 

If you wish to handle the back button - I suggest you set it yourself for now until we have a better solution.

ShaMan123:
as mentioned regarding motivation:
There is no way to react to a user pressing the native top bar back button, meaning there is no way to react to a screen being popped, resulting in a stack that is unmanageable.
Note that the `commandsObserver => Navigation.events().registerCommandListener` doesn't fire a pop either in this case.
I have no clue regarding ios - but this is a **MUST HAVE** in my opinion. Is there no way to do the same on ios?
This takes away a great deal from the effectiveness of the native top bar back button.

How would you handle a stack?


BTW, I have submitted another PR: #4227  that too lacks ios native code but could make this PR redundant.

guyca:
@ShaMan123 
I'm not reluctant to merge this PR as it doesn't fully address the issue. You can easily handle the back button by setting your own back button.
This is how we handle the TopBar back button in the Wix app and it works without any issues.

If all you wish is to run some code when the screen is popped - then use `componentWillUnmount`

bdrobinson:
@guyca could you point me towards some documentation on how to *"set your own back button"*? I can see that I can set `backButton: { visible: true }` when creating a modal, and the back button does indeed appear, but as @ShaMan123 notes there's no way to listen to when it gets pressed. Am I missing something?

Are you suggesting we should set the button using `leftButtons` or something? If so, wouldn't we need to provide our own asset of the back icon etc?

pkyeck:
I would propose to just fire the events like when popping the screen manually so we can handle them as well via

```
Navigation.events().registerCommandListener((...args) => {

});
```

BTW: Same is true for sideMenu - you can't be notified of the sideMenu-close-event when the user just clicks the overlaying content. So a toggleMenu function is not possible :(

ShaMan123:
I've implemented that on a pop in android => [check this](https://github.com/ShaMan123/react-native-navigation/tree/v2-addedFunctionality)

Issue #3669, opened 2018-07-31
you should be able to check the stack


reganperkins:
### Issue Description
There is currently no documented way to check what elements are in the stack or to check the stack length.

### Steps to Reproduce / Code Snippets / Screenshots
In the app I am creating there are multiple ways a user can navigate to different pages. The different ways a user might navigate to the page effects if I should use pop, pushTo or setStackRoot. For example lets say in the drawer I have a list of teams and workspaces, when changing to a team or workspace from the drawer I would use setStackRoot to make sure all the previous stack is cleared. When going through the main page I would be using pushTo when entering a team and like wise pushTo when entering a workspace. However when using the main page navigation to move back from the workspace to the team I would use pop. 

The issue comes in when you mix using navigation from the drawer and the main page. In the case of having used the drawer to get to the workspace and then using the main page navigation to go back to the team, the team is not in the nav stack and pop workspace fails. being able to check that the component is in the stack or that the stack is of a certain length would really help. 

If there is a method I can use that I did not find please let me know

---
### Environment
* React Native Navigation version: Current
* React Native version: .56
* Platform(s) (iOS, Android, or both?): All
* Device info Any

ItsNoHax:
Same goes for checking the current visibility of a drawer

PR #4227, opened 2018-10-25
V2 add method getStackChildrenIds to API

ShaMan123:
V2
----
Add method getStackChildrenIds to API
----

This PR enhances the js side of the repo, enabling to get the ids of the children of a stack.
Currently, managing the state of a stack is a real headache.
Adding this method is extremely useful because of the dynamic nature of a stack.
I would argue it is imperative. 

Use:

```
Navigation.getStackChildrenIds(stackId)
     .then(({ childrenIds }) => updateStackStore(childrenIds))
     .catch(error => handleMe(error))

```

### Help needed: 
iOS implementation of `getStackChildrenIds`.
Feel free to PR!

Enhancing #4226 
Issues: #3691 , #3669


guyca:
Hey there, Thanks for the PR 👍 
We prefer to avoid having a bunch of query methods is there's no limit to user's needs. Instead we expose a set of [events](https://wix.github.io/react-native-navigation/v2/#/docs/events) which we believe cover all reasonable uses cases.

To figure out which child components are pushed to a given stack, register a command listener and keep track of all child components pushed to and popped from that stack.

ShaMan123:
Just saying that if you guys want this repo to be adopted by many developers I think you should consider other ways of using it, not only the way you have in mind / are used to.
I'm sure that for you guys it's extremely straight forward but getting the library deeply implemented takes time and too much surrounding code in my opinion.
Still I do prefer using it over others.
Anyways, it's your call.

pkyeck:
@guyca could you explain how to use the events you were talking about to keep track of the current stack?

`componentDidDisappear` is called when we push a new screen, but we don't know whether or not we should remove the old screen from the stack b/c we don't know about push/pop at this point.

`registerCommandListener` only handles user commands but not the "built-in back-button". so we'll miss a lot of "pops".

PR #4266, opened 2018-11-01
Overlay iOS key window management

brandons:
In our application we make extensive use of overlays and sometimes need to show more than one at the same time. Currently each overlay that is displayed automatically becomes the key window. This creates a problem if you need to display an overlay at the same time as a view that may need the keyboard. One example of this is using an overlay for notifications. With the current implementation the keyboard would dismiss if a notification were to display. 

This is a breaking PR. It adds a flag called `becomeKeyWindow` that can be passed when showing an overlay. If an overlay has this flag, that window will become the key window. If not, the window will not become the key window. Additionally it stores windows that include this flag allowing for the display of multiple overlays and will focus the next window upon dismissal. 

An alternate solution would be an `avoidKeyWindow` option that would implement the opposite behavior. I'm also open to this solution, however I'm assuming the main use case would be for overlays that do NOT require the keyboard.

Tests likely need to be updated and I'm happy to do so once I receive feedback.


ouabing:
Yeah it's really annoying. I use 'react-native-view-shot' to take fullscreen screenshot, but it will only capture the key window. If I display a mini overlay over the screen, it will only capture the overlay.

henrikra:
@brandons Can you tell me when you want to use modal vs overlay? The documentation doesn't really say how do they differ 🤔

PR #4276, opened 2018-11-02
Update detox package

thinklinux:
I needed to update the package in order to build the playground project.


thinklinux:
This is actually the same issue. I don't know why it was closed - https://github.com/wix/Detox/issues/983

PR #4288, opened 2018-11-05
Fix docs. Change bottomTab title to text.

minhloi:
I believe there is a type in the docs. The title prop in bottomTab should be text.


guyca:
Thanks!

Issue #4295, opened 2018-11-06
Add optional componentId param to bindComponent


guyca:
`bindComponent` extracts the componentId from the component provided to it. This bad design causes unneeded coupling on the component's private logic and assumes it has componentId in props.

We can mitigate this by adding an optional `componentId?: string` argument, is it exist we won't resolve componentId from props.

guyca:
Resolved in #4299

PR #4298, opened 2018-11-07
add optional componentId to bindComponent

JasonMannon:
Adds optional componentId param to bindComponent 

https://github.com/wix/react-native-navigation/issues/4295

### DIFF
--- a/docs/api/ComponentEventsObserver.md
+++ b/docs/api/ComponentEventsObserver.md
@@ -10,7 +10,7 @@
 
 ## bindComponent
 
-`bindComponent(component: Component<any>, componentIdArg: string): EventSubscription`
+`bindComponent(component: Component<any>, componentIdArg?: string): EventSubscription`
 
 [source](https://github.com/wix/react-native-navigation/blob/v2/lib/src/events/ComponentEventsObserver.ts#L35)
 
--- a/docs/api/ComponentEventsObserver.md
+++ b/docs/api/ComponentEventsObserver.md
@@ -10,7 +10,7 @@
 
 ## bindComponent
 
-`bindComponent(component: Component<any>): EventSubscription`
+`bindComponent(component: Component<any>, componentIdArg: string): EventSubscription`
 
 [source](https://github.com/wix/react-native-navigation/blob/v2/lib/src/events/ComponentEventsObserver.ts#L35)
 
@@ -63,5 +63,3 @@
 [source](https://github.com/wix/react-native-navigation/blob/v2/lib/src/events/ComponentEventsObserver.ts#L69)
 
 ---
-
-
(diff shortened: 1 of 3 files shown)


docs/api/ComponentEventsObserver.md
```diff
@@ -10,7 +10,7 @@
 
 ## bindComponent
 
-`bindComponent(component: Component<any>): EventSubscription`
+`bindComponent(component: Component<any>, componentIdArg: string): EventSubscription`
```
henrikra: Mark also as optional here?

lib/src/events/ComponentEventsObserver.test.tsx
```diff
@@ -137,7 +176,11 @@ describe('ComponentEventsObserver', () => {
     subscription.remove();
 
     uut.notifyComponentDidAppear({ componentId: '123', componentName: 'doesnt matter' });
-    expect(didAppearFn).toHaveBeenCalledTimes(1);
+    expect(didAppearFn).toHaveBeenCalledTimes(1);it(`doesnt call unimplemented methods when optional componentId param is present`, () => {
```
henrikra: New line before `it`? :D

lib/src/events/ComponentEventsObserver.ts
```diff
@@ -38,8 +38,9 @@ export class ComponentEventsObserver {
     this.nativeEventsReceiver.registerPreviewCompletedListener(this.notifyPreviewCompleted);
   }
 
-  public bindComponent(component: React.Component<any>): EventSubscription {
-    const componentId = component.props.componentId;
+  public bindComponent(component: React.Component<any>, componentIdArg?: string): EventSubscription {
+    const componentId = componentIdArg ? componentIdArg : component.props.componentId;
+
     if (!_.isString(componentId)) {
```
henrikra: This check is not needed since the types are correct here :)

docs/api/ComponentEventsObserver.md
JasonMannon: @henrikra What's the preferred way to mark that argument as optional?

henrikra: `componentIdArg?: string` like this :)

lib/src/events/ComponentEventsObserver.ts
```diff
@@ -38,11 +38,9 @@ export class ComponentEventsObserver {
...
 
-  public bindComponent(component: React.Component<any>): EventSubscription {
-    const componentId = component.props.componentId;
-    if (!_.isString(componentId)) {
-      throw new Error(`bindComponent expects a component with a componentId in props`);
-    }
+  public bindComponent(component: React.Component<any>, componentIdArg?: string): EventSubscription {
+    const componentId = componentIdArg ? componentIdArg : component.props.componentId;
+
     if (_.isNil(this.listeners[componentId])) {
```
henrikra: Also we don't need to use lodash here you can do it like this `if (!this.listeners[componentId]) {` 👍

henrikra:
@guyca Wanna check this out?

guyca:
Hey @JasonMannon, Thanks a lot for the pr and @henrikra for reviewing. We merged #4299 which pretty much does the same thing and will be available in 2.1.2-snapshot.29

PR #4321, opened 2018-11-14
[V2] Fix build errors on Xcode 10

### LINKED ISSUE #3608: [V2] Build error in Xcode 10 "Multiple commands produce..."
### Issue Description

After following the [installation instructions](https://wix.github.io/react-native-navigation/v2/#/docs/Installing) multiple times and clearing caches in between I still get build errors in Xcode 10:

image

---
### Environment
* React Native Navigation version: 2.0.2425
* Device info (Simulator/Device? OS version? Debug/Release?): Simulator 
* No other dependencies except `react-native-navigation` installed

rajivshah3:
This PR deletes the `Copy Files` step in the Build Phases. It was creating duplicate files which didn't agree well with Xcode's new build system.

Fixes #3608 

Tested by successfully building the playground app with Xcode 10.1 on both the new and legacy build systems.


rajivshah3:
Yikes... just realized I used the wrong branch

henrikra:
Can you explain bit more why those files were copied and why we don't need to copy them anymore? Also would this mean that after this PR is merged the project is not built correctly on Xcode version below 10?

rajivshah3:
>  why those files were copied 

I'm not entirely sure. I'd have to look through the commit history to see why this was added.

> why we don't need to copy them anymore? 

It seems like they're duplicates. Xcode's new build system seems to treat some things that used to be warnings as errors, so that appears to be the case here. React Native and other projects have recently run into similar issues as well.

> Also would this mean that after this PR is merged the project is not built correctly on Xcode version below 10?

No. I've tested it on the new build system (the one that was having problems) as well as the legacy build system (the same one that Xcode 9 uses). It also builds successfully on the CI, which (judging by the logs) appears to be Xcode 9.3 or 9.4 since it's building with the iOS 11.3 SDK.

rajivshah3:
Fixed in b5d300f

PR #4328, opened 2018-11-15
Android: Use proper key for bottom tab height

Krizzu:
Hey,

Noticed wrong key was used for Bottom tabs height, so created a quick patch for it.


guyca:
Thanks!

PR #4335, opened 2018-11-19
Add support to pass any argument to <Provider />

edvinasbartkus:
I would like to suggest this change for `registerComponentWithRedux` method. For the last argument to accept a set of props that can be passed to `<Provider />`. This is backward compatible, but additionally, it would enable new use cases:

1. Using [unstated](https://github.com/jamiebuilds/unstated) instead of `redux`:

```js
import { Provider, Subscribe, Container } from 'unstated'

class Counter extends Container {
// ...
// <as per https://github.com/jamiebuilds/unstated#example>
// ...
}

Navigation.registerComponentWithRedux('screen.Home', () => HomeScreen, Provider, {inject: [counter]})
```

2. Enable `persistor` for redux

```js
import 'redux-persist'

const persistor = ...

Navigation.registerComponentWithRedux('screen.Home', () => HomeScreen, Provider, {store: store, persistor: persistor})
```

### DIFF
--- a/docs/api/ComponentRegistry.md
+++ b/docs/api/ComponentRegistry.md
@@ -2,9 +2,9 @@
 
 ## registerComponent
 
-`registerComponent(componentName: string, getComponentClassFunc: ComponentProvider, ReduxProvider: any, userStore: any): ComponentType<any>`
+`registerComponent(componentName: string, getComponentClassFunc: ComponentProvider, Provider: any, providerParams: any): ComponentProvider`
 
-[source](https://github.com/wix/react-native-navigation/blob/v2/lib/src/components/ComponentRegistry.ts#L10)
+[source](https://github.com/wix/react-native-navigation/blob/v2/lib/src/components/ComponentRegistry.ts#L9)
 
 ---
 
--- a/docs/api/ComponentWrapper.md
+++ b/docs/api/ComponentWrapper.md
@@ -2,17 +2,17 @@
 
 ## wrap
 
-`wrap(componentName: string, OriginalComponentClass: React.ComponentType<any>, store: any, componentEventsObserver: any, ReduxProvider: any, reduxStore: any): React.ComponentType<any>`
+`wrap(componentName: string, OriginalComponentGenerator: ComponentProvider, store: Store, componentEventsObserver: ComponentEventsObserver, Provider: any, providerParams: any): ComponentClass<any>`
 
-[source](https://github.com/wix/react-native-navigation/blob/v2/lib/src/components/ComponentWrapper.tsx#L7)
+[source](https://github.com/wix/react-native-navigation/blob/v2/lib/src/components/ComponentWrapper.tsx#L12)
 
 ---
 
-## wrapWithRedux
+## wrapWithProvider
 
-`wrapWithRedux(WrappedComponent: any, ReduxProvider: any, reduxStore: any): React.ComponentType<any>`
+`wrapWithProvider(WrappedComponent: ComponentClass<any>, Provider: any, providerParams: any): ComponentClass<any>`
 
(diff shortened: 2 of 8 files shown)


henrikra:
I think we should totally remove Redux dependency from this project and make view rendering in a way that any libraries can be used. We talked about this @guyca

guyca:
I'm all for removing any reference of redux from RNN. I think for now we can improve upon the solution proposed by @edvinasbartkus and deprecate `wrapWithRedux` in favor of a more general implementation; `wrapWithProvider`.

```js
static wrapWithProvider(WrappedComponent: React.ComponentClass<any>, Provider: any, providerParams: any): React.ComponentClass<any> {
    class ReduxWrapper extends React.Component<any, any> {
      render() {
        return (
          <Provider {...providerParams}>
            <WrappedComponent {...this.props} />
          </Provider>
        );
      }
    }
```

while still limited, it's less coupled on redux and doesn't include workarounds like the ternary operator checking `dispatch`. What do you guys think?

henrikra:
Yeah this is bit better. Maybe `wrapWithComponent` is more generic? Can you @edvinasbartkus try to make the more generic solution for this until we can totally remove redux and all wrap hacks from RNN?

edvinasbartkus:
@henrikra @guyca sounds great about having a generic solution for any type of the wrapper. WIll make the changes for the PR later today.
Thanks!

edvinasbartkus:
@henrikra @guyca 
Updated the PR with the new `registerComponentWithProvider` method.
Should I add backward capability to call `registerComponentWithRedux`?

henrikra:
Before going through the PR can you show me examples of how to use this `registerComponentWithProvider`?

edvinasbartkus:
1. In my case where I am using it with `unstated`

```js
import {Provider, Container} from 'unstated'
const container = new CustomContainer(data: []) // extends Container

Navigation.registerComponentWithProvider('screen.HomeScreen', () => HomeScreen, Provider, {inject: [container]})
```

```js
const HomeScreen = () => <Subscribe to={[CustomContainer]}>{container => <View />}</Subscribe>
```


2. In case of good old redux
```
import {Provider} from 'react-redux'

const store = 

Navigation.registerComponentWithProvider('screen.HomeScreen', () => HomeScreen, Provider, {store})
```

henrikra:
Hmm okay. I think this solution is better 👍 

Would it possible to do even more generic solution? That we would need only one `registerComponent` which can handle all the cases. The usage would be something like this
```ts
// with redux

import { Provider } from 'react-redux';
import { createStore } from 'redux';

import Home from './myScreens/Home';

const store = createStore();

Navigation.registerComponent('screen.HomeScreen', Home, WrappedComponent => (
  <Provider store={store}>
    <WrappedComponent />
  </Provider>
));

// with unstated
import { Provider, Container } from 'unstated';
const container = new CustomContainer({ data: [] });

Navigation.registerComponent('screen.HomeScreen', Home, WrappedComponent => (
  <Provider inject={[container]}>
    <WrappedComponent />
  </Provider>
));

// with nothing special
Navigation.registerComponent('screen.HomeScreen', Home, WrappedComponent => <WrappedComponent />);

```

What do you think @edvinasbartkus @guyca ?

henrikra:
Can you make wip pr so it is easier to look at?

guyca:
Figured out what the issue was :(
I registered my test screen twice by mistake, the second time without wrapping it with providers 🤦‍♂️

guyca:
Alright, published my [PR](https://github.com/wix/react-native-navigation/pull/4427) - @edvinasbartkus @henrikra your reviews and opinions is appreciated.

guyca:
After looking into this more, I realised there's no need for another api to wrap components with providers, we can use the existing `registerComponent` api by passing a functional component to it as shown in #4427

PR #4341, opened 2018-11-21
[Fix typescript] Add title alignment to OptionsTopBarTitle

minhloi:
Title alignment works on Android; however, the OptionsTopBarTitle interface (in Options.ts) is missing the optional alignment field preventing us to use that option.
```
    topBar: {
      title: {
        alignment: "center"
      }
    },
```


henrikra:
Add in the comment that this is Android only property?

minhloi:
Added android specific comment. Thanks.

PR #4373, opened 2018-11-29
[V2] Add props to TopBarButton type

gsdatta:
Based on my investigation, it looks like props can be passed to top bar buttons. If so, we should expose that in the types.

I modeled this off of other top bar options, but should this actually just be `LayoutComponent`?


gsdatta:
Ah, I see that #4373 is also open to address this.

guyca:
Thanks!

PR #4386, opened 2018-12-04
[V2]Fix disableIconTint iOS

Natkarock:
On iOS disableIconTint not working. I improved the solution of #4312

### DIFF
--- a/lib/ios/RNNNavigationButtons.m
+++ b/lib/ios/RNNNavigationButtons.m
@@ -114,8 +114,13 @@
 	
 	if (color) {
 		[textAttributes setObject:color forKey:NSForegroundColorAttributeName];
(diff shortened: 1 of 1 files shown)


guyca:
Hey there,
`disableIconTint` is actually obsolete - icon color is an optional property. If you don't want to color a specific icon then don't assign it a color.

Natkarock:
> Hey there,
> `disableIconTint` is actually obsolete - icon color is an optional property. If you don't want to color a specific icon then don't assign it a color.

When I not set a color to icon, in iOS I see default blue tint. But i want to see my real multicolor icon.

rsokz:
> Hey there,
> `disableIconTint` is actually obsolete - icon color is an optional property. If you don't want to color a specific icon then don't assign it a color.

With no color set, the icon defaults to the default blue tint. There should be a way to prevent this from happening in situations where you have a multicolor icon

PR #4427, opened 2018-12-10
Support Context api

guyca:
~~This PR adds improved support for React's Context api.~~ Actually no changes were needed 🤷‍♂️ 
As he playground app still used RN 51, I had to upgrade it first to RN 0.57.7 which turned out to be a bit of a challenge. I also took the opportunity to upgrade TypeScript to 3.2.

## Context API
```jsx
Navigation.registerComponent('navigation.playground.ContextScreen', () => (props) => (
  <TitleContext.Provider value={'Title from Provider'}>
    <ContextScreen {...props} />
  </TitleContext.Provider>
));
```
Obviously, using static context works as well without needing to wrap the component.

## Redux
```jsx
Navigation.registerComponent('navigation.playground.ReduxScreen', () => (props) => (
  <Provider store={reduxStore}>
    <ReduxScreen {...props} />
  </Provider>
));
```

## Plain Component
```jsx
Navigation.registerComponent('navigation.playground.MyScreen', () => MyScreen);
```

This pattern supports any library requiring the user to wrap the base component.


henrikra:
Hmm can you add more example to PR description? For example how to use the new `registerComponent` with Redux? Also with nothing: just plain component. So can we get rid of all Redux with this PR?

henrikra:
And how the api is used if I just have plain component?

guyca:
@Hellegab As usual, no change.

henrikra:
So basically we could deprecate all those redux wrappers? Aka we would not have any redux dependencies in this project?

henrikra:
This is good! Tell me when you are done so I can review this also! :)

guyca:
@henrikra PR's actually ready - I'm just trying to figure out why our CI is giving me such hard time. For some reason Detox times out...

henrikra:
@guyca Ok I can check it later today 👍 

Meanwhile could you check my iOS PR? I really need your guys help and opinions on this https://github.com/wix/react-native-navigation/pull/4434
This is very high prio for me at the moment ⚠️  I am grateful if you can review it 🙇

guyca:
@henrikra Absolutely - we'll review it tomorrow

henrikra:
Good to hear!

Hey since this PR is doing two things: can you specify files that has to be reviewed that are related to `registerComponent`? So people don't have to go through files that aren't related to that :D

henrikra:
Looks good! On suggestions

I think we could get the API from this

```
Navigation.registerComponent('navigation.playground.ReduxScreen', () => (props) => (
  <Provider store={reduxStore}>
    <ReduxScreen {...props} />
  </Provider>
));
```

to this
```
Navigation.registerComponent('navigation.playground.ReduxScreen', (props) => (
  <Provider store={reduxStore}>
    <ReduxScreen {...props} />
  </Provider>
));
```

Aka removing one arrow function. If you look `lib/src/components/ComponentWrapper.tsx` line 20 the `OriginalComponentGenerator` just calls it self without params so I guess there is no point of having that extra function there? Wouldn't it be more clear without it se we can just directly pass those components?

guyca:
@henrikra Thanks for reviewing!
I don't think that's change is possible as the arrow function returns a component, which in our case is a functional component.

Also, I had to change the api 😕
```js
Navigation.registerComponent('navigation.playground.ContextScreen', () => (props) => (
  <TitleContext.Provider value={'Title from Provider'}>
    <ContextScreen {...props} />
  </TitleContext.Provider>
), ContextScreen);
```
The motivation is that we need to hoist statics from the actual component to the wrapping component in order to support `static options`. Do you have a better suggestion for the api?

henrikra:
I think I have bit better idea. Could it be like this

```ts
Navigation.registerComponent('navigation.playground.ContextScreen', (WrappedScreenComponent) => (
  <TitleContext.Provider value={'Title from Provider'}>
    <WrappedScreenComponent />
  </TitleContext.Provider>
), MyComponentScreen);
```

So basically `MyComponentScreen` is my screen component which can be what ever. And `WrappedScreenComponent` what is coming back from `wrap` in `ComponentWrapper`. And then as you can see in the example I can wrap it what I want. If I just have plain component then api is like this:

```ts
Navigation.registerComponent('navigation.playground.ContextScreen', (WrappedScreenComponent) => <WrappedScreenComponent />, MyComponentScreen);
```

guyca:
```jsx
Navigation.registerComponent('navigation.playground.ContextScreen', (WrappedScreenComponent) => <WrappedScreenComponent />, MyComponentScreen);
```

This is a breaking change - we need to avoid breaking changes for now. The second argument needs needs to be a method which returns a Component, not an instance of a component.

henrikra:
Okay what about same points as I said before but with extra function

```ts
Navigation.registerComponent('navigation.playground.ContextScreen', () => (WrappedScreenComponent) => (
  <TitleContext.Provider value={'Title from Provider'}>
    <WrappedScreenComponent />
  </TitleContext.Provider>
), MyComponentScreen);
```

```ts
Navigation.registerComponent('navigation.playground.ContextScreen', () => (WrappedScreenComponent) => <WrappedScreenComponent />, MyComponentScreen);
```

henrikra:
Or what about this kind of API: we provide HOC from library like `withNavigation`

```ts
Navigation.registerComponent('navigation.playground.ContextScreen', () => (WrappedScreenComponent) => (
  <TitleContext.Provider value={'Title from Provider'}>
    <WrappedScreenComponent />
  </TitleContext.Provider>
), MyComponentScreen);
```

```ts
// in component file
import {withNavigation} from 'react-native-navigation';

class MyCoolComponent extends React.Component {}

export default withNavigation(MyCoolComponent);


// in screen file
import MyCoolComponent from './MyCoolComponent';
Navigation.registerComponent('navigation.playground.ContextScreen', MyComponentScreen);
```

What do you think about this? Yeah it is breaking but we can change function name so it is NOT breaking :D

guyca:
Hey @henrikra, I think we should merge the feature now as I don't have much time to work on it anymore. Lets keep discussing the api in a new issue - we can consider this api experimental in the next few versions until we're both satisfied with it.

henrikra:
Sounds good

Issue #3454, opened 2018-06-27
[v2] `topBar.buttonColor` is ignored for back button


ilovett:
```
topBar: {
  buttonColor: 'red'
}
```

Did not work, but 

```
backButton: {
  color: 'red'
}
```

Did

ujwal-setlur:
This seems to be a difference between iOS and android. On android, I need to use 
```js
backButton: {
  color: 'red'
}
```

on iOS, I need to use:
```js
buttonColor: 'red'
```

derrh:
Where exactly are you specifying

derrh:
Where are you specifying:

```
backButton: {
  color: 'red'
}
```

within `topBar`? within the screen's static options object? I've tried that everywhere I can think to try and I'm not having any luck affecting the color of the Android back button.

ujwal-setlur:
I set it in the default options:

```js
 // default options for navigation bar
  Navigation.setDefaultOptions({
    topBar: {
      backButton: { // android
        color: "rgba(60, 109, 240, 0.87)"
      },
      buttonColor: "rgba(60, 109, 240, 0.87)", // iOS
      title: {
        color: "rgba(60, 109, 240, 0.87)",
        fontFamily: "Nunito-Bold",
        fontSize: 18
      }
    },
    bottomTab: {
      iconColor: "black",
      selectedIconColor: "#3c6df0",
      backgroundColor: "white",
      fontFamily: "Nunito",
      fontSize: 10
    }
  });
```

oldtinroof:
thanks @ujwal-setlur using both of the button options you set worked for me :)

HoogsterInc:
Sadly it isn't working for me. It makes the Android back button (topBar) the color I want, but ios is greyed out instead of the color I set.

henrikra:
@guyca These two params should be easy to merge to one? Now I have to use two different params like @ujwal-setlur showed us

ujwal-setlur:
@henrikra Unfortunately, that seems to be the case :-(. For me, it works on Android, but on iOS, it is ignored.

ujwal-setlur:
Here is a patch that will address it:

```diff
--- a/node_modules/react-native-navigation/lib/ios/RNNBackButtonOptions.m
+++ b/node_modules/react-native-navigation/lib/ios/RNNBackButtonOptions.m
@@ -11,12 +11,46 @@ - (void)applyOn:(UIViewController *)viewController {
 		
[...]

PR #4437, opened 2018-12-12
Include missing backButton prop for android

tobias-leinss:
the generic buttonColor doesn't work for android. Backbutton color is not documented, see [#3454](https://github.com/wix/react-native-navigation/issues/3454)


guyca:
Thanks!

PR #4457, opened 2018-12-16
Fix setting bottomTabs.currentTabIndex on bottomTabs init

### LINKED ISSUE #4426: How to set initial selected tab index for bottom tabs?
### Issue Description

I want to select specific tab for initial selection. How to approach this in V2?

---
### Environment
* React Native Navigation version: 2.1.3
* React Native version: 0.57.7
* Platform(s) (iOS, Android, or both?): Both

### LINKED ISSUE #4433: [V2][iOS] currentTabIndex in common options has no effect
### Issue Description

After upgrade to 2.2.1 from ^2.0.2470 currentTabIndex has no effect in iOS app anymore.

in Android it works as expected but in iOS RNN alway starts tab based with first tab active instead of 3rd (index 2)

Set currentTabIndex option in common options like in the doc: 
https://wix.github.io/react-native-navigation/v2/#/docs/styling?id=common-options
``` ...},
  bottomTabs: {
   ...
    currentTabIndex: 2
   ...
```

**Tab based app with 4 tabs**
#### Expect behavior:
After app launch/reload have 3rd tab active.
#### Actual behavior:
After app launch/reload have 1st tab active.

---
### Environment
* React Native Navigation version: 2.2.1
* Device info (Simulator/Device? OS version? Debug/Release?): Simulator iPhone 6 OS 12.1.

yogevbd:
Apply `bottomTabs.currentTabIndex` on bottomTabs layout initialization in order to be able to initialize `bottomTabs` with specific selected index.

Fixes #4426
Fixes #4433


henrikra:
Okay so the problem was that when setting root with bottomTabs the options were never applied?

Also should you add test case to `RNNTabBarPresenterTest.m` to test if currentTabIndex has some value also? Not just empty options :)

PR #4460, opened 2018-12-16
Set stack root refactor

guyca:
This PR adds support for setting stack root with multiple children, which is to be expected as stack can be initialised with multiple children.

### DIFF
--- a/lib/src/Navigation.ts
+++ b/lib/src/Navigation.ts
@@ -1,3 +1,4 @@
+import { isArray } from 'lodash';
 import { NativeCommandsSender } from './adapters/NativeCommandsSender';
 import { NativeEventsReceiver } from './adapters/NativeEventsReceiver';
 import { UniqueIdProvider } from './adapters/UniqueIdProvider';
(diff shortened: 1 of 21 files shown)


henrikra:
Wow nice! Was this the real fix to the PR I started? :D https://github.com/wix/react-native-navigation/pull/4434

lib/src/Navigation.ts
```diff
@@ -146,8 +147,9 @@ export class NavigationRoot {
   /**
    * Sets new root component to stack.
    */
-  public setStackRoot(componentId: string, layout: Layout): Promise<any> {
-    return this.commands.setStackRoot(componentId, layout);
+  public setStackRoot(componentId: string, layout: Layout | Layout[]): Promise<any> {
+    const children: Layout[] = isArray(layout) ? layout : [layout];
```
henrikra: You dont need lodash. You can use  `Array.isArray` more info https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray

henrikra:
Currently testing version `2.2.5-snapshot.82` and looks like this feature is working on iOS and Android!

guyca:
@henrikra [Here's how we use it in the playground app](https://github.com/wix/react-native-navigation/blob/master/playground/src/screens/PushedScreen.js#L170)

PR #4461, opened 2018-12-17
Initialize splashscreen viewController anyway

### LINKED ISSUE #4459: [V2] iOS app stuck on launch screen
### Issue Description
First of all, thank you for your hard work on the plugin.

The application is stuck on launch screen if there is no image with a specific name for the launch screen.

It is enough to run the application without the required image of the launch screen.

https://github.com/wix/react-native-navigation/blob/62f6dbaa366855832b1b343acf1ff8ba1e135bfa/lib/ios/RNNSplashScreen.m#L69

If the image file is not found, the `viewController` is not initialized and not passed condition above in the `RNNSplashScreen.m` file. I thought the problem might be because of this, and when I added the code below before condition, everything worked as it should.

```
if (viewController == nil) {
		viewController = [[RNNSplashScreen alloc] init];
}
```
---
### Environment
* React Native Navigation version: 2.2.5
* React Native version: 0.57.8
* Device info (Simulator/Device? OS version? Debug/Release?): Both simulator and device, iOS 12.1, Debug and Release

dobrynia:
Resolve #4459.

We need to initialize the `viewController` and remove the `viewController != nil` condition.
If the image is not found, the window will remain invisible because of the `viewController` is not initialized and `makeKeyAndVisible` is not called.

Note this would be the right step also if the developer needs to implement the splash screen in a different way.

P.S. If this pull request is approved, I will implement a default splash screen in `viewController == nil ` section where the name of the application will be displayed.

Thanks.

### DIFF
--- a/lib/ios/RNNSplashScreen.m
+++ b/lib/ios/RNNSplashScreen.m
@@ -66,11 +66,13 @@
 		}
 	}
 	
-	if (viewController != nil) {
-		id<UIApplicationDelegate> appDelegate = [UIApplication sharedApplication].delegate;
-		appDelegate.window.rootViewController = viewController;
-		[appDelegate.window makeKeyAndVisible];
+	if (viewController == nil) {
+		viewController = [[RNNSplashScreen alloc] init];
 	}
+	
+	id<UIApplicationDelegate> appDelegate = [UIApplication sharedApplication].delegate;
+	appDelegate.window.rootViewController = viewController;
+	[appDelegate.window makeKeyAndVisible];
 }
 
 @end


dev6james:
@dobrynia can I ask why you closed this? I ran into the EXACT same issue today. 

Did you realize this was the wrong approach and that there was another solution?

dobrynia:
@dev6james I received no feedback on my pull request, so I closed it.

dev6james:
@dobrynia To be fair you did open it during the holiday season and closed it on new year's day. Might be worth another shot?

I feel like this is a good change. It frustrates me how common it has been in both v1 and v2 of this library for there to be no null condition handling other than to no-op.  Makes our debugging job so much harder.

Issue #4385, opened 2018-12-04
[V2] bottomTabs animate: true broken


dkimm96:
### Issue Description

Configuring bottomTabs to { visible: false, animate: true } via Navigation.mergeOptions changes the visibility but does not animate. TopBar is working fine, its just the bottomTabs that seems to be broken.

### Steps to Reproduce / Code Snippets / Screenshots

Navigation.mergeOptions(componentId, {
        topBar: {
          visible: true,
          animate: true,
        },
        bottomTabs: {
          visible: true,
          animate: true,
          drawBehind: true
        }
})

---
### Environment
* React Native Navigation version: 2.1.3
* React Native version: 0.57.4
* Platform(s) (iOS, Android, or both?): iOS
* Device info (Simulator/Device? OS version? Debug/Release?): Simulator (iPhone X) and Device (iPhone X) both. iOS 12.1, Debug scheme

michaelfreund:
I noticed a change in the behavior of the tab animation, which probably changed here: https://github.com/wix/react-native-navigation/compare/2.0.2595...2.0.2596

In version 2.0.2595, if you called the push command in a view with bottom tabs, the tab bar was pushed out of the screen together with the parent view. Starting with version 2.0.2596 it is only "hidden" and seems not to be connected with the parent view.

ADCrouch:
I'm also having the same issue, did you find a solution?

PR #4465, opened 2018-12-18
[V2][iOS] Fix bottomTabs’s animate option

wsliaw:
Resolve [#4385](https://github.com/wix/react-native-navigation/issues/4385)


henrikra:
Can you add example GIF showing this feature?

wsliaw:
@henrikra sorry i can't, but this is a port from v1

wsliaw:
@henrikra added

henrikra:
Nice! And can you also share what configuration you passed from JavaScript?

wsliaw:
aligned with the doc
```javascript
Navigation.mergeOptions(componentId, {
   bottomTabs: { visible: false, animate: true },
});
```

henrikra:
Ahaa so you are listening the scroll event in your code and when you are going down you will call this?

```ts
Navigation.mergeOptions(componentId, {
   bottomTabs: { visible: false, animate: true },
});
```

wsliaw:
yup :)
this pr is to fix the show/hide animation, this used to work in v1 but somehow not implemented in v2

henrikra:
You have to fix tests before this gets merged though :)

wsliaw:
can you give me some guidance?

lib/ios/RNNTabBarPresenter.m
```diff
@@ -55,7 +55,7 @@ - (void)mergeOptions:(RNNNavigationOptions *)newOptions currentOptions:(RNNNavig
 	}
 	
 	if (newOptions.bottomTabs.visible.hasValue) {
```
masarusanjp: Maybe you should also check `newOptions.bottomTabs.animate`.

lib/ios/RNNTabBarPresenter.m
```diff
@@ -16,7 +16,7 @@ - (void)applyOptions:(RNNNavigationOptions *)options {
 	[tabBarController rnn_setTabBarTranslucent:[options.bottomTabs.translucent getWithDefaultValue:NO]];
 	[tabBarController rnn_setTabBarHideShadow:[options.bottomTabs.hideShadow getWithDefaultValue:NO]];
 	[tabBarController rnn_setTabBarStyle:[RCTConvert UIBarStyle:[options.bottomTabs.barStyle getWithDefaultValue:@"default"]]];
-	[tabBarController rnn_setTabBarVisible:[options.bottomTabs.visible getWithDefaultValue:YES]];
+	[tabBarController rnn_setTabBarVisible:[options.bottomTabs.visible getWithDefaultValue:YES] animated:[options.bottomTabs.animate getWithDefaultValue:YES]];
```
masarusanjp: This is a breaking change because it does not animate in existing implementations. 
I think that it is better to default to NO.

lib/ios/RNNTabBarPresenter.m
wsliaw: updated

lib/ios/RNNTabBarPresenter.m
wsliaw: updated

lib/ios/UITabBarController+RNNOptions.m
```diff
@@ -31,8 +31,19 @@ - (void)rnn_setTabBarHideShadow:(BOOL)hideShadow {
 	self.tabBar.clipsToBounds = hideShadow;
 }
 
-- (void)rnn_setTabBarVisible:(BOOL)visible {
-	self.tabBar.hidden = !visible;
+- (void)rnn_setTabBarVisible:(BOOL)visible animated:(BOOL)animated {
```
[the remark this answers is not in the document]
wsliaw: updated, I made a little bit change, `tabBarFrame.origin.y` -> `tabBarFrame.size.height`

lib/ios/RNNTabBarPresenter.m
```diff
@@ -54,8 +54,8 @@ - (void)mergeOptions:(RNNNavigationOptions *)newOptions currentOptions:(RNNNavig
 		[tabBarController rnn_setTabBarHideShadow:newOptions.bottomTabs.hideShadow.get];
 	}
 	
-	if (newOptions.bottomTabs.visible.hasValue) {
-		[tabBarController rnn_setTabBarVisible:newOptions.bottomTabs.visible.get];
+	if (newOptions.bottomTabs.visible.hasValue && newOptions.bottomTabs.animate.hasValue) {
```
yogevbd: If `visible` has a value and `animate` doesn't, `rnn_setTabBarVisible` won't be called - This is a breaking change/bug.
I suggest we handle visibility change it two different ways:
1. Keep current behaviour as is - Visible has value and animate doesn't
2. Both visible and animate have values

Once you make this change we can move forward and approve this PR. Thanks

wsliaw: Changed :)
However, there's an Android test failed:
```Failed to install the following Android SDK packages as some licences have not been accepted.```
How should I fix this?

LRNZ09:
This would be really useful

lukebrandonfarrell:
@yogevbd Any status on merging this ?

kyle-ssg:
Doesn’t look like there’s any reason not to merge this?

wsliaw:
hi @yogevbd , is there anything I need to change or fix? thanks

PR #4497, opened 2018-12-24
Road to implicitAny part 4

henrikra:
## What is this?
This PR continues the work on making this project type safe with TypeScript. Goal is that the flag `noImplicitAny` will be set to `true`

## What this PR does?
- all files inside `adapter` folder is valid for `noImplicitAny`!
- `resolveAssetSource` is now imported from the current place where it should. Previously it was imported from legacy location. More info here: https://facebook.github.io/react-native/docs/image#resolveassetsource
- formatted many files
- add missing interfaces
- fix `onPressIn` prop on `TouchablePreview`

## What's next?
Next I am planning to make same treatment for `commands` folder and many it valid for `noImplicitAny` 🎉

### DIFF
--- a/lib/src/adapters/NativeCommandsSender.ts
+++ b/lib/src/adapters/NativeCommandsSender.ts
@@ -1,7 +1,24 @@
 import { NativeModules } from 'react-native';
 
+interface NativeCommandsModule {
+  setRoot(commandId: string, layout: { root: any, modals: any[], overlays: any[] }): Promise<any>;
+  setDefaultOptions(options: object): void;
+  mergeOptions(componentId: string, options: object): void;
+  push(commandId: string, onComponentId: string, layout: object): Promise<any>;
+  pop(commandId: string, componentId: string, options?: object): Promise<any>;
+  popTo(commandId: string, componentId: string, options?: object): Promise<any>;
+  popToRoot(commandId: string, componentId: string, options?: object): Promise<any>;
+  setStackRoot(commandId: string, onComponentId: string, layout: object): Promise<any>;
+  showModal(commandId: string, layout: object): Promise<any>;
+  dismissModal(commandId: string, componentId: string, options?: object): Promise<any>;
+  dismissAllModals(commandId: string, options?: object): Promise<any>;
+  showOverlay(commandId: string, layout: object): Promise<any>;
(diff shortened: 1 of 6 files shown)


henrikra:
@guyca Thx for merging my last PR so quickly! Here is next follow up :D

lib/src/adapters/NativeCommandsSender.ts
```diff
@@ -1,12 +1,29 @@
...
+
 export class NativeCommandsSender {
-  private readonly nativeCommandsModule;
+  private readonly nativeCommandsModule: NativeCommandsModule;
   constructor() {
     this.nativeCommandsModule = NativeModules.RNNBridgeModule;
   }
 
-  setRoot(commandId: string, layout: { root: any, modals: any[], overlays: any[] }) {
+  setRoot(commandId: string, layout: { root: any; modals: any[]; overlays: any[] }) {
```
guyca: What is changing `,` to `;` needed?

lib/src/adapters/NativeEventsReceiver.ts
```diff
@@ -7,22 +7,24 @@ import {
   SearchBarUpdatedEvent,
   SearchBarCancelPressedEvent,
   PreviewCompletedEvent,
-  ModalDismissedEvent
+  ModalDismissedEvent,
```
guyca: Can we avoid adding comma character where it's avoidable?

lib/src/adapters/NativeEventsReceiver.ts
```diff
@@ -31,39 +33,57 @@ export class NativeEventsReceiver {
     return this.emitter.addListener('RNN.AppLaunched', callback);
   }
 
-  public registerComponentDidAppearListener(callback: (event: ComponentDidAppearEvent) => void): EventSubscription {
+  public registerComponentDidAppearListener(
```
guyca: Lets return these method declarations to single line - I think it's readable enough in a single line.

lib/src/adapters/NativeCommandsSender.ts
henrikra: Yeah that is how types separate properties in TypeScript https://www.typescriptlang.org/docs/handbook/interfaces.html

henrikra:
I have now done the changes you requested :)

PR #4501, opened 2018-12-26
Fixed the bug that number will be passed to modalPresentationStyle when using OptionsModalPresentationStyle

masarusanjp:
## Problem

 using OptionsModalPresentationStyle would pass a number instead of string to modalPresentationStyle.

The reason is OptionsModalPresentationStyle is defined as a numeric enumeration.

## Fixed

I fixed OptionsModalPresentationStyle as a string enum


guyca:
Thanks @masarusanjp :)

masarusanjp:
Thanks for your rapid reviewing!

PR #4507, opened 2018-12-27
Fixed a bug that making navigationBar transparent makes it impossible…

masarusanjp:
### Issue

When making the navigation bar transparent with the following options (with noBorder: true), mergeOptions make it impossible to return to opaque.

```

{
      topBar: {
        background: {
          color: "transparent"
        },
        title: {
          text: "NavigationBar Color"
        },
        noBorder: true
      }
    };
}
```

Here is a demo code.

https://gist.github.com/masarusanjp/05727caf5a1dc9f693bfc884125dd8cb

### Screenshofts

before | after
--- | ---
before | after


henrikra:
Can you give more context what this solves? Show for example before and after screenshots?

masarusanjp:
@henrikra Sorry, I write it to issue's description. And put screenshots.

masarusanjp:
Sorry, I close this pr and make an issue about this.https://github.com/wix/react-native-navigation/issues/4511

My changes works in this situation, 
but I thought that changing translucent is not good as there may be side effects.

PR #4508, opened 2018-12-27
Road to noImplicitAny part 6 (FINAL part)

henrikra:
## What this PR is about?
Finally all strict rules are enables for TypeScript compiler! 🎉 Basically previously `noImplicitAny` was turned off but now it plus all strict rules are turned ON! 💋 

## Some highlights of the changes
- `Element` renamed internally to `SharedElement` to make more sense what is it
- `LayoutTreeParser`'s responsibility is now to do all `id` related stuff. Previously it was spread between `LayoutTreeCrawler` and `LayoutTreeParse` so it is more simple now
- clean up a lot of tests because they were testing duplicate stuff that was covered by other tests already
- removed all usages of `static get options` and replaces them with `static options`. This is how it is in the docs plus you cannot have `static get options(passProps)` because it is impossible to have getter with parameters.

## Tell me what do you think!
Now we have all benefits of strict TypeScript! After all what is the point of using TypeScript is strict types are not on! That is like doing normal JavaScript! :D

### DIFF
--- a/lib/src/commands/LayoutTreeParser.test.ts
+++ b/lib/src/commands/LayoutTreeParser.test.ts
@@ -1,14 +1,19 @@
-import * as  _ from 'lodash';
+import * as _ from 'lodash';
 import { LayoutTreeParser } from './LayoutTreeParser';
 import { LayoutType } from './LayoutType';
 import { Layout } from '../interfaces/Layout';
 import { OptionsSplitView } from '../interfaces/Options';
+import { UniqueIdProvider } from '../adapters/UniqueIdProvider';
+import { mock, instance, when, anything } from 'ts-mockito';
 
 describe('LayoutTreeParser', () => {
   let uut: LayoutTreeParser;
+  let mockedUniqueIdProvider: UniqueIdProvider;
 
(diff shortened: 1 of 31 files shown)


henrikra:
@guyca Here is the final part of Road to noImplicitAny! What do you think 💥 ?

henrikra:
@guyca Do you have time to check this PR on this week? :)

lib/src/commands/LayoutTreeParser.test.ts
```diff
@@ -174,7 +197,7 @@ const optionsSplitView: OptionsSplitView = {
   displayMode: 'auto',
   primaryEdge: 'leading',
   minWidth: 150,
-  maxWidth: 300,
```
guyca: This change made me so happy and gave me energy to continue reviewing this looooong pr 😅

henrikra: Glad you like it xD

guyca:
Merged! Thanks a lot for the great work @henrikra 🙇 💪

henrikra:
No worries!

PR #4515, opened 2018-12-28
Only remove the launch screen when the initial root view is ready

### LINKED ISSUE #3661: [V2] [iOS] Quick white screen (flash) after LaunchImage
### Issue Description

Between the LaunchImage and display of the first RNN screen, there is a small window where the screen is white. In Debug mode it takes longer; this is where you see the green "Downloading JavaScript bundle" bar up top. The statusbar also goes dark here even though I set it to light using `UIStatusBarStyle` in Info.plist.

Existing solutions exist that require editing `AppDelegate.m` code related to the `rootView`. However when using RNN this is impossible, since we have to replace everything with a custom bootstrap method:

```objc
[ReactNativeNavigation bootstrap:jsCodeLocation launchOptions:launchOptions];
```

Is there any way you guys on the native side could make it so that the launch image and UIStatusBarStyle are not changed until the first RNN screen has loaded? This is some critical functionality impacting the UX of every app using RNN.

n/a

---
### Environment
* React Native Navigation version: 2.0.2435
* React Native version: 0.55.4

danielgindi:
This will avoid the flashing white screen on app launch (Fixes #3661)

Currently RNN creates a mock of the Launch Screen (`RNNSplashImage`), and removes it on two opportunities:
1. `onBridgeWillReload` - when someone is doing a Command+R to reload - or *on the first load*. At this stage it is `null`ified, leaving the `UIWindow` empty.
2. `setRoot` - so we avoided the `= nil` in the bridge load, but still when the react code was loaded, executed and the user's code call `Navigation.setRoot(...`, the `UIWindow`'s `rootViewController` with the new root, which starts up as blank, white screen. This is because the first layout/render cycle did not happen yet.

This PR fixes this in two stages:
1. The very first `onBridgeWillReload`, when the `rootViewController` is still the `RNNSplashImage` - it will *not* set it to null anymore. It will still do this on subsequent reloads in debug mode.
2. The `setRoot` will wait for the first react ui event, and only then it will set the root view controller.
3. Fixes to `RNNSplashImage` to take default status bar style/hidden from the bundle's plist file.

### DIFF
--- a/lib/ios/RNNCommandsHandler.m
+++ b/lib/ios/RNNCommandsHandler.m
@@ -69,7 +69,7 @@ static NSString* const setDefaultOptions	= @"setDefaultOptions";
 	
 	__weak __typeof(self) _self = self;
 	
-	[vc.getCurrentLeaf waitForReactViewReady:YES waitForUIEvent:YES perform:^{
+	void (^complete)() = ^{
 		__typeof(self) __self = _self;
 		
 		// Sanity!
@@ -78,13 +78,32 @@ static NSString* const setDefaultOptions	= @"setDefaultOptions";
 		// Now we finally replace the launch screen
 		__self->_mainWindow.rootViewController = __self->_pendingRoot;
 		__self->_pendingRoot = nil;
-	}];
+		
+		// Let us know later that it's not the launch screen anymore
+		_isRootLaunchScreen = NO;
+		
+		// Let whoever know that we're ready
+		[_eventEmitter sendOnNavigationCommandCompletion:setRoot params:@{@"layout": layout}];
+		completion();
+	};
 	
-	// Let us know later that it's not the launch screen anymore
-	_isRootLaunchScreen = NO;
-	
-	[_eventEmitter sendOnNavigationCommandCompletion:setRoot params:@{@"layout": layout}];
-	completion();
+	if (vc.resolveOptions.animations.setRoot.waitForRender)
+	{
+		[vc.getCurrentLeaf waitForReactViewReady:YES waitForUIEvent:YES perform:^{
+			complete();
+		}];
+	}
+	else
+	{
+		_mainWindow.rootViewController = _pendingRoot;
+		
+		// Let us know later that it's not the launch screen anymore
+		_isRootLaunchScreen = NO;
+		
+		// Let whoever know that we're ready
+		[_eventEmitter sendOnNavigationCommandCompletion:setRoot params:@{@"layout": layout}];
+		completion();
+	}
 }
 
 - (void)removeRootIfNotLaunchScreen {
--- a/lib/ios/RNNCommandsHandler.m
+++ b/lib/ios/RNNCommandsHandler.m
@@ -69,27 +69,15 @@ static NSString* const setDefaultOptions	= @"setDefaultOptions";
 	
 	__weak __typeof(self) _self = self;
 	
-	[vc.getCurrentLeaf waitForReactViewRender:YES perform:^{
-		// Now the RCTContentDidAppearNotification is emitted, but it only means that the React Native view was added to the hierarchy, not that it started rendering!
+	[vc.getCurrentLeaf waitForReactViewReady:YES waitForUIEvent:YES perform:^{
+		__typeof(self) __self = _self;
 		
-		RCTUIManager *uiManager = [ReactNativeNavigation getBridge].uiManager;
+		// Sanity!
+		if (!__self || __self->_pendingRoot != vc) return;
 		
-		// UIManager methods could only be called on the UIManager queue
-		dispatch_async(uiManager.methodQueue, ^{
-			
-			// Make this the first call that happens with the UI on the view that was just added
-			[uiManager prependUIBlock:^(RCTUIManager *uiManager, NSDictionary<NSNumber *,UIView *> *viewRegistry) {
-				__typeof(self) __self = _self;
-				
-				// Sanity!
-				if (!__self || __self->_pendingRoot != vc) return;
-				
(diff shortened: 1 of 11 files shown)


lib/ios/RNNCommandsHandler.m
```diff
@@ -47,6 +49,7 @@ - (instancetype)initWithStore:(RNNStore*)store controllerFactory:(RNNControllerF
 	_stackManager = stackManager;
 	_overlayManager = overlayManager;
 	_mainWindow = mainWindow;
+    _isRootLaunchScreen = YES;
```
henrikra: Can you check if your tab settings are different than in this project? Many of the tabs are in correct 🤔

henrikra: Yeah I don't like tabs either but consistency is key 🔑  :)

henrikra:
Can you also give more context what this PR does and why does it work for you like this?

henrikra:
Also looks like e2e tests on iOS are failing https://jenkins-oss.wixpress.com/job/rnn-ios-pr/1041/console

danielgindi:
> Can you also give more context what this PR does and why does it work for you like this?

I've updated the PR's description

danielgindi:
> Also looks like e2e tests on iOS are failing https://jenkins-oss.wixpress.com/job/rnn-ios-pr/1041/console

This is a leftover from my first trials with this, when controlling the first root replacement from user code (custom render events etc.). This worked great except it requires polluting the code with an extra event, and that it will most probably create 100s of new issues by users who forget to make the call... Because I'm not that evil, I moved on and tried something more automatic :-)

I cleaned up the code and rebased.

henrikra:
Hmm the tests are still failing 🤔

danielgindi:
Okay we have a problem with `[ReactNativeNavigation getBridge]` crashing the tests `testSetRoot_setRootViewControllerOnMainWindow, testSetRoot_removeAllComponentsFromMainWindow` because it's trying to dispatch on `UIManager` queue, but it's `nil`. The bridge itself is `nil`.
Is there a way to make the bridge not `nil` in the tests?

danielgindi:
E2e tests are broken now, as they do not wait for first ui event, but for 'ready' event.
I've pushed a commit where the plain `waitForReady`, to see if maybe it will be sufficient in `Release` configurations. As on debug it may have extra white flashes due to the bridge doing different work.

I'll update here if the PR is okay and stable in release.

guyca:
@danielgindi 
Thanks for submitting the pr, I think this behavior should be configurable and opt-in though. We've implemented similar opt-in behavior on Android and are planning on implementing it for iOS soon.

usage will be as follows:
```js
Navigation.setRoot({
  bottomTabs: { // Or any other layout
    children: [
      {
        stack: {
          children: [
            {
              component: {
                name: 'someComp'
                options: {
                  topBar: {
                    title: {
                      component: {
                        name: 'titleComponent',
                        waitForRender: true
                      }
                    }
                  }
                }
              }
            }
          ]
        }
      }
    ],
    options: {
      animations: {
        setRoot: {
          waitForRender: true
        }
      }
    }
  }
});
```

Setting root like this will cause the ui to become visible only after both components are renders, avoiding the flicker you're experiencing.

danielgindi:
@guyca Yes I've just responded on the issue itself, that we should make this one configurable on `waitForRender` flag, which will help avoid the testing issues with `detox`.
If your guy is not working on it, then I'll probably get to it in the next few days.

danielgindi:
@guyca Now we are supporting `waitForRender` as per your example (as an opt in feature). Also rebased on master.

danielgindi:
@guyca I guess it did not make it to the milestone...

guyca:
RIght, want to take this? As discussed, we need to make it configurable on multiple levels:

1. current child of the root - meaning `waitForRender` will wait only for render of first (initial) tab and its current child.
2. TopBar title component
3. TopBar background component
4. TopBar button components

for components - they now include a `waitForRender` option.

danielgindi:
@guyca This PR is about the `setRoot` specifically, it does not affect other `waitForRender`s that are already implemented or not implemented yet.
And it's already configurable by the `waitForRender` in `setRoot`'s options

guyca:
@yogevbd Could we leverage the work done in this PR for the waitForRender feature we need? We also need to support waiting for both TopBar components and button components.

danielgindi:
@guyca You could, as I've modified the `waitForReactViewReady:perform:` method to include an additional flag specifying whether you want to wait for a RN UI event also, or just for the basic "ready" which just means "added to the superview".
So if/when you find out that a specific component has to wait for actual render and not just adding to view hierarchy, you switch it to `YES` and the magic happens. I think that there will always be different situations, like if you are waiting for root's view controller's render, then it's subcomponents will be rendered on the same frame, same UI event, so you will not really have to specify fine-grain it for other components in this situation.

Anyway this PR solves the main pain point in iOS with this lib, people are waiting a lot of time on that issue :-)

danielgindi:
@guyca @yogevbd I see that you are working on a different solution for this.
Please remember to take my changes to `RNNSplashScreen` - they are important to keep consistency with the real splash screen until the root is rendered.
Also - note that `onBridgeWillReload` will always cause the splash to be replaced with the not-ready-yet root view controller, unless you detect that and defer it until it's ready.

danielgindi:
I'm closing this, as you've done a lot of work there and already on master.
What's left is the splash screen itself which has missing status bar styling according to the `Info.plist`. There's a very simple PR here: https://github.com/wix/react-native-navigation/pull/4698 for that.
Hope you'll merge it soon. There's not really that much to review there.

PR #4516, opened 2018-12-28
Added missing types

masarusanjp:
Some styling options are missing on https://github.com/wix/react-native-navigation/blob/master/lib/src/interfaces/Options.ts.
So I got warnings when I use them from typescript.

### What to do

I added some types what existed on the Objective-C source and  possible to use.

### DIFF
--- a/lib/src/interfaces/Options.ts
+++ b/lib/src/interfaces/Options.ts
@@ -82,6 +82,13 @@ export enum OptionsModalPresentationStyle {
   none = 'none'
 }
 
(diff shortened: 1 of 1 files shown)


henrikra:
@guyca Is documentation being generated from TS types?

henrikra:
@masarusanjp Maybe you should mark that these are iOS only options?

masarusanjp:
Maybe yes. Options that I added are only for iOS. ( I could not find them on android codes)

PR #4543, opened 2019-01-06
Add dependency injection

henrikra:
## What this PR does?

This library heavily doing OOP in the codebase on TS side which is great! What was not so great is the dependency handling. As you can see from the `Navigation.ts` it created all the dependencies which isn't its job. Also that file wasn't testable because it created all the dependencies.

## What has changed

I added library called typedi which is meant for dependency injecting with TS and JS! In typedi you just mark a service class with `@Service(<serviceNameHere>)` and then you can inject to any other service using `@Inject(<serviceNameHere>)`! Very handy! These principles are coming from Java and other OOP languages.

## Opinions?

Tell me what do you guys think? I think the code functions now as it should when thinking in terms of OOP ☕️

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,4 @@
+{
+  "printWidth": 100,
+  "singleQuote": true
+}
--- a/lib/src/Navigation.ts
+++ b/lib/src/Navigation.ts
@@ -11,7 +11,7 @@ import { LayoutRoot, Layout } from './interfaces/Layout';
 import { Options } from './interfaces/Options';
 import { Service, Inject } from 'typedi';
 
-@Service()
+@Service(NavigationRoot.name)
 export class NavigationRoot {
   public readonly Element = SharedElement;
   public readonly TouchablePreview = TouchablePreview;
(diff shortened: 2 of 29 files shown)


henrikra:
@guyca @yogevbd What do you guys think about this? 🎉

guyca:
Hey @henrikra !
Thanks for the PR and sorry for not replying sooner.

My feelings towards this pr are a bit ambivalent. On the one hand one could argue about the benefits of DI, but on the other hand there's no denying the risk introduced by another dependency to this _library_, especially one which relies on stage 2 and experimental TS features.

Looking at the diff, the code base definitely looks cleaner, `Navigation.ts` especially, and future refactors might be easier and more contained thanks to this. Unfortunately I think the added risk is not worth it.

I actually don't think there's anything wrong with how RNN currently does DI - injecting dependencies in constructor is pretty straight forward.

lets keep this an open discussion, hopefully other members can void their opinions as well.

hannta:
Nice, `Navigation.ts` looks a lot of cleaner now.

Dependency injection might not be the most popular pattern on JS/TS projects, but that does not make it any worse.

henrikra:
@guyca How are your feelings now towards this? :)

PR #4545, opened 2019-01-07
Document options for Peek & Pop

karlsander:
This is the Information I needed to figure out how to implement Peek and Pop previews. Since the rest of the documentation is fairly concise I was wondering wether it would be appropriate to include "recipe" type things, like disabling scroll during preview, reacting to preview state in the previewed screen etc here or maybe in some other section?

### DIFF
--- a/docs/docs/animations.md
+++ b/docs/docs/animations.md
@@ -69,5 +69,5 @@ const Button = (
   </Navigation.TouchablePreview>
 );
 ```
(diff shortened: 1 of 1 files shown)


henrikra:
Could you update TypeScript types as well?

karlsander:
Looks like the types are already up to date and also commented very well. I also just noticed these options are already listed in the styling documentation (where I didn't think to look when implementing it). So maybe it's not a good idea to add this information in a third place, since one will likely go out of sync at some point.

I saw in the types that I was missing a few more options though.

PR #4561, opened 2019-01-10
Overlay with scroll view

guyca:
Fixes an issue with ScrollView inside an Overlay. While down events were propagated to the view, move events were consumed by the overlay preventing scroll components from functioning correctly.


guyca:
cc @eladgel

PR #4578, opened 2019-01-12
Adding hideNavBarOnFocusSearchBar option

sganti564:
This PR will allow the user to add a new property to the `topBar` options ` hideNavBarOnFocusSearchBar`
1. if the value of `hideNavBarOnFocusSearchBar` is false then the nav bar wont hide on focus the search bar.
2. if the value of `hideNavBarOnFocusSearchBar` is true or don't have this property mentioned in the `topBar` options then it will work as existing.


sganti564:
Did someone get a chance to look at this small PR? This is very important feature for one the components we build. Right now, user have to use multiple clicks (one for cancel and other for nav button) in order to accomplish a functionality. After this fix, they can use nav button directly and ignore the cancel. I want this to be merged asap. Thanks.

PR #4597, opened 2019-01-17
Tell prop's type with generic

henrikra:
With this you can type props same way you can already do with `push` :)
This PR continues the work done in https://github.com/wix/react-native-navigation/pull/4223


henrikra:
@guyca Are you able to check this? This should be pretty familar to you :)

ruscoder:
Hello! I found that `showModal` should be type safe too. I can prepare a tiny PR for it.

henrikra:
@ruscoder Go ahead! 🚀

PR #4608, opened 2019-01-20
Move animationType property to parent

robinvw1:
The docs is not correct. The `animationType` is not part of `left`, `right` or `center`, but should be listed under the `sideMenu` object.


RichardLindhout:
Does not have effect in root, sidemenu, left or right.

PR #4612, opened 2019-01-21
Refactor SplitView options and make them work with Tabs

zzorba:
The bottomTab settings were not being properly applied if a SplitView
controller was nested. Part of this seems to be that the 'options' for
SplitView were in their V1 form, i.e. they did not have access to the
full topBar/bottomTab/etc.

I refactored the options to be in their own sub object. NOTE: the
settings for `splitView` are not being applied currently -- they weren't
before and they still are not being respected now.

With access to the full options now, I was able to change
RNNSplitViewController to properly override
willMoveToParentViewController to get the bottomTab icon to properly
show up.

I would love to get the actual `splitView` settings to properly work, but I need some guidance on how to do that in the new Presenter framework.  Since this seems to be no worse off than before, I figure it is still worth merging in the meantime.

### DIFF
--- a/docs/docs/layout-types.md
+++ b/docs/docs/layout-types.md
@@ -200,10 +200,12 @@ const splitView = {
     // All layout types accepted supported by device, eg. `stack`
   },
   options: {
-    displayMode: 'auto', // Master view display mode: `auto`, `visible`, `hidden` and `overlay`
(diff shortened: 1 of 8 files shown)


zzorba:
Unsure why the android build failed, as this is an iOS only feature.

PR #4637, opened 2019-01-25
Pull from the root TabBarController height

### LINKED ISSUE #4190: [V2] topBarHeight constant always 0
### Issue Description

It seems that when accessing the `topBarHeight` from `Navigation.constants()` that the value is always set to `0`. 

I'm not certain if there's another way to access the height of the navigation bar.

```js
const constants = await Navigation.constants();
console.table(constants);
```

---
### Environment
* React Native Navigation version: 2.0.2594
* Device info (Simulator/Device? OS version? Debug/Release?): Simulator / iOS 12 / Debug

paubins:
Pulls the height from the root UITabBarController.

Fixes #4190

### DIFF
--- a/lib/ios/Constants.m
+++ b/lib/ios/Constants.m
@@ -15,7 +15,7 @@
 }
 
 + (CGFloat)bottomTabsHeight {
-	return UIApplication.sharedApplication.delegate.window.rootViewController.tabBarController.tabBar.frame.size.height;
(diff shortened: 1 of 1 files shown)


LRNZ09:
As pointed by @MrtnvM on the commit page https://github.com/wix/react-native-navigation/commit/8cee74533489e1877ffea95c1622bd4ab6fb33ff this way of getting tab bar height leads to a crash and should be fixed.

PR #4647, opened 2019-01-27
Fix splitView options

zzorba:
This change brings the SplitView controller in line with the rest of the V2 refactor.  Introduces a new Presenter and an intermediate UISplitViewController+Options to apply changes.

I tested all 4 options, including with mergeOptions, and it seems to behave as expected.


zzorba:
So this seems to work -- though there appears to be something off with teh combination of split view + tab (but it might be a preexisting issue).

image

It seems to be related to the spacing for the bottom tab bar if I weld off the following bit of code in UIViewController+RNNOptions.h
```
- (void)rnn_setDrawBehindTabBar:(BOOL)drawBehindTabBar {
	if (drawBehindTabBar) {
		[self setExtendedLayoutIncludesOpaqueBars:YES];
		self.edgesForExtendedLayout |= UIRectEdgeBottom;
	} else {
		self.edgesForExtendedLayout &= ~UIRectEdgeBottom;
	}
}
```
to always execute the drawBehindTabBar=true path, then the extra spacing goes away.  I think what should happen is the childViewControllers need to act as if they are NOT in a tab bar (from a height calculation POV), but I don't know the best way to achieve that.

exentrich:
Great feature! Thank very much!
I found several problems:
- `statusBar.style` option is ignored, global too
- Can we customize border between master and detail? Show/hide is enough.
- App is crashed on iPad when it started or moved to "split view" - when two apps showed simultaneously (for example 80% of screen is Safari, 20% my app)

PR #4658, opened 2019-01-30
[v2] [iOS] Feature - support for topBar customization (Android-like)

StasDoskalenko:
This PR gives an ability to set topBar `borderColor` and `borderHeight` for iOS applications like Android version of react-native-navigation does.
image


StasDoskalenko:
Updated with master to fix tests

PR #4681, opened 2019-02-03
Improve navBarBlur and navBarTranslucent documentation

kyle-ssg:
Although it was quite obvious after, it wasn't mentioned that translucent: true was required for navBarBlur to work.

Also, I think adding required options to the JS snippets make this easier to follow.


guyca:
Thanks! ❤️

PR #4683, opened 2019-02-04
[v2][iOS] Feature - support mergeOptions on bottomTab iOS

magrinj:
Actually when you try to do a mergeOptions on bottomTab on iOS nothing happen, this PR fix this, now you can use mergeOptions to update the text and icon of a tab.


yogevbd:
Solved here: https://github.com/wix/react-native-navigation/pull/4683
Thanks!

PR #4699, opened 2019-02-07
Add Icon insets support for buttons

yogevbd:
```js
{
  topBar: {
    rightButtons: [{
      id: '',
      icon: 'require()'
      iconInsets: {
        top: 0,
        left: 8,
        bottom: 0,
        right: 8
      }
    }]
  }
}
```

### DIFF
--- a/lib/ios/RNNButtonOptions.h
+++ b/lib/ios/RNNButtonOptions.h
@@ -1,4 +1,5 @@
 #import "RNNOptions.h"
+#import "RNNInsetsOptions.h"
 
 @interface RNNButtonOptions : RNNOptions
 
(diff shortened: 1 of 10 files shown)


calendee:
@yogevbd Can you describe what the `iconInsets` property does? I've applied it to my buttons hoping to solve issue #5000, but it has no effect that I can see.  What is an "inset"?

yomybaby:
This PR make https://github.com/wix/react-native-navigation/issues/4708 issue.

Scrolling has higher priority than below function.
```
[self.target performSelector:self.action
						  withObject:self
						  afterDelay:0];
```
I don't know how to fix this exactly. But previous code is work properly. :)

PR #4707, opened 2019-02-07
Fix launch image matching for iPhone XR/XS Max portrait

oblador:
This PR fixes so that the splash screen properly resolves to the appropriate launch image for the new iPhone screen sizes released last year.


henrikra:
Is there anyway to make more robust solution because when Apple release new iPhone with new screen size then we again see only a black screen. I think if we can't find matching splash image then we just have black splash and start the app normally. Because now if we don't have matching splash screen then the app just stays there and does nothing xD

PR #4713, opened 2019-02-09
Remove tests that don't really test RNN

henrikra:
## Background
I started to investigate why in `NativeEventsReceiver.ts` this `try catch` block was needed. And the reason was that the tests in `integration` using `registerComponent` failed because `NativeModule` is always null.

## Why tests were removed?
I check the tests out and they did not seem to test RNN functionality at all. For example in Remx test it just did some component testing and store testing which is not part of this library. Also some of the tests tested that `hoist-non-react-statics` library works. But that is not needed either since that library has tests https://github.com/mridgway/hoist-non-react-statics/blob/master/tests/unit/index.js

## Conclusion
If you really really want to make test that RNN work for example with Redux then there should be E2E test for that. Otherwise these tests are useless and therefore can be removed :)


henrikra:
@guyca I did small pr for you guys :)

guyca:
Hey @henrikra! 👋 
These tests are integration tests, they ensure RNN works as expected in the same environment with other third party libraries.
Their purpose is to improve our confidence and ensure RNN is compatible with these popular libraries, I think we better keep them.

PR #4715, opened 2019-02-10
[v2] [iOS] Back button: Allow for both icon and title, use topbar title attributes

kyle-ssg:
I'm not an Objective C developer, but this solves an issue I'm currently seeing.

1 - You can't have custom back icon with title, in this scenario it will only render the icon

2- Setting the topbar font has no affect on the back button title

After: Back has both custom font and icon, back button title renders next to icon

Also, tiny nit but this also removes some trailing spaces.

### DIFF
--- a/lib/ios/UINavigationController+RNNOptions.m
+++ b/lib/ios/UINavigationController+RNNOptions.m
@@ -106,8 +106,8 @@ const NSInteger BLUR_TOPBAR_TAG = 78264802;
 		? [[icon withTintColor:color] imageWithRenderingMode:UIImageRenderingModeAlwaysOriginal]
 		: icon;
 
-		[self.navigationBar setBackIndicatorImage:[UIImage new]];
-		[self.navigationBar setBackIndicatorTransitionMaskImage:[UIImage new]];
+		[self.navigationBar setBackIndicatorImage:backImage];
+		[self.navigationBar setBackIndicatorTransitionMaskImage:backImage];
 	}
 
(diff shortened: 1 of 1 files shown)


guyca:
Seems like one unit test is failing
```
19:08:38 /Users/jenkins/.jenkins/workspace/rnn-ios-pr/lib/ios/ReactNativeNavigationTests/UINavigationController+RNNOptionsTest.m:38: error: -[UINavigationController_RNNOptionsTest testSetBackButtonIcon_withColor_shouldSetIcon] : ((icon) equal to (viewController.navigationItem.backBarButtonItem.image)) failed: ("<5023d723 eb7f0000>") is not equal to ("<00000000 00000000>")
```

kyle-ssg:
I'll take a look, I imagine it'll be quite straight forward to update.

It seems like the android side isn't able to have both an icon and text too for the back button, was this done by design @guyca ?

guyca:
Currently thats not supported as it's not a common Android pattern.

kyle-ssg:
Duh of course, because of the navbar title positioning :D, think for iOS through the above is definitely a pattern so this change has some value. OK, will fix the tests up.

guyca:
@kyle-ssg Sounds good, if you run into issues with the tests let us know.

kyle-ssg:
Snowed under atm, aiming to do this on the weekend. Sidenote, this PR also fixes custom back button positions - it's further right than normal before this change.

lukebrandonfarrell:
Any update on this @guyca @kyle-ssg ? 

Does it just need to be rebased and tests fixed on Android? if so I can give it a shot.

Issue #4628, opened 2019-01-23
[FEAT] Statically switching tabs


MaxMaes:
### Issue Description

I have a screen with 4 tabs, but upon selecting the last tab I need to hide the tabbar so the screen kind of appears as an overlay. When I close that screen, I need to navigate back to the previously selected tab.

In my current setup, I store the previously selected screen and want to switch the tab using mergeOptions. But when I use mergeOptions, the tabbar is still hidden on the other screen. So, ok,  can make it visible for that screen again when I also set it to visible in the mergeOptions, which works. But when I go back to my 'overlay tab', the tabbar doesn't hide.

I have a specific static options setup for each of my tab screens, so it is very tedious to be working with mergeOptions for switching tabs.

Would it be possible to add a top level API for selecting a tab, eg Navigation.selectTabIndex and/or Navigation.selectTabId where the statically declared options of the screen are applied upon opening said screen?

### Environment
* React Native Navigation version: 2.7.1
* React Native version: 0.57.8
* Platform(s) (iOS, Android, or both?): Both
* Device info (Simulator/Device? OS version? Debug/Release?): All

MaxMaes:
This has been added by @RemyLivewall, not stale?

guyca:
Closing, see https://github.com/wix/react-native-navigation/pull/4739#issuecomment-468293356

PR #4739, opened 2019-02-15
[V2] Add support for switching tabs without using merge options

RemyLivewall:
This pull request adds support for using a static method on Navigation to change the current bottom tabs index. This might be a fix for issue #4628

It takes into account situations where a user might not have a tab bar and sends a warning accordingly. If the index is out of bounds iOS does nothing and a warning will be logged to logcat on Android.

I updated the commands test to make sure it still passes, but I am having trouble adding tests to test the new functionality in e2e and ios/android.


guyca:
Hey @RemyLivewall 
We want to keep the api as lean as possible. Therefore we elected to expose all methods which mutate the navigation state through `mergeOptions`.

To switch between tabs statically, give your bottomTabs a predefined id, and invoke mergeOptions using that id.

```js
bottomTabs: {
  id: 'MyBottomTabs',
  children: [ ... ]
}


Navigation.mergeOptions('MyBottomTabs', {
  bottomTabs: {
    currentTabIndex: 1
  }
});
```

RemyLivewall:
Thank you for your response. This is understandable, although it is unfortunate that using merge options doesn't respect static options declared in the target screen. I will create an for the issues we have had with using merge options.

PR #4740, opened 2019-02-15
Prevent unwanted app destroy on Android

umerbaig:
Hi,

I was running into an annoying issue while trying to ask user permissions in my home screen on android (I have a map on the Deals maps screen so I need to ask the user for his location). I am using react-native-maps and since it requires location permission so when the user accepts or decline the permission, user will end up with the error `Tried to use permissions API while not attached to an Activity`.

So after some investigations I realize that when the native permissions pop up shows up, the app is put in pause and the `currentActivity` is null. I forgot to mention that I have a Login flow first and at the end I call `startSingleScreenApp` so the first activity is destroyed and replaced by the new one. The problem here is that the `onDestroy` callback is called after the `onPause` of the new screen. So when is get trough `destroyJsIfNeeded` function, `currentActivity` is null and the app is destroyed, causing the `Tried to use permissions API while not attached to an Activity`.

I apologize if it is not very clear but it is pretty hard to explain !
I don't have a global view of the project so please tell me if I am doing something wrong

### DIFF
--- a/android/app/src/main/java/com/reactnativenavigation/controllers/NavigationActivity.java
+++ b/android/app/src/main/java/com/reactnativenavigation/controllers/NavigationActivity.java
@@ -55,6 +55,7 @@ public class NavigationActivity extends AppCompatActivity implements DefaultHard
      * Along with that, we should handle commands from the bridge using onNewIntent
      */
     static NavigationActivity currentActivity;
+    static int numberOfActivities = 0;
     private static Promise startAppPromise;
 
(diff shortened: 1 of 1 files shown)


jamesgestier:
Hi @guyca, is there any chance of getting a release of v1 with this fix included? This issue is crashing my Android app and the code changes fix the crash so it would be awesome if it could be released. Thanks, James.

YaliWang0523:
unnamed

Issue #4485, opened 2018-12-20
[V2] Toptabs are not showing up, regression in newest version of RNN


the-friyia:
### Issue Description

There has been a regression in Top Tabs Navigation for React-Native-Navigation. After version  2.0.2459 topTabs will no longer appear. 

I am currently using the code below with version 2.2.5, however, that code allows me to swipe right and left and the app will transition screens BUT tabs themselves will not appear. 

If I downgrade to 2.0.2459 again, the top tabs once again appear.

### Steps to Reproduce / Code Snippets / Screenshots

```
        stack: {
          id: 'mainStack',
          children: [
            {
              topTabs: {
                children: [
                  {
                    component: {
                      name: 'Screen1',
                      options: {
                        topTab: {
                          title: 'Tab 1',
                        },
                      },
                    },
                  },
                  {
                    component: {
                      name: 'Screen2',
                      options: {
                        topTab: {
                          title: 'Tab 2',
                        },
                      },
                    },
                  },
                ],
              },
            },

```  

### Environment
* React Native Navigation version: 2.2.5
* React Native version: 0.55.0
* Platform(s) (iOS, Android, or both?): Android
* Device info (Simulator/Device? OS version? Debug/Release?): Device

Ibrahim-GHub:
I am also facing this issue with "react-native-navigation": "^2.3.0"

`Navigation.push('insp-req', {
    stack: {
        id: 'insp-req-info-stack',
        children: [{
            topTabs: {
                currentTabIndex: 0,
                currentTabId: 'insp-req-info',
                children: [{
                    component: {
                        id: 'insp-req-info',
                        name: 'InspectionRequestDocumentInfo',
                        options: {
                            topTab: {
                                title: 'Info'
                            },
                            topBar: {
                                title: {
                                    text: 'Details'
                                }
                            }
                        },
                        passProps: {
                            id: id,
                            projectId: projectId
                        }
                    }
                }, {
                    component: {
                        id: 'insp-req-workflow',
                        name: 'InspectionRequestDocumentWorkflow',
                        options: {
                            topTab: {
                                title: 'Workflow'
                            },
                            topBar: {
                                title: {
                                    text: 'Details'
                                }
                            }
                        },
                        passProps: {
                            id: id,
                            projectId: projectId
                        }
                    }
                }, {
                    component: {
                        id: 'insp-req-attachments',
                        name: 'InspectionRequestDocumentAttachments',
                        options: {
                            topTab: {
[...]

PR #4763, opened 2019-02-19
Add support for tvOS 

hufkens:
## What this PR does?
We want to use react-native-navigation for our TV Apps. Other navigation solutions don't work that well and we want the native feel. Also react-native is dropping TabBarIOS and NavigatorIOS from the core. So we need a good navigation library that we can use on al platforms.

## What has changed
I added a target `ReactNativeNavigation-tvOS` to that builds to the tvOS SDK. This means I had to exclude not supported code using compiler flags. 

## Opinions?
Tell me what do you guys think? It's a start and a work in progress, and I hope that you are interested to give some feedback. We are intending to use it in our apps (iOS, Android, tvOS, Android TV).

## To Do
I need to find a way to hook the Menu gesture handler, so that it will pop() the Screen until the last one and then exit the app.

### DIFF
--- a/lib/ios/Constants.m
+++ b/lib/ios/Constants.m
@@ -10,9 +10,15 @@
 	return UIApplication.sharedApplication.delegate.window.rootViewController.navigationController.navigationBar.frame.size.height;
 }
 
+
 + (CGFloat)statusBarHeight {
+	#if TARGET_OS_TV
+	return 0;
+	#else
 	return [UIApplication sharedApplication].statusBarFrame.size.height;
+	#endif
 }
+
 
 + (CGFloat)bottomTabsHeight {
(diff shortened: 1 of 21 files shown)


guyca:
Hey @hufkens, Looks very promising! You're welcome to ping me on discord if you need anything on our end 👍

hufkens:
@guyca We were looking into using react-native-navigation for our TV apps. I managed to get it functionally working. But we stopped working on it because of this issue https://github.com/wix/react-native-navigation/issues/4485

We also need the TopTabs to be working but the seem to be broken now.

There are also still 2 open issue:
- Correctly handle the "menu" button on tvOS. Pressing the menu should go back in the UINavigationController and/or close a Modal.
- The Splash screen logic now works with the xib, but there are no xibs on Apple TV. That code is commented for now and there is no splash screen shown.

I have an account on discord but don't use it that much.

guyca:
Thanks for the update, perhaps someone will pick this up in the future 👍

PR #4765, opened 2019-02-20
Use the same dispatch group for rendering layout tree

yogevbd:


guyca:
@yogevbd Is this relevant? What does this PR do?

PR #4766, opened 2019-02-20
Fixes command completion event commandId

yogevbd:
Until now `registerCommandCompletedListener` returned `commandName` as the `commandId`. it will now return both `commandName` and the correct `commandId`.


guyca:
@yogevbd Seems like this commit breaks an iOS specific e2e test, can you please look into it?

guyca:
@yogevbd Lets get this merged asap

PR #4778, opened 2019-02-23
Prevent keyboard from hiding when overlay is shown

tomhicks:
If you use overlays to show in-app notifications
it would closes the keyboard every time an overlay is
opened.


guyca:
@yogevbd Think this be tested with Detox?

ItsNoHax:
Guys, this code change is causing weird effects now regarding TextInput focusing. It's breaking behaviour like autofocus prop or focusing the field on componentDidMount

cspicuzza:
`<TextInput autoFocus>` is being ignored on showModal as well.  I think it applies to this issue?

Issue #4117, opened 2018-10-10
Can't dismissModal without animation


smokokoi:
dissmissModal on iOS is always animated. I tried to dismiss it without animation using the following command:

```js
Navigation.dismissModal(componentID, {
        animations: {
          dismissModal: {
            enable: false,
          }
        }
    });
```

yogevbd:
Replacing `enable` with `enabled` should solve this on iOS
```js
Navigation.dismissModal(componentID, {
        animations: {
          dismissModal: {
            enabled: false,
          }
        }
    });
```

smokokoi:
@yogevbd @guyca - it seems I can't dismiss a modal without animation on both platforms.

PR #4780, opened 2019-02-24
Add  enabled option to interface `OptionsAnimationProperties`

taichi-jp:
There is `enabled` option in `OptionsAnimationProperties` , but it is not defined in the interface.
Related to https://github.com/wix/react-native-navigation/issues/4117 .


guyca:
Thanks!

PR #4877, opened 2019-03-18
[V2] Tweak app lifecycle docs

ericketts:
I've done two things here: first I have made the capitalisation of `UI` consistent; and a much greater change, I've expanded upon the android JS context vs android UI context documentation, including adding info for how to detect and handle a divergence of the lifecycle of the two contexts.

If this is too verbose please let me know, I tried to balance a reasonable amount of text with leaving as little room for ambiguity as possible. 

thanks for taking the time to review, and again for putting up these docs so quickly, I think it really will help folks take full advantage of this very comprehensive navigation solution!


docs/docs/app-launch.md
```diff
@@ -1,28 +1,28 @@
...
+!> Register the listener with `registerAppLaunchedListener` as soon as possible - it should be one of the first lines in your `index.js` file.
 If you're observing a "white screen" or a hanging splash screen after relaunching your app, it probably means `Navigation.setRoot` isn't called once the app has launched. Perhaps the listener was registered too late.
 
 ## The difference between the platforms
 When your app is launched, RN makes sure Js context is running. Js context is what enables you to execute JavaScript code.
 There are a few differences between iOS and Android in this regard
 
 ### iOS
-Each time the app moves the the background, Js context is destroyed as the app's process is destroyed.
+Whenever the app is put into the background, the app process could potentially be destroyed by the system. If this destruction of the app takes place, the Js context will be destroyed along with the app process.
```
guyca: AFAIK, on iOS, the process is always killed by the system once the app moves to the background.
This means that, on iOS, each time the app moves to foreground, the bundle is parsed and executed.

guyca:
Wonderful additions! Left one small comment :)

docs/docs/app-launch.md
ericketts: I think that it [certainly can be killed in the background](https://developer.apple.com/documentation/uikit/core_app/managing_your_app_s_life_cycle/preparing_your_app_to_run_in_the_background), but that this only happens after a period of time, or if memory is needed by a different, resource hungry app; by default, view controllers and app delegate should stay instantiated between backgrounding and foregrounding. At least this is the case for our RN app using RNN right now :)

guyca: You're right 👍

PR #4909, opened 2019-03-25
Add flow type definitions

alpha0010:
Generates flow type definitions from the typescript. Mostly automatic, however lib/src/types.js.flow (and line 17 of scripts/gen-flow.js) may need occasional updates for imported external types.


retyui:
Flow typed already have types 
https://github.com/flow-typed/flow-typed/pull/3225

lib/src/types.js.flow
```diff
@@ -0,0 +1,93 @@
...
+declare export class TouchableHighlight extends React$Component<any, any> {}
+
+declare export class TouchableNativeFeedback extends React$Component<
+  any,
+  any
+> {}
+
+declare export class TouchableOpacity extends React$Component<any, any> {}
+
+declare export class TouchableWithoutFeedback extends React$Component<
```
retyui: `Touchable*` types from react-native

henrikra:
@alpha0010 Cmon Facebook is going to kill Flow since they are already refactoring Yarn and Jest to TypeScript :D So if maker of Flow is them selves using TypeScript you should not either ;)

retyui:
@henrikra 
>  Facebook is going to kill Flow

You should read about jest and yarn why it will be refactoring to TS! Before you express your disrespect to the Flow.js community.

And more complex answer flow team about that news: https://medium.com/flow-type/what-the-flow-team-has-been-up-to-54239c62004f

guyca:
Hey @retyui @alpha0010 - I'm not familiar with flow, do we need these declarations in RNN even though they are available in flow-typed?

retyui:
@guyca 

The biggest problem is that typings are targeting specified range of flow versions and having them directly in the repository kinda locks them down to that range and might cause problems for people using versions from outside the range.

From my point of view, if you are not going to support flow types, then it’s better to let the community support

alpha0010:
@retyui Last I checked, flow typed did not have defs for this library; looks like I was off by a few days.

Having types in this library has the advantage of them always being up to date with the api of this library.  I have not changed flow versions very frequently, so I cannot speak to how likely these generated definitions are to break on different flow binaries.

If the types are bundled in this library, they will "just work" when a user installs the library (no need for the extra flow typed step).  However, if flow versions are significantly incompatible you would end up with support requests...

lib/src/types.js.flow
alpha0010: Yes, though it does not seem possible for flow to import/export definitions between libraries, hence the filler.

retyui:
@alpha0010  
> Having types in this library have the advantage of them always being up to date with the api of this library 

Your script does not add types for the API library itself!

I see no one method of `Navigation` class (`setRoot()` \ `push()` \ ... only garbage types `any`

alpha0010:
@retyui
> Your script does not add types for the API library itself!

Hmm.  Did you `npm run build-flow`?  Inside of `lib/dist/index.js.flow` I have:
```js
// @flow
import { NavigationRoot } from './Navigation';
declare export var Navigation: NavigationRoot;
declare export * from './adapters/Constants'
declare export * from './interfaces/ComponentEvents'
declare export * from './interfaces/Events'
declare export * from './interfaces/EventSubscription'
declare export * from './interfaces/Layout'
declare export * from './interfaces/Options'
```
And `lib/dist/Navigation.js.flow`:
```js
// @flow

import type { ComponentProvider } from './types';
import { EventsRegistry } from './events/EventsRegistry';

import { SharedElement } from './adapters/SharedElement';
import { TouchablePreview } from './adapters/TouchablePreview';
import type { LayoutRoot, Layout } from './interfaces/Layout';
import type { Options } from './interfaces/Options';
declare export class NavigationRoot {
  Element: typeof SharedElement;
  TouchablePreview: typeof TouchablePreview;
  constructor(): this;

  /**
   * Every navigation component in your app must be registered with a unique name.
   * The component itself is a traditional React component extending React.Component.
   */
  registerComponent(
    componentName: string | number,
    componentProvider: ComponentProvider,
    concreteComponentProvider?: ComponentProvider
  ): ComponentProvider;

[...]
```

retyui:
@alpha0010 

> Hmm. Did you npm run build-flow? Inside of lib/dist/index.js.flow I have:

so why types are not stored in the repository?

> if flow versions are significantly incompatible you would end up with support requests...

How does it work for you? I made some changes in .flow files and then before the release your script owerwrite my changes?

alpha0010:
@retyui The goal was to have (near-) zero maintenance flow types.  So, these are not committed, just like the typescript type definition files are not committed.  On release, the type files (both typescript and flow) are created from the source code of this project.  That way, types are always exactly synchronized with the source.

With the exception of the .flow file defining external types, the files are not intended to be manually edited.  If they contain weak/poor types for certain functions, they should be updated in the source.  This process insures the types are validated to be correct (by the typescript compiler at build/check time) and that improvements to types benefit both the typescript and flow worlds.

If the generated flow files are incompatible with some flow binaries but not others, well I am uncertain what the best course of action would be.  Again, I have only rarely changed flow, so it is not an issue I have any experience with.

retyui:
And if they are not changeable, then any errors during critical changes in Flow.js put us in a dead end. since we have to wait for the edits to be introduced during the `flowgen`

I can not trust the generated types

The best solution in my opinion is to store the types during `flow-typed` covered tests

alpha0010:
In any case, it is best to have a single source of truth; closing this pr since rnn is already handled by flow-typed.

PR #4929, opened 2019-03-30
Revert "Prevent retaining button component in componentRegistry (#4888)"

jfrolich:
This reverts commit 0186b1ac36e919fb6b2a796677db1905b48aec7e.

This commit introduces buggy behavior with some components disappearing at random. It's pretty hard to create an easily reproducable small example for a native project, but I included a video of the behavior in my app.

To clarify, the title that disappeared is a custom react-component in the top bar.

correct behavior (this pull request): https://youtu.be/wCPNokoa-FA
buggy behavior: https://youtu.be/uQGfdX5XTcU

### DIFF
--- a/lib/ios/RNNReactComponentRegistry.m
+++ b/lib/ios/RNNReactComponentRegistry.m
@@ -2,7 +2,7 @@
 
 @interface RNNReactComponentRegistry () {
 	id<RNNRootViewCreator> _creator;
-	NSMapTable* _componentStore;
+	NSMutableDictionary* _componentStore;
 }
 
 @end
@@ -12,7 +12,7 @@
 - (instancetype)initWithCreator:(id<RNNRootViewCreator>)creator {
 	self = [super init];
 	_creator = creator;
-	_componentStore = [NSMapTable strongToWeakObjectsMapTable];
+	_componentStore = [NSMutableDictionary new];
 	return self;
 }
 
@@ -31,7 +31,7 @@
 }
 
 - (NSMutableDictionary *)componentsForParentId:(NSString *)parentComponentId {
-	if (![_componentStore objectForKey:parentComponentId]) {
+	if (!_componentStore[parentComponentId]) {
 		[_componentStore setObject:[NSMutableDictionary new] forKey:parentComponentId];;
 	}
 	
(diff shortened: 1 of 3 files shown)


jfrolich:
BTW I don't have enough objective C knowledge to know why this happens :) Just triangulated the issue to this commit, by stepping through each commit to see where the issue was introduced.

guyca:
Hey @jfrolich 
We weren't able to reproduce the issue on both playground and Wix app. I suggest you PR a failing detox test . I think a simple test which pushes and pops a screen will fail at some point. You can reproduce the problematic scenario and wrap the test (`it` statement) in a for loop so it will run multiple times, eventually it will crash.

jfrolich:
Hi, would it be possible to revert this commit in master as it obviously has problems, and make sure we only commit something back when the issues are resolved.

jfrolich:
Did you test it with a custom react component in the topbar?

jfrolich:
I provided a reproducible example in the playground. You can reproduce it by unreverting the commit and then: 

1. Go to the options tab
2. Click on "Set ReactTitleView"
3. Click on "Push ReactTitleView"
4. Click back
5. Potentially repeat step 3-4

guyca:
@jfrolich I wasn't able to reproduce with your changes.
I suggest you add some logs to the lifecycle methods of your title component - check if it gets unmounted when pushing. Perhaps it's still mounted but for some reason props change and the `props.title` is lost

jfrolich:
Did you unrevert the commit (you need to do that otherwise you can't reproduce in this branch because it has already be fixed).

I could record the actions in a video, but with the steps you should be able to reproduce it!

jfrolich:
> Perhaps it's still mounted but for some reason props change and the props.title is lost
Yep that is probably the case, because without props it still works. BTW if you look closely there is also some flickering low framer ate going on.

jfrolich:
Anyway the burden of testing it properly so it doesn't break functionality is IMO on the burden of the person who committed this change. So I think reverting this (as it's introducing unwanted bugs) is the best course of action, and then only re-introduce this PR when it is properly tested.

guyca:
@jfrolich My apologies, I just realised my working directory wasn't clean when I tried to reproduce the issue. I sorted out my local clone and indeed it reproduces.

@yogevbd - Seems like custom components are unmounted after being displayed for the second time.

guyca:
@jfrolich I'm closing the PR but we'll address this asap 👍

jfrolich:
Thanks!

Issue #4716, opened 2019-02-11
registerCommandListener => pass passedProps


jfrolich:
### Issue Description

It would be nice to have the passedProps available in registerCommandListener. For instance when globally sending screen events for analytics purposes.

guyca:
This is indeed a great suggestion, We've noticed this as well. The issue is actually pretty simple to fix, perhaps someone would like to submit a PR.

Basically, passProps are omitted from commands and stored in internally. This is done since props are not guaranteed to be serialisable so they can't be passed over the bridge. Also, props might be pretty large, in which case passing them over the bridge hinders performance.

Once a component is mounted we inject props and clear them from store. We should continue omitting props from objects passed over the bridge, while still including them in command events.

jfrolich:
OK let me see if I can add a pull request for this

pribeh:
@jfrolich any luck with this?

jfrolich:
Nope, wish I had the time, but we need this soon in our app, so I'll will have to look into this soon.

jfrolich:
Just published a pull request for this!

pribeh:
@jfrolich wicked! How would i use copiedLayoutNode then?

jfrolich:
You don't need to use that, that just makes sure that passed props is not always null in the screen parameter of `registerCommandListener`.

guyca:
Thanks to @yogevbd, passProps are now available in command events 👍

Frenchford:
@guyca what about v7.1.0?
passProps are unavailable

PR #4930, opened 2019-03-30
Allow to access passProps in registerCommandListener

jfrolich:
see issue: https://github.com/wix/react-native-navigation/issues/4716

### DIFF
--- a/lib/src/commands/Commands.ts
+++ b/lib/src/commands/Commands.ts
@@ -92,7 +92,7 @@ export class Commands {
 
     const commandId = this.uniqueIdProvider.generate('push');
     const result = this.nativeCommandsSender.push(commandId, componentId, layout);
-    this.commandsObserver.notify('push', { commandId, componentId, copiedLayout });
+    this.commandsObserver.notify('push', { commandId, componentId, layout: copiedLayout });
     return result;
   }
 
--- a/lib/src/commands/Commands.ts
+++ b/lib/src/commands/Commands.ts
@@ -60,11 +60,12 @@ export class Commands {
   public showModal(layout: Layout) {
     const layoutCloned = _.cloneDeep(layout);
     const layoutNode = this.layoutTreeParser.parse(layoutCloned);
+    const copiedLayoutNode = _.cloneDeep(layoutNode);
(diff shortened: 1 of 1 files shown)


lib/src/commands/Commands.ts
```diff
@@ -86,11 +87,12 @@ export class Commands {
     const input = _.cloneDeep(simpleApi);
 
     const layout = this.layoutTreeParser.parse(input);
+    const copiedLayout = _.cloneDeep(layout);
     this.layoutTreeCrawler.crawl(layout);
 
     const commandId = this.uniqueIdProvider.generate('push');
     const result = this.nativeCommandsSender.push(commandId, componentId, layout);
-    this.commandsObserver.notify('push', { commandId, componentId, layout });
+    this.commandsObserver.notify('push', { commandId, componentId, copiedLayout });
```
guyca: @jfrolich This line broke the tests, it should probably be `layout: copiedLayout`

jfrolich:
Yes. Sorry had to retype the commit after prettier changed all the formatting, introducing this typo. Should be good now.

guyca:
Hey @jfrolich 
I just realised that this change is a bit problematic. Your commit changes things so that the original layout (which includes passProps) is passed to the command - but since it hasn't been processed by RNN, **layout nodes don't include id's**.

We omit passProps because we don't want to pass them over the bridge. As they might be very big (think about passing base64 image in passProps), they might put unneeded strain on the bridge.

We can't merge your solution as it currently is. What we can do, is add a function to `LayoutTreeCrawler` which will add props to components after the command has passed over the bridge, and before we notify observers.
WDYT?

jfrolich:
Hmm yeah I think we need to seperate removing the passProps from layoutCrawler, so we can deepClone in between.

guyca:
Alright, I'm leaving the PR open in case you'll want to reiterate. I'm also available on discord if you want to consult with me while working on the PR 👍

jfrolich:
Yes, will do

jfrolich:
Need to keep this one active.

guyca:
This was fixed in a different PR and already available in master 👍

PR #4947, opened 2019-04-04
 Explicit attaching and detaching viewControllers in store

yogevbd:
Register component when viewController is created and release it on dealloc.
This PR allow's more explicit behaviour of attaching and detaching components.
This also solves crashes on multiple setRoot and other edge cases


Collin3:
I think there is a regression from these changes related to the topbar buttons. I upgraded from version `v2.16.0` to `v2.17.0` and now my top bar buttons are not showing up. I have a right button component defined and set with setDefaultOptions, but that doesn't seem to be working anymore. To get around it I have to define static options or do a mergeOptions to redefine that icon in my top bar on every single screen in my app. 

I have also tried with left buttons and by using an `icon` instead of a `component` for my buttons, but those seem to be broken as well when set as default options.

I traced it back to the `v2.16.0-snapshot.270` release which lead me to this PR. That's when it was first introduced and it is also still an issue for me on the latest snapshot `v2.17.0-snapshot.283`.

Issue #5047, opened 2019-04-30
[V2][Android] StatusBar `drawBehind` can not be set to false after being set to true


aminosman:
### Issue Description

When you have a tab based app with different `statusBar` with different `drawBehind` settings each tab does not respect its setting. In my setup we have `statusBar` (and `topBar`) hidden with `drawBehind` set to false and the next tab. On the next tab over both are visible and `drawBehind` is false however the `topBar` is being drawn under the status bar. It seems like the issue is once you set `drawBehind: true` on the status bar, setting it to false does not take. 

This seems to be the same issue seen in these issues
https://github.com/wix/react-native-navigation/issues/3729
https://github.com/wix/react-native-navigation/issues/4857

### Environment
* React Native Navigation version: 2.18.3
* React Native version: 0.57.8
* Platform(s) (iOS, Android, or both?): Android
* Device info (Simulator/Device? OS version? Debug/Release?): All

adeelraza:
+1

aminosman:
Created a PR with a fix 
https://github.com/wix/react-native-navigation/pull/5052

cesarm16:
+1

LRNZ09:
I am having the same issue.

LRNZ09:
Still an issue.

PR #5052, opened 2019-04-30
Fix issue with status bar draw behind change from screen to screen

aminosman:
Right now if you set `statusBar` `drawBehind` to true the `topBar` will always get drawn under the `statusBar` even if the screen settings are to `drawBehind:false`. This fix allows the `topMargin` to change even if the `statusBar` is still visible.

Addresses: https://github.com/wix/react-native-navigation/issues/5047

### DIFF
--- a/lib/android/app/src/main/java/com/reactnativenavigation/presentation/Presenter.java
+++ b/lib/android/app/src/main/java/com/reactnativenavigation/presentation/Presenter.java
@@ -106,10 +106,8 @@ public class Presenter {
     }
 
     private void setDrawBehindStatusBar(View view, StatusBarOptions statusBar) {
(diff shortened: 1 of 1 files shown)


guyca:
This change introduces unwanted margin when status bar is visible.
I'm refactoring the entire logic of drawing under status bar, currently the feature isn't working properly as it causes other views to shift if a screen is pushed with the drawBehind flag enabled.

aminosman:
@guyca when can we expect the rewrite? I have patched our app for now as everything seems to work fine. Can you tell me more about under what circumstances this unwanted margin becomes an issue so I can make sure our app does not suffer from it in the mean time?

LRNZ09:
I'm also having this issue on `2.20.2`

PR #5076, opened 2019-05-06
update gradle typo

LoganArnett:
This PR now implements a much simpler solution for ignoring the other RNN flavors and fixes a typo found while installing.

### DIFF
--- a/docs/docs/Installing.md
+++ b/docs/docs/Installing.md
@@ -226,7 +226,7 @@ To do so edit `android/build.gradle` and add:
 +            android {
 +                variantFilter { variant ->
 +                    def names = variant.flavors*.name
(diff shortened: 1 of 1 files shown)


docs/docs/Installing.md
```diff
@@ -226,7 +226,7 @@ To do so edit `android/build.gradle` and add:
 +            android {
 +                variantFilter { variant ->
 +                    def names = variant.flavors*.name
-+                    if (names.contains("reactNative51") || names.contains("reactNative55")) {
++                    if (!names.contains("reactNative57")) {
```
LoganArnett: This is the main change where we can simply check if the name does not contain the exact version of you want

Issue #5010, opened 2019-04-21
[V2] [Android] Calling setRoot a second time or currentTabIndex breaks the bottom tab bar


wfilleman:
### Issue Description

Running 2.17.0-snapshot.285 calling setRoot multiple times in the app's lifecycle or calling
```js
Navigation.mergeOptions('MainBottomTab', {
    bottomTabs: {
      currentTabIndex: 0,
    },
  });
```
causes the Android bottom tab bar to become stuck on tab index 0. Trying to tap on any other tab causes the tab bar to appear to switch to that tab, but it immediately reverts back to tab index 0. Only killing the app and restarting it can restore bottom tab functionality until the next setRoot or currentTabIndex call. 

### Steps to Reproduce / Code Snippets / Screenshots

On Android call setRoot multiple times with two or more bottomTabs or call 
```js
Navigation.mergeOptions('MainBottomTab', {
    bottomTabs: {
      currentTabIndex: 0,
    },
  });
```
after setRoot.

---
### Environment
* React Native Navigation version: 2.17.0-snapshot.285
* React Native version: 0.59.5
* Platform(s) (iOS, Android, or both?): Android
* Device info (Simulator/Device? OS version? Debug/Release?): Emulator and Device 9.0. Debug only tested so far.

matt-oakes:
I downgraded to 2.16.0 which does not have this issue.

React Native Navigation version: >= 2.17.0
React Native version: 0.59.3
Platform(s) (iOS, Android, or both?): Android only
Device info (Simulator/Device? OS version? Debug/Release?): Tested on Android Emulator, multiple versions in debug mode

MaxMaes:
I have the same issue on Android after upgrading to 2.18.3. When I call currentTabIndex with my BottomTabBar passed as componentid it breaks. But when I use one of my screens component ids, its all fine.

For now im downgrading as well as this is a wonky workaround.

kentdev92:
@matt-oakes @MaxMaes except downgrade version of RNN anything else guys? Because i have downgraded to RNN ver 2.16.0 it's still happened. But my case is different, i have custom topbar: button search and textinput also, click to search button i have mergeOptions to show TextInput, and click Cancel button to hide TextInput, 2 time work as well but 3 time when i click Cancel button, my app was crash! Please help me if you have any ideas to fix this!

minhtc:
[...]

PR #5089, opened 2019-05-09
Fix merge options current tab index

minhtc:
fix https://github.com/wix/react-native-navigation/issues/5010

```js
Navigation.mergeOptions('MainBottomTab', {
    bottomTabs: {
      currentTabIndex: 0,
    },
  });
```
causes the Android bottom tab bar to become stuck on tab index 0. Trying to tap on any other tab causes the tab bar to appear to switch to that tab, but it immediately reverts back to tab index 0. Only killing the app and restarting it can restore bottom tab functionality until the next setRoot or currentTabIndex call.


arnireynir:
Any news? I really need this.

guyca:
Hey @minhtc, thanks for taking the time to look into this. Should be fixed by #5107

solonifer:
Same issue in v3, rn 0.59.10
Setting currentTabIndex cause stacking on the tab and become unchangeable.

Issue #4724, opened 2019-02-12
[V2] Installed app with bottom tabs crashes on startup


holparb:
### Issue Description

The app runs without problems when run with the react-native cli using `react-native run-android`.
**However when starting the already installed app on the phone (not with the react-native cli) the app crashes with the following NullPointerException**:

```
java.lang.NullPointerException: Attempt to invoke virtual method 'android.graphics.drawable.Drawable android.graphics.drawable.Drawable.mutate()' on a null object reference
        at com.aurelhubert.ahbottomnavigation.AHHelper.getTintDrawable(AHHelper.java:36)
        at com.aurelhubert.ahbottomnavigation.AHBottomNavigation.createSmallItems(AHBottomNavigation.java:584)
        at com.aurelhubert.ahbottomnavigation.AHBottomNavigation.createItems(AHBottomNavigation.java:292)
        at com.reactnativenavigation.views.BottomTabs.superCreateItems(BottomTabs.java:48)
        at com.reactnativenavigation.views.BottomTabs.createItems(BottomTabs.java:36)
        at com.reactnativenavigation.views.BottomTabs.enableItemsCreation(BottomTabs.java:24)
        at com.reactnativenavigation.viewcontrollers.bottomtabs.BottomTabsController.applyOptions(BottomTabsController.java:91)
        at com.reactnativenavigation.viewcontrollers.ViewController.onViewAppeared(ViewController.java:204)
        at com.reactnativenavigation.viewcontrollers.ChildController.onViewAppeared(ChildController.java:34)
        at com.reactnativenavigation.viewcontrollers.ViewController.onGlobalLayout(ViewController.java:257)
        at android.view.ViewTreeObserver.dispatchOnGlobalLayout(ViewTreeObserver.java:945)
        at android.view.ViewRootImpl.performTraversals(ViewRootImpl.java:2273)
        at android.view.ViewRootImpl.doTraversal(ViewRootImpl.java:1421)
        at android.view.ViewRootImpl$TraversalRunnable.run(ViewRootImpl.java:6842)
        at android.view.Choreographer$CallbackRecord.run(Choreographer.java:1026)
        at android.view.Choreographer.doCallbacks(Choreographer.java:838)
        at android.view.Choreographer.doFrame(Choreographer.java:769)
        at android.view.Choreographer$FrameDisplayEventReceiver.run(Choreographer.java:1012)
        at android.os.Handler.handleCallback(Handler.java:789)
[...]

PR #5132, opened 2019-05-22
fix_4724: Fixes android crashes on startup

MakhouT:
Fixes the startup crashes on android when releasing.
https://github.com/wix/react-native-navigation/issues/4724

### DIFF
--- a/playground/android/app/build.gradle
+++ b/playground/android/app/build.gradle
@@ -2,7 +2,10 @@ apply plugin: "com.android.application"
 
 project.ext.react = [
         root     : "../../../",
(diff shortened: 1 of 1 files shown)


guyca:
Hey @MakhouT 
I don't think this is an RNN issue - notice you've modified a file in the playground application, not the library itself.

MakhouT:
Correct. However this fixes the crashes in the playground when releasing. I had the issue with the playground when trying to run it. But I indeed agree that the fix is not related to RNN. This PR is just to make the playground run in release mode.

Issue #5140, opened 2019-05-23
[V2][iOS] setStackRoot has no animation


Jazqa:
### Issue Description

setStackRoot has no push animation [(same as this issue, but for setStackRoot)](https://github.com/wix/react-native-navigation/commit/42e26d77b8d231debe0489dbe874fc06d9a97589)

### Steps to Reproduce / Code Snippets / Screenshots

1. setStackRoot
2. new StackRoot just flashes in without the platform push animation

---
### Environment
* React Native Navigation version: 2.20
* React Native version: 0.59.8
* Platform(s) (iOS, Android, or both?): iOS

Jazqa:
Fixed in #5217

PR #5141, opened 2019-05-23
[V2][iOS] Added waitForRender support for setStackRoot

### LINKED ISSUE #5133: [iOS] Animation properties (waitForRender) for setStackRoot
### Issue Description

Having similar animation properties for `setStackRoot` and `push` would be very useful. 

At the moment, I am using `waitForRender` for screens that use `ImageBackground`, because without `waitForRender` the Application's `layout.backgroundColor` flashes before the image is rendered for each view, making the navigation flow look very disturbing.

The undocumented `waitForRender` animation property already exists and works for `showModal`, `push` and `setRoot`, but unfortunately `setStackRoot` is missing animation support completely, making it look very inconsistent compared to the other navigation actions used in the application.

EDIT: Investigated this more and on Android the `animations.push` is also used for `setStackRoot`, making this iOS only issue. [Solution for iOS in this comment](https://github.com/wix/react-native-navigation/pull/5217#issuecomment-503988547).

Jazqa:
setStackRoot.animations.waitForRender and setStackRoot.animations.enable are read from the last child of the new stack. Fixes #5133 (no custom animations, just waitForRender) and #5140.


Jazqa:
I am uncertain whether you prefer having `animations.setStackRoot` as an independent option or if you want to use `animations.push` for setRoot, [since iOS code suggests you'd support `animations.setStackRoot`](https://github.com/wix/react-native-navigation/pull/5141/files#diff-402a1f2cdd98bfe5ae49b8eecb54f391L181), [but Android is using `animations.push` even when calling `setStackRoot`](https://github.com/wix/react-native-navigation/blob/99032e060d2e0a429d3da2775884f624e8cd5fd5/lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/stack/StackController.java#L150).

The PR does not require any Android changes, unless you want to support the `animations.setStackRoot`option, because all the functionality already exists through `animations.push`.

kyle-ssg:
Needed this for so long, please merge!

JunoVW:
I also really need this!

Jazqa:
Renamed the PR, since the Animation itself was fixed in #5217.  Unlike @yogevbd's comment in the linked PR, `waitForRender` support was not added in the previous PR, so this PR would still add `waitForRender` support for `setStackRoot`.

Jazqa:
Come on. Since the feature already exists on Android and even on iOS for every other navigation method, this is an obvious bug. The fix is also extremely simple and is already used for all the other navigation methods on iOS. 

Such a tiny fix would improve the user experience by eliminating app's background color flashing when navigating with `setStackRoot` to a screen with large `<Image/>` or `<ImageBackground/>`.  Unfortunately, the app I'm working on has lots of cases like this.

If there is something blatantly wrong with the PR, at least drop a comment and I'll dedicate some of my time to fix it.

sintylapse:
@Jazqa are you sure this is implemented by default on Android? `waitForRender` doesn't work for me on Android with `setStackRoot`

Jazqa:
@sintylapse, a quick peek from https://github.com/wix/react-native-navigation/blob/99032e060d2e0a429d3da2775884f624e8cd5fd5/lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/stack/StackController.java#L150 suggests that setStackRoot would inherit .push options

Jazqa:
@sintylapse, if I remember correctly, Android’s setStackRoot uses the animation options of the push method. I’m not 100% sure, but I’ll check first thing monday morning. If that’s not the case, I’ll open a PR for Android implementation.

sintylapse:
@Jazqa sounds great, thanks a lot!

Jazqa:
@sintylapse, you are right, the `waitForRender` for `setStackRoot` is not working on Android. In fact, even the native animation of `setStackRoot` is not working on Android.

@yogevbd, the issue is [new stack being initialised](https://github.com/wix/react-native-navigation/blob/898cf7ae2703656459f28e21123f3b0f8a40b22e/lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/stack/StackController.java#L186) before the animation starts, which causes [the push method to assume there is no ViewController toRemove and completely ignore the animations](https://github.com/wix/react-native-navigation/blob/898cf7ae2703656459f28e21123f3b0f8a40b22e/lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/stack/StackController.java#L143). One possible solution is duplicating `push` method to a new `setStackRoot` method that would take the old stack as a parameter. That solution would also make `options.animations.setStackRoot` available for Android (at the moment `options.animations.push` is being used).

sintylapse:
@Jazqa thank you for your work!

exentrich:
On iOS in version `3.1.2` `setStackRoot.enabled = false` not works. Animation now always performed.
Worse than that animation always looks as `push`. Configuration for `animations.setStackRoot` is ignored in any case (global, per screen, on setStackRoot).
Can you imagine that `setStackRoot` animated like usual `push`? This is weird UX!

Please revert back changes or fix this behaviour, at least support for `enabled = false`.

Issue #5138, opened 2019-05-23
[v2][iOS] Custom Topbar Background not working properly


FRizzonelli:
### Issue Description

So basically there is an issue with custom component in background of the topbar. Issue is introduced from build **2.17.0**, with previous releases it works! I'm using a bottomTabs layout with a stack in each children. I've already tried to also remove the stack and set only a component for each tab but the bug remains. 
Basically what is happening is in the first tab I'm seeing the custom component, but it's covering title and buttons (probably related to this https://github.com/wix/react-native-navigation/issues/5026). In second and third tab the custom background isn't even rendered.

My setup use one setRoot for the login, and after the login is successful, I set another root with the bottomTabs layout (I've already tried to skip the first setRoot and use only bottomTabs, bug remains).

Sharing my layout and a video of the working build (2.16.0) vs the bugged one (2.17.0). I've digged into the commits, seems related to changes made [here](https://github.com/wix/react-native-navigation/commit/ac60d2fe6ad036528c31954a2997109b06f0c947) or [here](https://github.com/wix/react-native-navigation/commit/28300597ede5de1f08d7b32ba4a9313ffdf4aac1)

I've tried all releases after 2.17.0, and it keeps to be bugged.

Maybe @yogevbd could help us to understand why is this happening?

### Steps to Reproduce / Code Snippets / Screenshots

This is my setRoot
```ts
Navigation.setRoot({
    root: {
        bottomTabs: {
            children: [
                {
                    stack: {
                        children: [
                            {
                                component: {
                                    name: ScreenKeys.dashboardScreen,
                                    options: navigatorStandardOptions({ title: 'Dashboard' })
                                }
                            }
                        ],
                        options: bottomTabStandardOptions({
                            title: 'Dashboard',
                            icon: 'dashboard-outline',
                            selectedIcon: 'dashboard'
                        })
                    }
                },
                {
                    stack: {
                        children: [
                            {
                                component: {
                                    name: ScreenKeys.timesheetScreen,
                                    options: navigatorStandardOptions({ title: 'Presenze' })
                                }
                            }
                        ],
                        options: bottomTabStandardOptions({
                            title: 'Presenze',
                            icon: 'watch-outline',
                            selectedIcon: 'watch'
                        })
                    }
                },
                {
                    stack: {
                        children: [
                            {
                                component: {
                                    name: ScreenKeys.expenseSectionScreen,
[...]

PR #5144, opened 2019-05-23
Fix topBar.background.component render behind title

yogevbd:
Addresses #5138


FRizzonelli:
Perfectly working! Thank you sir @yogevbd

PR #5153, opened 2019-05-26
Set titleDisplayMode to hidden when tabes have no labels

guyca:
If all BottomTab items have no labels defined - set TitleDIsplayMode to ALWAYS_HIDE so the icons are centred vertically.


guyca:
This is a bad idea.

Issue #5180, opened 2019-06-04
[iOS] set topBar leftButton with custom component via mergeOptions crashes


MarianPalkus:
Hi, 
thanks for publishing/maintaining this library and your work!

### Issue Description

I try to set a custom component to `topBar.leftButtons` via `mergeOptions` (maybe this issue is related to side menus):

```js
const CustomTopBarButton = ( )=> {
  return <Text>TEST BUTTON</Text>
}

Navigation.registerComponent('CUSTOM_TOP_BAR_BUTTON', () => CustomTopBarButton);

Navigation.mergeOptions(this, {
      topBar: {
        leftButtons: [
          {
            id: 'CUSTOM_BUTTON_ID',
            component: {
              name: 'CUSTOM_TOP_BAR_BUTTON',
            }
          }
        ],
      }
    })
```

```
Exception 'NSLayoutConstraint for (null): Constraint must contain a first layout item' was thrown while invoking mergeOptions on target RNNBridgeModule with params (
    SideMenuCenter,
        {
        topBar =         {
            leftButtons =             (
                                {
                    component =                     {
                        componentId = CustomComponent19;
                        name = "CUSTOM_TOP_BAR_BUTTON";
                    };
                    id = "CUSTOM_BUTTON_ID";
                }
            );
        };
    },
    694,
    695
)
callstack: (
	0   CoreFoundation                      0x000000010427f1bb __exceptionPreprocess + 331
	1   libobjc.A.dylib                     0x0000000102964735 objc_exception_throw + 48
	2   CoreFoundation                      0x000000010427f015 +[NSException raise:format:] + 197
	3   Foundation                          0x000000010258bfbd ResolveConstraintArguments + 464
	4   Foundation                          0x000000010258c2cb +[NSLayoutConstraint constraintWithItem:attribute:relatedBy:toItem:attribute:multiplier:constant:] + 80
	5   playground                          0x000000010058f2bd -[RNNUIBarButtonItem init:withCustomView:] + 397
	6   playground                          0x00000001005d06ee -[RNNNavigationButtons buildButton:defaultStyle:insets:] + 2478
	7   playground                          0x00000001005cf67f -[RNNNavigationButtons setButtons:side:animated:defaultStyle:insets:] + 639
	8   playground                          0x00000001005cf27c -[RNNNavigationButtons applyLeftButtons:rightButtons:defaultLeftButtonStyle:defaultRightButtonStyle:] + 492
[...]

PR #5189, opened 2019-06-06
Fix crash on ios when setting topbar left buttons with custom component

MarianPalkus:
Possible fix for crash on ios when setting topbar left buttons with custom component.

Fixes issue #5180.


hemith:
Setting topbar left buttons with custom component on android is not working too. check it out please.

MarianPalkus:
@hemith i could not reproduce a crash for the same situation on android (emulator).

hemith:
@MarianPalkus OK I got it. 
I have a problem setting topbar left buttons with custom component on android.
Right buttons are appear. but left buttons are not appear.  (no crash)
I didn't check it out on iOS.
Anyway i think my problem is not much related with this pull request.
thank you! :-)

MarianPalkus:
@hemith Ok, thanks for clarification :-)

PR #5208, opened 2019-06-16
feat: add support to load icon from server

zcgong:


guyca:
Hey,
This is interesting, I don't think RNN should have a network stack. A better approach is for you to download the file, save it to disk then require it and use it.
Moreover, your implementation invokes network calls on the main thread...

PR #5217, opened 2019-06-18
Fix setStackRoot options

yogevbd:
`setStackRoot` options should be resolve from the last presented viewController and not from `getCurrentChild` of this viewController.

### DIFF
--- a/lib/ios/RNNCommandsHandler.m
+++ b/lib/ios/RNNCommandsHandler.m
@@ -55,23 +55,23 @@ static NSString* const setDefaultOptions	= @"setDefaultOptions";
 
 - (void)setRoot:(NSDictionary*)layout commandId:(NSString*)commandId completion:(RNNTransitionCompletionBlock)completion {
 	[self assertReady];
-
+	
 	if (@available(iOS 9, *)) {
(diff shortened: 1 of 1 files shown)


Jazqa:
Could also resolve the renderTreeAndWait from the last presented viewController #5141, #5133

yogevbd:
Yes it should solve those issues as well. Can you please test it with this version: 2.21.0-snapshot.369?

Jazqa:
@yogevbd, it still doesn't support the `waitForRender` (on iOS terms `renderTreeAndWait`). You'll have to wrap the 

    [_stackManager setStackChildren:childViewControllers fromViewController:fromVC animated:[options.animations.setStackRoot.enable getWithDefaultValue:YES] completion:^{
		[weakEventEmitter sendOnNavigationCommandCompletion:setStackRoot commandId:commandId params:@{@"componentId": componentId}];
		completion();

inside 

    [childViewControllers.lastObject renderTreeAndWait:([options.animations.setStackRoot.waitForRender getWithDefaultValue:NO]) perform:^{ ...}];


push, showModal and every other action is already doing that

PR #5280, opened 2019-07-13
Allow setting UITabBarItem.title to default value

danilobuerger:
The default value for `UITabBarItem.title` is `nil` as specified by the docs:
> You should set this property before adding the item to a bar. The default value is nil.
https://developer.apple.com/documentation/uikit/uibaritem/1616412-title

This changes allows for this to be `nil`.

### DIFF
--- a/lib/ios/RNNTabBarItemCreator.m
+++ b/lib/ios/RNNTabBarItemCreator.m
@@ -12,7 +12,7 @@
 	
 	tabItem.image = [self getIconImage:icon withTint:iconColor];
 	tabItem.selectedImage = [self getSelectedIconImage:selectedIcon selectedIconColor:selectedIconColor];
-	tabItem.title = [bottomTabOptions.text getWithDefaultValue:@""];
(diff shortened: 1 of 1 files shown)


guyca:
Hey @danilobuerger, Thanks for the PR.
Out of curiosity, Has the current logic of setting empty string cause any bugs?

danilobuerger:
@guyca I came across this while trying to vertically center tab items on iOS when there is no title and taking the horizontal size class into account (on iPad >= iOS 11, the title is rendered to the side of the icon, not beneath, so they are already vertically centered and don't need a inset). This can't be achieved via js-land as it requires handling inside the `traitCollectionDidChange` callback. Anyway, while doing that, I came across this and just made the PR to get the correct setting. It doesn't actually fix a bug (that I know of).

As a side note: If centering tab items while taking horizontal size classes into account is something desired by rnn, I can submit a separate PR for that too.

guyca:
I wasn't aware there's an issue with tab items not being centered. If you can resolve the issue in your own project, I suggest we refrain from making a PR regarding this until there are issues reported about it.

PR #5283, opened 2019-07-14
BottomTab dot indicator

### LINKED ISSUE #4910: Possibility to have tab indicator
Is it possible to have a feature like the active indicator on bottom tabs? like a line or something? or maybe it can be achieved through some option (I didn't see any doc about this).

Thank you for your full support

guyca:
This pr adds support for adding a dot indicator to BottomTabs. When both textual badge and dot indicator are defined, the indicator takes precedence, both elements won't be visible at the same time.

image

# API
```js
options: {
  bottomTab: {
    dotIndicator: {
      color: 'green', // default red
      size: 8, // default 8
      visible: true // default false
    }
  }
}
```

# Hiding the indicator
To hide the indicator, call `mergeOptions` with `visible: false`.

```js
      bottomTab: {
        dotIndicator: { visible: false }
      }
    });
```

Fixes #4910

### DIFF
--- a/lib/ios/RNNBasePresenter.m
+++ b/lib/ios/RNNBasePresenter.m
@@ -5,11 +5,19 @@
 #import "UIViewController+LayoutProtocol.h"
 #import "DotIndicatorOptions.h"
 #import "UITabBar+Utils.h"
+#import "RNNDotIndicatorPresenter.h"
 
 @interface RNNBasePresenter ()
+@property(nonatomic, strong) RNNDotIndicatorPresenter* dotIndicatorPresenter;
 @end
 
 @implementation RNNBasePresenter
+
+-(instancetype)init {
+    self = [super init];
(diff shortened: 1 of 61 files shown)


lib/ios/RNNBasePresenter.m
```diff
@@ -3,140 +3,159 @@
...
-		readyBlock();
-		readyBlock = nil;
-	}
+    if (readyBlock) {
+        readyBlock();
+        readyBlock = nil;
+    }
 }
 
+- (void)viewDidLayoutSubviews {
```
yogevbd: Not needed

maleriepace:
You did it! Thank you! :)

emilioheinz:
This should be available after what version?

guyca:
Yeah, Took a while.. even learned objc for this one 😅

PR #5297, opened 2019-07-19
[ios] Fix bug that reverts navbar title size to 17

dcvz:
Title says it all :)


dcvz:
Only changes iOS files.. so not really sure why Android tests are failing

PR #5298, opened 2019-07-19
[ios] Fix setting nav bar no border

dcvz:
At the moment this is broken on iOS - this fixes it


guyca:
Hey there, `noBorder` seems to work on the playground app.
Can you please post a use case in which it doesn't work?

Issue #4669, opened 2019-01-31
[boilerplate] RNN V2 + MobX + Firebase starter kit


alexwasner:
I made a starter kit for V2 for anyone that wants to use Firebase and/or MobX instead of Redux. Others opened tickets for this, so I'm not sure the normal process. Maybe @guyca can add it to the readme if people find it useful https://github.com/alexwasner/react-native-mobx-firebase-starter

retyui:
The best logo :+1: 
![sorry](https://raw.githubusercontent.com/alexwasner/react-native-mobx-firebase-starter/master/src/assets/images/logo.png)

ItsNoHax:
@alexwasner Looks good, I suggest you create a PR to include this in the docs' showcases/starter kits.

Going to close this issue as I'm trying to clean up issues. Please don't forget to create a PR!

PR #5304, opened 2019-07-21
[V2] [Docs] Add alexwasner/RNMFS starter kit

alexwasner:
I mentioned in #4669 that I created a starter kit. Here is a link in the docs to it (only a few months after the fact)


guyca:
Amazing work! And that logo is sick! 🔥

PR #5317, opened 2019-07-25
Fix transparent topBar background color transition

yogevbd:
• Fix the transition between transparent background color to colorized background.
• Remove storing topBar background image.


cesarm16:
is this on V3.alpha-11?

PR #5318, opened 2019-07-25
[v3] Added autolinking for android

danilobuerger:
Introduced with rn 0.60 we now have autolinking: https://github.com/react-native-community/cli/blob/master/docs/autolinking.md

It works out of the box with rnn on iOS, but needs modification for Android. Since the Android source dir is not the default `android`, the `react-native.config.js` file was added to specify the correct path.

The cli then wants to include a package. However, the rnn package needs access to the `ReactNativeHost` which is not available when the packages are instantiated. So, with this PR I added a dummy package to be included by the cli.


luskin:
All we had to do was swap out the `react-native-navigation` recommended method with this in our `MainApplication.java` and auto-linking works great:

```java
  protected List<ReactPackage> getPackages() {
    List<ReactPackage> packages = new PackageList(this).getPackages();  // <-- this
    return packages;
  }
```

Bogidon:
Can confirm that @luskin's solution attached my auto-linked packages correctly.

LeoMosiah:
> All we had to do was swap out the `react-native-navigation` recommended method with this in our `MainApplication.java` and auto-linking works great:
> 
> ```java
>   protected List<ReactPackage> getPackages() {
>     @SuppressWarnings("UnnecessaryLocalVariable")
>     List<ReactPackage> packages = new PackageList(this).getPackages();  // <-- this
>     // Packages that cannot be autolinked yet can be added manually here, for example:
>     // packages.add(new MyReactNativePackage());
>     return packages;
>   }
> ```

Can you show me how your ```MainApplication.java``` looked?

yangfanyu:
> All we had to do was swap out the `react-native-navigation` recommended method with this in our `MainApplication.java` and auto-linking works great:
> 
> ```java
>   protected List<ReactPackage> getPackages() {
>     @SuppressWarnings("UnnecessaryLocalVariable")
>     List<ReactPackage> packages = new PackageList(this).getPackages();  // <-- this
>     // Packages that cannot be autolinked yet can be added manually here, for example:
>     // packages.add(new MyReactNativePackage());
>     return packages;
>   }
> ```

thanks,  it woked

rendomnet:
> ```java
> protected List<ReactPackage> getPackages() {
>     @SuppressWarnings("UnnecessaryLocalVariable")
>     List<ReactPackage> packages = new PackageList(this).getPackages();  // <-- this
>     // Packages that cannot be autolinked yet can be added manually here, for example:
>     // packages.add(new MyReactNativePackage());
>     return packages;
>   }
> ```

So to install RNN 3.2+ we need to only modify "MainApplication.java" and run "yarn add react-native-navigation" ? no other edits required?

PR #5338, opened 2019-07-30
[v3] Do not show and hide bottom tabs automatically in nested scrolling views on Android

jliuhtonen:
When `onMeasure()` is called on BottomTabs for the first time, the parent class
`AHBottomNavigation` will create and attach an
[`AHBottomNavigationBehavior`](https://github.com/wix-playground/ahbottomnavigation/blob/master/ahbottomnavigation/src/main/java/com/aurelhubert/ahbottomnavigation/AHBottomNavigation.java#L1281) that listens to `onNestedScroll()`, showing and hiding the bottom tabs respectively when nested scrolling happens up or down.

This fix disables the functionality by setting `behaviorTranslationEnabled` in `AHBottomNavigation` to `false`, until this behaviour can be controlled through the react-native-navigation options from React Native code.


guyca:
@jliuhtonen Thanks a lot for the PR! Unfortunately I wasn't quick enough to merge it and the diff is too big, I went ahead and added this one liner in 76881964814e0545f9e4f124e5a3874e5ac713ce

jliuhtonen:
Great, thanks @guyca! Yeah, I didn't rebase the branch on top of master, which probably explains the diff.

PR #5342, opened 2019-08-01
Send back button navigation event

eightyfive:


bneigher:
@eightyfive is this also for iOS?

eightyfive:
Hey, nope. I did not do the iOS part, which should be a showstopper actually.

Problem is I have very (very) basic knowledge of Objective-C, so I won't be able to dig into the code.

Maybe someone at Wix can do that part?

bneigher:
yea that would be really great.  this is basically the last issue I am having changing from react-native-router-flux.

These events are critical to making custom UI's know exactly when the nav state changes.

Strange tho - it works in android for me without your PR.  I believe if you add an `id` to the backButton style option it should fire the event

eightyfive:
How do you add the `id`? I mean do you provide a custom image as well? Or does it just work specifying the `id` in config without a custom image/icon.

Cause yeah, I'd like to keep the default native image provided by the library/OS. Not to mention that it's a lot less "config code" if the event is triggered for library's `Constants.BACK_BUTTON_ID`.

bneigher:
Here is my defaultConfig:
```
  statusBar: {
    style: Platform.select({ ios: "dark", android: "light" }),
    drawBehind: true
  },
  topBar: {
    elevation: 2,
    title: {
      alignment: "center",
      component: {
        alignment: "center"
      }
    },
    backButton: {
      id: "BACK",
      icon: backIcon
    }
  },
  modalPresentationStyle: "overCurrentContext",
  layout: {
    backgroundColor: "transparent",
    componentBackgroundColor: "transparent"
  },
  screenBackgroundColor: "transparent",
  popGesture: false
});
```

Here is my event
```
Navigation.events().registerNavigationButtonPressedListener(event => {
  if (event.buttonId === "BACK") {
    handledPop(screen.componentId); // my custom pop wrapper
  }
});
```

Basically (on Android) in order to get the event to fire at all, you would need to manually add the id: "BACK" or something to the backButton options.

But I agree with you nontheless - we shouldn't need to do this (unless there are significant performance costs to by default send the event with the Constants.BACK_BUTTON_ID).

That would be the only stipulation that would make me think not sending it without the id was done on purpose.

eightyfive:
Closing.

See https://github.com/wix/react-native-navigation/pull/4226#issuecomment-433683885

> This change breaks parity with iOS ⚠️

I guess we have our answer!

eightyfive:
Same exact [PR #4226](https://github.com/wix/react-native-navigation/pull/4226) closed ~ a year ago.

With similar [commit](https://github.com/wix/react-native-navigation/pull/4226/commits/1c3014b66e147a76d112b3f75fed3f35d081dbf2) to mine.

PR #5345, opened 2019-08-01
[v3] Support autolinking `PackageList`

dcvz:
As part of RN 0.60's auto linking - packages can be auto imported by using the `PackageList` package. Example:

```java
  @Nullable
  @Override
  public List<ReactPackage> createAdditionalReactPackages(ReactNativeHost host) {
    List<ReactPackage> packages = new PackageList(host).getPackages();
    return packages;
  }
```

As part of using the package, you must pass into it a valid ReactNativeHost object. Since `createAdditionalReactPackages` was being called at instantiation, `getReactNativeHost()` returns `null`. I've changed the constructor to allow for the host to be passed into the function.


rendomnet:
any updates?

Issue #4610, opened 2019-01-21
react-native-navigation stucks in “component does not have a componentId”


JoseFMP:
### Issue Description

react-native-navigation stucks in “component does not have a componentId”
I tried:
- Adding and removing the prop componentId
- Making it optional/compulsory
- Setup and not set-up registerAppLaunchedListener 
- Making the prop static

Bulding is fine, but on run time when the debugger hits this line it will complain that FirstScreenName does not have a componentId. However, It really has the prop componentId set.

### Steps to Reproduce / Code Snippets / Screenshots
Create a bare react-native application and try to use react-native-navigation according to the documentation.
This statement crashes:
```ts
Navigation.registerComponent("FirstScreenName", () => () =>
 (<FirstScreen componentId="FirstScreen" />));
```
This also:
```ts
Navigation.registerComponent("FirstScreenName", () => FirstScreen);
```
This also:
```ts
Navigation.registerComponent("FirstScreenName", () => (props) => <FirstScreen {...props} />);
```
Complaining that `FirstScreenName` does not have a componentId.
Am I missing to set the componentId somewhere by hand or I am using the API incorrectly?

The crash happens before hitting the `setRoot` call, so it might be unrelated to `setRoot`. That said, the `setRoot` call is done as described in the documentation:
```ts
Navigation.events().registerAppLaunchedListener(() => {
  Navigation.setRoot({
    root: {
      component: {
        name: 'navigation.playground.WelcomeScreen'
      }
    }
  });
});
```
---
### Environment
* React Native Navigation version: 2.7.1
* React Native version: 0.57.8
* Platform: Android
* Device info :  LG V20, debug mode.

L-Yeiser:
You are registering your component incorrectly. In your case `componentId` will be passed to the second anonymous arrow function.
```
Navigation.registerComponent("FirstScreenName", () => FirstScreen);
```

JoseFMP:
@L-Yeiser  Thank you for your answer. 
I did not understand what you mean. So you mean that  
```
Navigation.registerComponent("FirstScreenName", () => FirstScreen);
```
is the expected way? I get the exact same error.

byonghun:
@JoseFMP The error is probably coming from when you `setRoot`. Can you show me the code where you register your app?

JoseFMP:
@byonghun Thank you for your answer.
Apparently now. I.e. without any `setRoot` call, the described error appears.

byonghun:
[...]

PR #5360, opened 2019-08-06
[Docs] Installation Instructions | Fix nested list numbering

JacobJaffe:
Current numbering causes the line 
```In Xcode, you will need to edit this file: AppDelegate.m. This function is the main entry point for your app:```
to appear as instruction 1, not 3.

image

(And some other indentation weirdness on the docs) 

*I'm unsure how to test the build changes on the website (The changes make the markdown render correctly in places like Github's view though) -- is there a documented process for this?*

I encountered this issue while attempting to set up a new project with RNN V2 -- I ran into very similar issues as #4610 -- I think that the current docs have mislead folks into thinking that step 3 for iOS installation is only necessary if there is an error (I know this is what happened to me at least 😄 ).


guyca:
Thanks a lot @JacobJaffe !

Issue #5348, opened 2019-08-02
[V2] componentWillUnmount not running on screen pop.


SudoPlz:
### Issue Description

I'm pushing a screen like normal, with `Navigation.push`, and then pop it w/ `Navigation.pop`.
I noticed after the pop takes place, the animation runs, the screen is no longer visible, bot `componentWillUnmount` **doesn't** run.

As I investigated further I found out that `Navigation.pop` never returns anything, the promises that it's supposed to return NEVER resolves, or rejects.

I decided to dig further into the code and found out that [this](https://github.com/wix/react-native-navigation/blob/d09d0108d1530cf10e24c46efb6c9d9962807ead/lib/ios/RNNNavigationStackManager.m#L40:L44) never runs and that's because [this](https://github.com/wix/react-native-navigation/blob/d09d0108d1530cf10e24c46efb6c9d9962807ead/lib/ios/RNNNavigationStackManager.m#L87:L89) never runs.

Now I tried to reproduce that on the playground app but couldn't. Not sure why that's happening on my end.

GOOD NEWS though, I also found [this SO answer](https://stackoverflow.com/questions/27470130/catransaction-completion-block-never-fires)  and tried surrounding the animation block with a main thread block.

After that the issue was gone!!
[The completion block runs again](https://github.com/wix/react-native-navigation/blob/d09d0108d1530cf10e24c46efb6c9d9962807ead/lib/ios/RNNNavigationStackManager.m#L87:L89) and so does `componentWillUnmount` on every pop.

I still have no idea what's causing it, but here's how the new code looks like:

```
[self performAnimationBlock:^{
        poppedVC = [viewController.navigationController popViewControllerAnimated:animated];
}
```

is now:

```
[self performAnimationBlock:^{
        dispatch_async(dispatch_get_main_queue(), ^{
            poppedVC = [viewController.navigationController popViewControllerAnimated:animated];
        });
}
```
---
### Environment
* React Native Navigation version: 2.26.2
* React Native version: 0.59.9
* Platform(s) (iOS, Android, or both?): iOS
* Device info (Simulator/Device? OS version? Debug/Release?): iPad Pro 12.9 Inch 3rd Generation iOS 12.4 Simulator

SudoPlz:
@guyca this is the issue I've been telling you about ^, it doesn't reproduce on the playground app, but I'm pretty sure one of your iOS devs will have a better idea as to why that's happening.

SudoPlz:
[...]

PR #5363, opened 2019-08-06
Making sure we run the animation blocks on the main thread

SudoPlz:
- Fixes: https://github.com/wix/react-native-navigation/issues/5348


SudoPlz:
Hello?

SudoPlz:
Anyone there?

guyca:
Seems like native module commands are sometimes not invoked on the main thread. This is related to a performance optimisation RN does, if too many commands are queued on the main thread, module commands are posted on a different thread other then the main thread.
We'll look into this next week, closing the PR for now.

PR #5380, opened 2019-08-16
[Docs] Installation Instructions | Add CocoaPods instructions for iOS

jtibbertsma:
I just generated a new react-native project with react-native 0.60.4 and the installation instructions for iOS don't work. It's much easier to install with cocoapods. I went ahead and added a "Installing with cocoapods" section to the docs.


guyca:
Thanks! 👍

Issue #5305, opened 2019-07-22
[V3] Android FAB incorrect gravity


SalemC:
### Issue Description

I've just migrated from RNN v1 to V3 (lots of changes) & am currently trying to get the android FAB working. The FAB is displayed but doesn't seem to be using the default gravity, so I've tried setting it myself but that doesn't seem to make a difference.

### Steps to Reproduce / Code Snippets / Screenshots

```js
    static styles = {
        RECENT: {
            layout: {
                backgroundColor: colors.OFF_WHITE,
                orientation: ['portrait'],
            },

            topBar: { visible: false }

            statusBar: {
                style: 'light',
                backgroundColor: colors.PRIMARY,
           },
        },
    }

    /**
     * Generate a screen
     */
    static generateScreen = (name, options = {}, passProps = {}) => ({
        root: {
            ...this.generateModal(name, options, passProps),
        },
    })

    /**
     * Generate a modal
     */
    static generateModal = (name, options = {}, passProps = {}) => ({
        stack: {
            children: [{
                component: {
                    name,
                    passProps,
                    options,
                },
            }],
            options: {
                fab: {
                    id: 'fab',
                    visible: true,
                    backgroundColor: '#000000',
                    clickColor: '#ffffff',
                    rippleColor: '#000000',
                    iconColor: '#ffffff',
                    alignVertically: 'bottom',
                    alignHorizontally: 'right',
                    actions: [{
                        id: 'test',
                        backgroundColor: '#000000',
                    }, {
                        id: 'test2',
                        backgroundColor: '#ffffff',
                    }],
                },
            },
        },
    })

    static showMainScreen () {
        const name = 'telltrail.Main.Recent';
        const main = this.generateScreen(name, this.styles.RECENT);

        Navigation.setRoot(main);
    }
```
![fab](https://user-images.githubusercontent.com/32452313/61627586-103d0c00-ac78-11e9-8771-4bcb65fc3826.png)

![fab2](https://user-images.githubusercontent.com/32452313/61627652-395d9c80-ac78-11e9-82e2-843d465b44a1.png)
---
### Environment
* React Native Navigation version: 3.0.0-alpha.2
* React Native version: 0.59.1
* Platform(s) (iOS, Android, or both?): Android
* Device info (Simulator/Device? OS version? Debug/Release?): Simulator, Android 8.0 API 26, Debug

Damar95:
In V3 the layouttype of the StackLayout changed to CoordinatorLayout.
[...]

PR #5382, opened 2019-08-18
[V3, Android] Fix Fab Gravity and Margin

Damar95:
Fix Fab Gravity for V3
ifx for #5305


Damar95:
The margin is still wrong. it seems like it uses the Display and not the inner View.
Because of this the Fab might be hidden behind the TopBar/Tabs or BottomTabs

Damar95:
I added a commit which "fixes" the margin at the bottom.
The problem is, that ist still not correct because the height returned by [getBottomInset()](lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/stack/StackController.java#L387) is to much as the screenshot showes


and with the spacing margin and bottomTabs margin its lookes like this
grafik
grafik

Agungkes:
```
Navigation.setRoot({
      root: {
        bottomTabs: {
          children: [
            {
              stack: {
                children: [
                  {
                    component: {
                      name: 'example-stack',
                    },
                  },
                ],
                options: {
                  bottomTab: {
                    text: 'First',
                    icon: bookOpen,
                  },
                  fab: {
                    id: 'test-fab',
                    visible: true,
                    actions: [
                      {
                        id: 'act-1',
                      },
                      {
                        id: 'act-2',
                      }
                    ]
                  }
                },
              },
            },
            {
              component: {
                name: 'example-1',
                options: {
                  bottomTab: {
                    text: 'example',
                    icon: award,
                  },
                },
              },
            },
            {
              component: {
                name: 'example-test',
                options: {
                  bottomTab: {
                    text: 'example2',
                    icon: user,
                  },
                },
              },
            },
          ]
        },
      },
    });
```
That's my Navigation hierarchy. 

All went well until I switched screens using bottomTabs. After I move to the next then return again, then this error message appears
```
java.lang.NullPointerException: Attempt to invoke virtual method 'android.view.ViewGroup$LayoutParams com.reactnativenavigation.views.Fab.getLayoutParams()' on a null object reference
```
I just want to use fab on the first bottomTabs

Damar95:
@Agungkes i added a new commit which should fix this crash.
Thanks for the detailed error description. :)

Damar95:
@Agungkes as you described in #5447 some options dont work. i added a fix for the visable attribut in the last commit, even its not part of this PR.

Might look into it, but i'm not effected by this bugs so i don't know when...

@guyca You can merge this PR.

Agungkes:
@Damar95 only icon option don't work, so can't change "plus".
 
Hope this PR will merge soon :D thanks @Damar95

Damar95:
@guyca the tests are failing because of this line: https://github.com/Damar95/react-native-navigation/blob/57b5c43980a6be6779b609a0ec0c9880f14f436f/lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/stack/StackController.java#L406

`fabOptionsPresenter.applyTopInset(getTopInset(getCurrentChild()));`

Calling `fabOptionsPresenter.applyTopInset()` with any int as param works.
Calling `getCurrentChild()` works.
Calling `getTopInset(getCurrentChild())` does not work.

Do you know why?

omid-ebrahimi:
We really need this fix. Please merge it.

Damar95:
@omid-ebrahimi, @guyca is not avaiable this week and can't fix the unit tests.

We also found a other problem we need to fix, which needs to be fixed.

itsam:
Could you please advise how to apply this PR to my instance until it is merged in master?
Also, any timeplan when (and if) will be merged to master wiil be highly appreciated.

Thanks

lb90:
+1

itsam:
@Damar95 any news on this?  Thank you in advance!

Issue #5444, opened 2019-09-02
[V3] options:sideMenu:enabled: false overwritten to true on pan gesture


Royce:
### `sideMenu.enabled` flag overwritten when opening/closing menu.

On iOS, the flags to control whether the side menus are enabled or disabled is reset to `true` when a pan gesture (to open the menu) occurs.

### Steps to Reproduce / Code Snippets / Screenshots

With the following setup. Both menu's disabled...

```js
      Navigation.setRoot({
        root: {
          sideMenu: {
            id: 'RootLayoutId',
            options: {
              // The options are optional, but makes the bug more obvious.
              sideMenu: {
                left: { enabled: false },
                right: { enabled: false },
              },
            },
            left: { component: { name: 'MenuScreen' } },
            right: { component: { name: 'MenuScreen' } },
            center: { component: { name:'MainScreen' } },
          },
        },
      });
```

Then, triggered off a button or something, we explicitly enable one (and disable the other) of the two menu's:

```js
      Navigation.mergeOptions("RootLayoutId", {
        sideMenu: {
          left: { enabled: true }, 
          right: { enabled: false }  // ** Need to set this too to get the bug.
        }
      });
```

1. Swipe left (should open menu). Now close it again.
2. Swipe right (should not open menu)... **but it does.**

#### The problem.

The function `SideMenuPresenter.applyOptions` contains the following code, and is called when we `Navigation.setRoot`, but also when `SideMenuController.panGestureCallback` is triggered.

```objective-c
[sideMenuController side:MMDrawerSideLeft enabled:[withDefault.sideMenu.left.enabled getWithDefaultValue:YES]];
[sideMenuController side:MMDrawerSideRight enabled:[withDefault.sideMenu.right.enabled getWithDefaultValue:YES]];
```

The `getWithDefaultValue:YES` part means that enabled keeps getting set to true. No good!

#### Workaround / hack.

Wrap an `if` around these lines of code.
https://github.com/Royce/react-native-navigation/commit/8d2b20c1678e71ca84932d6461eb0beea12da978

But this may have side effects that I am not considering.

---
### Environment
* React Native Navigation version: 3.1.2 (also 3.0.0)
* React Native version: 0.60.5
* Platform(s) (iOS, Android, or both?): both, but iOS fixed by 3.1.2-snapshot.490

guyca:
@Royce Could you please try version `3.1.2-snapshot.490` and let me know if there are any issues?

Royce:
[...]

PR #5445, opened 2019-09-02
Stop overwriting enabled flag when opening flag.

Royce:
`RNNSideMenuPresenter.applyOptions` is called when
Navigation.setRoot() is called but also when the
menu's become visible (via a pan gesture) and it set *enabled:true*
for both menu's. ( In some circumstances, see #5444 )

This means that setting enabled:false doesn't
last very long.

This possibly naive PR removes the "defaulting to true" behaviour for the `enabled` flag to remove this bug.


guyca:
Hey @Royce 
Thanks for taking the time to look into this issue! I looked into it, Thanks for the reproduction, and I'll have a PR with a fix later today 👍

PR #5466, opened 2019-09-09
Add support for getLaunchArgs on Android

swabbass:
Add support for getLaunchArgs on Android

### DIFF
--- a/lib/android/app/src/main/java/com/reactnativenavigation/react/NavigationModule.java
+++ b/lib/android/app/src/main/java/com/reactnativenavigation/react/NavigationModule.java
@@ -73,6 +73,27 @@ public class NavigationModule extends ReactContextBaseJavaModule {
     }
 
     @ReactMethod
+    public void getLaunchArgs(String commandId,Promise promise) {
+        NavigationActivity activity = activity();
+        if (activity != null ) {
+            Intent intent = activity.getIntent();
+            if (intent != null) {
+                Bundle launchArgs = intent.getBundleExtra("launchArgs");
+                if(launchArgs != null){
+                    WritableMap writableMap = Arguments.fromBundle(launchArgs);
+                    promise.resolve(writableMap);
+                }else{
+                    promise.resolve(Arguments.createMap());
+                }
+            } else {
+                promise.resolve(Arguments.createMap());
+            }
+        } else {
+            promise.resolve(Arguments.createMap());
+        }
+    }
+    
+    @ReactMethod
     public void getConstants(Promise promise) {
         ReactApplicationContext ctx = getReactApplicationContext();
         WritableMap constants = Arguments.createMap();


lib/android/app/src/main/java/com/reactnativenavigation/react/NavigationModule.java
```diff
@@ -72,6 +73,27 @@ public String getName() {
         return NAME;
     }
 
+    @ReactMethod
+    public void getLaunchArgs(String commandId,Promise promise) {
+        NavigationActivity activity = activity();
```
guyca: NavigationModule should delegate to other entities. Please move all this logic to `LaunchArgsParser`,
Then `getLaunchArgs` will include a single line of code - `promise.resolve(LaunchArgsParser.parse(activity()))`

lib/android/app/src/main/java/com/reactnativenavigation/react/NavigationModule.java
```diff
@@ -72,6 +73,27 @@ public String getName() {
...
+        if (activity != null ) {
+            Intent intent = activity.getIntent();
+            if (intent != null) {
+                Bundle launchArgs = intent.getBundleExtra("launchArgs");
+                if(launchArgs != null){
+                    WritableMap writableMap = Arguments.fromBundle(launchArgs);
+                    promise.resolve(writableMap);
+                }else{
```
guyca: Seems like all of these else statements are identical and can be removed.
If you're unable to obtain the `launchArgs` - return `Arguments.createMap()`. Don't need do return it from every nested conditional.

lib/android/app/src/main/java/com/reactnativenavigation/react/NavigationModule.java
swabbass: done

lib/android/app/src/main/java/com/reactnativenavigation/react/NavigationModule.java
swabbass: refactored

guyca:
Looks great, could you please comment with examples on how to pass launch args to the application from cmd so I can later document this feature? Thanks

lib/android/app/src/main/java/com/reactnativenavigation/react/NavigationModule.java
```diff
@@ -72,6 +73,11 @@ public String getName() {
         return NAME;
     }
 
+    @ReactMethod
+    public void getLaunchArgs(String commandId,Promise promise) {
```
danilobuerger: nit pick: there is a space missing between `,` and `Promise`

guyca: Not anymore 🙈

PR #5485, opened 2019-09-15
return nil for NoColor default value

alon-ha:


guyca:
FIxed in latest version 👍

PR #5490, opened 2019-09-16
Font weight option support for iOS

yogevbd:
This feature allow using the iOS native system font weights. Providing system font names as `.SFUIText-Heavy` are no longer supported in iOS 13.

• Add `fontWeight` support for:
  • `topBar.title` option
  • `topBar.largeTitle` option
  • `topBar.subtitle` option
  • `bottomTab` option
  • `topBar.leftButtons` and `topBar.rightButtons` options

• Move all font creation to `RNNFontAttributesCreator`

### DIFF
--- a/docs/docs/styling.md
+++ b/docs/docs/styling.md
@@ -82,6 +82,7 @@ Navigation.mergeOptions(this.props.componentId, {
       fontSize: 14,
       color: 'red',
       fontFamily: 'Helvetica',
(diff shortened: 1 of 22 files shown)


exentrich:
Thank you very much! When this be merged to V3 branch?

manuhook:
It's merged but it's doesn't work :(

guyca:
@manuhook it's available in v3.3.0, perhaps you tried in a previous version.

manuhook:
You're right it's working on the playground but I can't get it work on my app. I will tell you what the problem is if I find out. Thank you.

PR #5517, opened 2019-09-25
V2: Add Navigation buttons

moonjava2005:
Dear all,
I'm using react-native-navigation for along time, and I recognize the problem is, when JS is lag, we touch the button on JS, then, call to native to push/pop/... on navigation. It makes the app lag
So, if we support the package "supports" that contains some native buttons. On each button, we can set the prop for action like "push", "pop",... In this case, when users touch the button, it run on native right now, no need to call from JS to native
Best regards,


guyca:
Hey @moonjava2005 
Seems like your PR isn't rebased. I'm closing it for now, please submit it again after rebasing your branch. Also make sure you're branching out of an up to date master 👍

PR #5518, opened 2019-09-25
Run detox e2e tests on iPhone 11

yogevbd:


package.json
```diff
@@ -122,7 +122,7 @@
       "ios.none": {
         "binaryPath": "playground/ios/DerivedData/playground/Build/Products/Debug-iphonesimulator/playground.app",
         "type": "ios.none",
-        "name": "iPhone X",
```
Evgeniy-boop: 11

Evgeniy-boop: "name": "iPhone 11",

PR #5534, opened 2019-10-01
Add modalInPresentation option (fixes memory leak in iOS13)

jfrolich:
This add modalInPresentation option. When this is selected (and modals are the default pageSheet presentationStyle), the modal will not close when swiped down.

To allow swiping down of modals the `presentationControllerDidDismiss` callback needs to be implemented. I tried to do this. But I couldn't get this to work properly. It will probably quite an easy fix for someone with more objective-c experience. (Here is a good blog post documenting the changes in iOS 13: https://medium.com/@hacknicity/view-controller-presentation-changes-in-ios-13-ac8c901ebc4e)


jfrolich:
Possible to add one more approval?

lib/src/interfaces/Options.ts
```diff
@@ -919,6 +919,12 @@ export interface Options {
    * #### (Android specific)
    */
   modalTransitionStyle?: OptionsModalTransitionStyle;
+  /**
+   * In case of a "pageSheet" modalPresentationStyle, can you dismiss the modal by swiping
+   * 
+   * #### (iOS specific)
+   */
+  modalInPresentation?: boolean;
```
guyca: Lets rename the property we expose through options to `dismissGesture`. `modalInPresentation` is coupled to iOS implementation and naming conventions.

lib/ios/RNNComponentPresenter.m
```diff
@@ -90,6 +93,12 @@ - (void)mergeOptions:(RNNNavigationOptions *)options resolvedOptions:(RNNNavigat
 	if (options.modalPresentationStyle.hasValue) {
 		[viewController setModalPresentationStyle:[RCTConvert UIModalPresentationStyle:options.modalPresentationStyle.get]];
 	}
+
+	if (options.modalInPresentation.hasValue) {
```
guyca: Instead of checking if `options.modalInPresentation.hasValue` - use `[withDefault.bottomTabs.visible getWithDefaultValue:NO]
Notice you should use the `withDefault` object instead of the actual `options passed to this function.

**ALSO:** please add this logic to `mergeOptions` as well 👍

jfrolich: This is the code for mergeOptions, and all options use `options.modalInPresentation.hasValue`. I use `withDefault` in above. Can you look again?

lib/src/interfaces/Options.ts
jfrolich: Ok renamed the option in the upcoming push! 👍

jfrolich: I actually renamed it to disableDismissGesture because the dismiss gesture is enabled by default. It also directly maps to the API (you can turn it **off** by setting the modalInPresentation option)

jfrolich:
@guyca Can you have a look at this?

jfrolich:
Anyone?

guyca:
Hey @jfrolich Thanks for the PR. Your pr looks great, but it requires xcode 11 to compile. We're migrating the project to xcode 11 now (dropping support of 10) and will add the missing prop.

yogevbd:
We added the ability to control `modalInPresentation` in: https://github.com/wix/react-native-navigation/pull/5709 and the memory leak is already fixed in master.
Thanks!

jfrolich:
Great!

PR #5547, opened 2019-10-04
[V3] Android: re-add support for navigationBarColor

mcuelenaere:


exentrich:
Would be great if you also add support for `navigationBar.style = 'light' | 'dark'`.
Use case is the same as `statusBar`.

Issue #5524, opened 2019-09-26
[V3] Android Custom TopBar Right Buttons Not Working Consistently


gpsugy:
### Issue Description

This is my first github issue post, so please excuse any poor quality aspects of it (feedback to improve is welcome).

I am not sure what is exactly causing this problem, but on Android, when using a custom component for the `rightButtons` field for a `topBar` (whether in a stack for `bottomTabs` or a `Navigation.push` screen), the button image renders, but is **sometimes not clickable**. The problem seems to occur non-deterministically, and trying various attempts to find patterns has been unsuccessful so far.

My project only started experiencing this issue upon upgrading from RN v59 to RN v60 and RNN v2 to RNN v3. This issue does not seem to apply with regular `rightButtons` with the `navigationButtonPressed` listener.

**At the very least, is anyone experiencing similar issues?** Any workaround solutions are welcome.

### Steps to Reproduce / Code Snippets / Screenshots

```js
// navigation.ts

rightButtons: [
  {
    id: 'customRightBtnId',
    component: {
      name: 'CustomTopBarBtn',
      passProps: {
        iconSrc: require('../assets/icons/custom-btn.png'),
      },
    },
  },
],
```

```js
// CustomTopBarBtn.tsx

const CustomTopBarBtn: React.SFC<Props> = (props) => {
  const { iconSrc } = props;

  return (
    <TouchableOpacity onPress={() => console.log('pressed btn')}>
      <View style={styles.container}>
        <Image source={iconSrc} resizeMode="contain" />
      </View>
    </TouchableOpacity>
  );
};

const styles = StyleSheet.create({
  container: {
    justifyContent: 'center',
    flexDirection: 'column'
    height: '100%',
    width: '100%',
    marginRight: 5,
    marginLeft: 5,
  },
});
```

---
### Environment
* React Native Navigation version: 3.2.0
* React Native version: 0.60.5
* Platform(s) (iOS, Android, or both?): Android
* Device info (Simulator/Device? OS version? Debug/Release?): All Android Devices

gpsugy:
Add whitespaces to JS code snippet

gpsugy:
Discovered some helpful information that may narrow down the cause.

Here is the style for the original post's code:

```js
const styles = StyleSheet.create({
  container: {
    justifyContent: 'center',
    flexDirection: 'column', // problematic
    height: '100%',
    width: '100%',
    marginRight: 5, // problematic
    marginLeft: 5, // problematic
  },
});
[...]

PR #5594, opened 2019-10-17
Fix title bar being too wide and overlapping right buttons on Android

ianwelsh:
This seems to fix https://github.com/wix/react-native-navigation/issues/5524


samparmenter:
Can confirm that this is still an issue and this fixes it.

PR #5601, opened 2019-10-22
Fix pageSheet modals causing memory leak when swiped

### LINKED ISSUE #5370: iOS13 pageSheet modal: no cleanup after dismissing by gesture
In iOS13 pageSheet modals can be dismissed using a swipe. When this is happening currently there is no cleanup (`componentWillUnmount` is not called), causing several edge cases.

yogevbd:
On iOS 13, when a `pageSheet` modal is dismissed with a swipe, the react view was not unmounted properly caused `componentWillUnmount` to never called.
This PR listen to this event and detach it from memory.

Closes #5370

### DIFF
--- a/lib/ios/RNNModalManager.m
+++ b/lib/ios/RNNModalManager.m
@@ -29,6 +29,10 @@
 	UIViewController* topVC = [self topPresentedVC];
 	topVC.definesPresentationContext = YES;
 	
+	if (viewController.presentationController) {
+		viewController.presentationController.delegate = self;
+	}
+	
 	RNNAnimationsTransitionDelegate* tr = [[RNNAnimationsTransitionDelegate alloc] initWithScreenTransition:viewController.resolveOptions.animations.showModal isDismiss:NO];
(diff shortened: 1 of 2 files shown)


jfrolich:
Awesome. As a non-ios dev tried to accomplish this, but couldn't get it to work. Just a few lines of code 🙈 Together with #5534 it fixes all ios13 modal issues.

lib/ios/RNNModalManager.m
```diff
@@ -105,6 +109,10 @@ - (void)dismissedModal:(UIViewController *)viewController {
 	[_delegate dismissedModal:viewController];
 }
 
+- (void)presentationControllerDidDismiss:(UIPresentationController *)presentationController {
+	[_presentedModals removeObject:presentationController.presentedViewController];
```
edvinasbartkus: We are missing here:
```
[_delegate dismissedModal:presentationController.presentedViewController];
```

Without it, `events().registerModalDismissedListener` is not triggered correctly.

Issue #5572, opened 2019-10-12
[v3] installation guide 0.60


rendomnet:
### Issue Description

Can we have installation guide for RNN v3+ and RN 0.60.
Is RNN support autolinking if so then why in this doc we linking it manualy https://wix.github.io/react-native-navigation/#/docs/Installing?

syedabuthahirm:
@rendomnet I think by default linking doesn't change `MyActivity.java` and `MainActivity.java`. In order to use React Native Navigation We have to modified those files.

syedabuthahirm:
Check this repo [repo](https://github.com/syedabuthahirm/react-native-navigation-install-guide) for fresh installation.

AndreiCalazans:
The way we are setting up android breaks auto-linking. 

Specially the logic around the getPackages. 

https://github.com/syedabuthahirm/react-native-navigation-install-guide/blob/master/android/app/src/main/java/com/rnninstallguide/MainApplication.java

https://github.com/facebook/react-native/blob/master/template/android/app/src/main/java/com/helloworld/MainApplication.java#L24

syedabuthahirm:
@AndreiCalazans Do you have any other native modules in your project?

AndreiCalazans:
I have a few like react-native-maps, react-native-reanimated, react-native-firebase. 

[error.txt](https://github.com/wix/react-native-navigation/files/3721704/error.txt)
<img width="341" alt="Screen Shot 2019-10-13 at 09 02 42" src="https://user-images.githubusercontent.com/20777666/66715310-5a75eb00-ed98-11e9-8f97-5e1222a69621.png">


To fix this error I have to manually link my native modules inside the MainApplication file.

AndreiCalazans:
Note, it looks like automatic linking works half way since I don't need to add the dependencies inside build.gradle

rendomnet:
what version of RNN should we use now? 3.2.0?

syedabuthahirm:
If you like to use RN 0.60 and higher versions use V3. i will try to update that repo as soon as possible when RNN & RN release new updates.

Gerlison:
> I have a few like react-native-maps, react-native-reanimated, react-native-firebase.
> 
> [error.txt](https://github.com/wix/react-native-navigation/files/3721704/error.txt)
[...]

PR #5603, opened 2019-10-23
Update docs for RN 0.60.0 and above

lb90:
Update docs with bits from #5572

### DIFF
--- a/docs/docs/Installing.md
+++ b/docs/docs/Installing.md
@@ -316,6 +316,8 @@ If you have any **react-native** related methods, you can safely delete them.
 
 This file is located in `android/app/src/main/java/com/<yourproject>/MainApplication.java`.
 
+*  For React Native < 0.60
+
 ```diff
 ...
 import android.app.Application;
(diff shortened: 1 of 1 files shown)


lb90:
Note: I can't test on MacOS and I don't know if the iOS instructions need update too

Also see: #5345, #5318

alpha0010:
I think you can further simplify.  My class is:
```java
package com.yourproject;

import com.facebook.react.PackageList;
import com.facebook.react.ReactNativeHost;
import com.facebook.react.ReactPackage;
import com.reactnativenavigation.NavigationApplication;
import com.reactnativenavigation.react.NavigationReactNativeHost;

import java.util.List;

public class MainApplication extends NavigationApplication {

    @Override
    protected ReactNativeHost createReactNativeHost() {
        return new NavigationReactNativeHost(this) {
            @Override
                return "index";
            }
        };
    }

    @Override
    public boolean isDebug() {
        return BuildConfig.DEBUG;
    }

    @Override
    public List<ReactPackage> createAdditionalReactPackages() {
        List<ReactPackage> packages = new PackageList(this).getPackages();
        // Add packages here that cannot yet be autolinked.
        return packages;
    }
}
```

lb90:
Thanks! I don't have much knowledge on that, but I'm going to update this with your code.

lb90:
Ready for review

docs/docs/Installing.md
```diff
@@ -369,6 +371,54 @@ import java.util.List;
...
+
+-import android.app.Application;
+-import android.content.Context;
+ import com.facebook.react.PackageList;
+-import com.facebook.react.ReactApplication;
+ import com.facebook.react.ReactNativeHost;
+ import com.facebook.react.ReactPackage;
+-import com.facebook.soloader.SoLoader;
+-import java.lang.reflect.InvocationTargetException;
+-import java.util.List;
```
zeevl: this import needs to stay

Issue #5492, opened 2019-09-17
[V3] BottomTabs Crashing on Android 0.60.5


FrederickEngelhardt:
### Issue Description

Android is crashing on bottomTab on react-native 0.60.5.

App will boot then crash immediately.

iOS is stable, no issues.

### Steps to Reproduce / Code Snippets / Screenshots

```
Navigation.events().registerAppLaunchedListener( () => {

	registerScreens(routes) // register your route
	Navigation.setRoot({
		root: {
			bottomTabs: {
				children: [
					{
						component: {
							id: 'App.Home',
							name: 'App.Home',
						}
					}
				]
			}
		}
	})
})

---
### Environment
* React Native Navigation version: 3.2.0
React Native version: 0.60.5
* Platform(s) (iOS, Android, or both?): Android
* Device info (Simulator/Device? OS version? Debug/Release?): Nexus6 API 29 | API 28

**Comment by Agungkes** [comment]
Please follow the documentation here [https://wix.github.io/react-native-navigation/#/docs/layout-types?id=bottomtabs](BottomTabs).

BottomTabs will crash on Android if it doesn't have an icon 
```
options: {
 bottomTab: {
    text: 'Tab 1',
    icon: require('../images/one.png')
 }
}
```

thanhpcc96:
same issue
`2019-10-02 17:09:23.921 29632-29632/com.p605 D/ReactNative: ReactInstanceManager.detachViewFromInstance()
2019-10-02 17:09:23.922 29632-29632/com.p605 D/ReactNative: ReactInstanceManager.onJSBundleLoadedFromServer()
2019-10-02 17:09:23.924 29632-29632/com.p605 D/ReactNative: ReactInstanceManager.recreateReactContextInBackground()
2019-10-02 17:09:23.924 29632-29632/com.p605 D/ReactNative: ReactInstanceManager.runCreateReactContextOnNewThread()
2019-10-02 17:09:23.924 29632-29632/com.p605 D/ReactNative: ReactInstanceManager.tearDownReactContext()
2019-10-02 17:09:23.927 29632-29632/com.p605 D/ReactNative: CatalystInstanceImpl.destroy() start
2019-10-02 17:09:23.928 29632-29632/com.p605 W/unknown:ReactNative: Packager connection already open, nooping.
2019-10-02 17:09:23.931 29632-30651/com.p605 D/ReactNative: ReactInstanceManager.createReactContext()
[...]

Issue #5435, opened 2019-08-31
[v3]: app crashes when setting bottomTabs via setRoot


abdullah-sr:
### Issue Description

```js             
registerScreens();
  Navigation.events().registerAppLaunchedListener(() => {
    Navigation.setRoot({
      root: {
        bottomTabs: {
          children: [
            {
              component: {
                name: HomeScreen.navID,
              },
            },
          ],
        },
      },
    });
  });
```
when setting bottomTabs, app crashes on lanuch.
Tested on Android. version 3.1.2

---
### Environment
* React Native Navigation version: 3.1.2
* React Native version: RN60
* Platform(s) (iOS, Android, or both?) Android, not sure about iOS
* Device info (Simulator/Device? OS version? Debug/Release?): physical device

abdullah-sr:
nvm. didn't set up the project properly because docs are out of date.

Andarius:
@abdullah-sr where did you find the correct way ? Can you share a working example ?

davidmarinangeli:
@abdullah-sr your solution could be useful to some other people!

Andarius:
@davidmarinangeli  this worked for me:

`    Navigation.setRoot({
      root: {
        bottomTabs: {
          options: {
            bottomTabs: {
              titleDisplayMode: 'alwaysShow'
            },
          },
          id: 'RootTabs',
          children: [
            {
              stack: {
                id: 'Stack1',
                children: [
                  {
                    component: {
                      name: Screens.SCREEN_1,
                      options: SCREEN_OPTIONS[Screens.SCREEN_1]
                    }
                  }
                ],
                options: {
                  bottomTab: {
                    text: 'Screen 1',
                    icon: require("../icons/icon1.png")
                  }
                }
              }
            },
            {
              stack: {
                id: 'Stack2',
                children: [
                  {
                    component: {
                      name: Screens.SCREEN_2,
                      options: SCREEN_OPTIONS[Screens.SCREEN_2]
                    }
                  }
                ],
                options: {
                  bottomTab: {
                    text: 'Screen 2',  
                    icon: require("../icons/icon2.png"),
                  }
                }
              }
            }
          ]
        }
      }
    });
`

Issue #5331, opened 2019-07-29
[v2.26.1] The app will run and shut down immediately on the screen where navigation has been implemented.


wonjong85:
### Issue Description

java.lang.NullPointerException: Attempt to invoke virtual method 'android.graphics.drawable.Drawable android.graphics.drawable.Drawable.mutate()' on a null object reference
        at com.aurelhubert.ahbottomnavigation.AHHelper.getTintDrawable(AHHelper.java:48)
        at com.aurelhubert.ahbottomnavigation.AHBottomNavigation.createSmallItems(AHBottomNavigation.java:568)
        at com.aurelhubert.ahbottomnavigation.AHBottomNavigation.createItems(AHBottomNavigation.java:295)
        at com.reactnativenavigation.views.BottomTabs.superCreateItems(BottomTabs.java:46)
        at com.reactnativenavigation.views.BottomTabs.createItems(BottomTabs.java:34)
        at com.reactnativenavigation.views.BottomTabs.enableItemsCreation(BottomTabs.java:23)
        at com.reactnativenavigation.viewcontrollers.bottomtabs.BottomTabsController.applyOptions(BottomTabsController.java:91)
        at com.reactnativenavigation.viewcontrollers.ViewController.onViewAppeared(ViewController.java:208)
        at com.reactnativenavigation.viewcontrollers.ChildController.onViewAppeared(ChildController.java:34)
        at com.reactnativenavigation.viewcontrollers.ViewController.onGlobalLayout(ViewController.java:261)
        at android.view.ViewTreeObserver.dispatchOnGlobalLayout(ViewTreeObserver.java:945)
        at android.view.ViewRootImpl.performTraversals(ViewRootImpl.java:2357)
        at android.view.ViewRootImpl.doTraversal(ViewRootImpl.java:1460)
        at android.view.ViewRootImpl$TraversalRunnable.run(ViewRootImpl.java:7183)
        at android.view.Choreographer$CallbackRecord.run(Choreographer.java:949)
        at android.view.Choreographer.doCallbacks(Choreographer.java:761)
        at android.view.Choreographer.doFrame(Choreographer.java:696)
        at android.view.Choreographer$FrameDisplayEventReceiver.run(Choreographer.java:935)
        at android.os.Handler.handleCallback(Handler.java:873)
        at android.os.Handler.dispatchMessage(Handler.java:99)
        at android.os.Looper.loop(Looper.java:193)
[...]

Issue #4990, opened 2019-04-17
[Android][V2] Navigation.setRoot crashes the application


Jazqa:
### Issue Description

Calling Navigation.setRoot often causes crashes (more-so on slow devices). Wrapping `await Navigation.setRoot` in a try catch doesn't help as the application crashes in the middle of the `Navigation.setRoot` call (missing a catch in RNN native code).

### Steps to Reproduce / Code Snippets / Screenshots

Have an app where you setRoot multiple times (e.g. login -> logout -> login -> logout) and repeat the setRoot until the application crashes. Very inconsistent.

---
### Environment
* React Native Navigation version: tested on 2.11.0, 2.13.0 and 2.17.0
* React Native version: tested on 0.59 and 0.59.4
* Platform(s) (iOS, Android, or both?): Android
* Device info (Simulator/Device? OS version? Debug/Release?): Multiple Android devices and emulators

guyca:
@Jazqa Please add a crash log.

Jazqa:
@guyca, there doesn't seem to be anything useful in the logs

```
04-17 15:57:16.134 20247 20247 I crash_dump32: obtaining output fd from tombstoned, type: kDebuggerdTombstone
04-17 15:57:16.135  1911  1911 I /system/bin/tombstoned: received crash request for pid 20025
04-17 15:57:16.138 20247 20247 I crash_dump32: performing dump of process 20025 (target tid = 20107)
04-17 15:57:16.138 20025 20107 I libc    : orig_dumpable: 1
04-17 15:57:16.138 20247 20247 F DEBUG   : *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
04-17 15:57:16.139 20247 20247 F DEBUG   : Build fingerprint: 'Nokia/Frontier_00WW/FRT:8.1.0/O11019/00WW_1_55B:user/release-keys'
04-17 15:57:16.139 20247 20247 F DEBUG   : Revision: '0'
04-17 15:57:16.139 20247 20247 F DEBUG   : ABI: 'arm'
04-17 15:57:16.139 20247 20247 F DEBUG   : pid: 20025, tid: 20107, name: mqt_js  >>> com.xxx.xxx.dev <<<
04-17 15:57:16.139 20247 20247 F DEBUG   : signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0xbbadbeef
[...]

PR #5619, opened 2019-10-31
Add docs for bottomTab requiring icon on Android

v1zix:
When creating a new project using this library with TypeScript I noticed that the definition for `icon` with `bottomTab` was optional, but I ran into a crash on Android when I omitted it.

It appears this issue has come up several times before:
https://github.com/wix/react-native-navigation/issues/4534
https://github.com/wix/react-native-navigation/issues/4684
https://github.com/wix/react-native-navigation/issues/4990
https://github.com/wix/react-native-navigation/issues/5331
https://github.com/wix/react-native-navigation/issues/5435
https://github.com/wix/react-native-navigation/issues/5492

From what I can tell when I investigated, this library is used on Android for bottom tabs: https://github.com/aurelhubert/ahbottomnavigation

I didn't see a way to create a bottom tab without an icon in that library and the icon is actually optional on iOS. Because of that it didn't seem like it made sense to update the TS definition.

### DIFF
--- a/docs/docs/layout-types.md
+++ b/docs/docs/layout-types.md
@@ -148,6 +148,8 @@ const bottomTabs = {
 }
 ```
 
+!> Note! On Android an `icon` is required when using `bottomTabs`.
+
 ### Selecting tabs programmatically
 
 The selected index is a style property which can be updated using the `mergeOptions` command. In order to update the BottomTabs options, Pass the BottomTabs `componentId` or the `componentId` of one of its children.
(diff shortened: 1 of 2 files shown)


StasD:
If an `icon` is required on Android only, that is, it is optional on iOS, then removing the question mark in typescript declaration (which does not differentiate between iOS and Android), is probably wrong...

N3TC4T:
making ```icon``` required will force devs to enter icon param in every ```mergeOptions``` for bottomTab update , like changing badge etc ...

I think it's better to revert this https://github.com/wix/react-native-navigation/pull/5619/commits/d3f34a8cd9e8006cd69772794003a19766d1010a

guyca:
Thanks for pointing this out @N3TC4T, I'll revert 👍

PR #5623, opened 2019-11-03
[BREAKING] Introduce updateProps command

guyca:
The `Navigation.updateProps` command allows to update props for a component registered with Navigation.registerComponent.
The updated props are handled by shouldComponentUpdate and componentDidUpdate lifecycle methods.

API
```js
Navigation.updateProps('myComponentId', {
  text: 'new value'
});
```

This commit builds upon the work done in 291f16177d2f67a474d3a980a503a85d0acf2b2a and is a breaking change.

### DIFF
--- a/lib/src/commands/Commands.test.ts
+++ b/lib/src/commands/Commands.test.ts
@@ -30,6 +30,7 @@ describe('Commands', () => {
     const optionsProcessor = instance(mockedOptionsProcessor) as OptionsProcessor;
 
     uut = new Commands(
+      mockedStore,
       instance(mockedNativeCommandsSender),
       new LayoutTreeParser(uniqueIdProvider),
       new LayoutTreeCrawler(instance(mockedStore), optionsProcessor),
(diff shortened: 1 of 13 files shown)


lib/src/commands/Commands.test.ts
```diff
@@ -131,6 +132,17 @@ describe('Commands', () => {
     });
   });
 
+  describe('updateProps', () => {
+    it('delegates to store', () => {
+      uut.updateProps('theComponentId', {someProp: 'someValue'});
+      verify(mockedStore.updateProps('theComponentId', deepEqual({someProp: 'someValue'})));
+    });
+    it('notifies commands observer', () => {
```
justtal: add empty line above

lib/src/components/Store.ts
```diff
@@ -6,10 +6,9 @@ export class Store {
   private propsById: Record<string, any> = {};
   private componentsInstancesById: Record<string, IWrappedComponent> = {};
 
-  setPropsForId(componentId: string, props: any) {
+  updateProps(componentId: string, props: any) {
     this.propsById[componentId] = props;
     const component = this.componentsInstancesById[componentId];
-
```
justtal: Don't you prefer empty line before if?

guyca: No. Ideally, functions should not contain more than 4-5 lines. In this case empty lines are not needed as it's easy to understand such a small function.

lib/src/commands/Commands.test.ts
guyca: 👌

gustavogialim:
Hi. Same problem, any solution?

Issue #5198, opened 2019-06-10
[V2][Android] Initial BottomTabs background color is always white 


danielang:
### Issue Description

The initial BottomTabs background color is always white.

### Steps to Reproduce / Code Snippets / Screenshots

`Navigation.setRoot` with a bottom tabs layout with an different background color than white:
```js
bottomTabs: {
    backgroundColor: '#333333',
}
```

### Solution/Fix
Due to the fact that AHBottomNavigation sets the `defaultBackgroundColor` (white) on item creation. RNN should set the `defaultBackgroundColor` in the `BottomTabsPresenter` with the corresponding setter `setDefaultBackgroundColor` in lines [80](https://github.com/wix/react-native-navigation/blob/master/lib/android/app/src/main/java/com/reactnativenavigation/presentation/BottomTabsPresenter.java#L80) and [141](https://github.com/wix/react-native-navigation/blob/master/lib/android/app/src/main/java/com/reactnativenavigation/presentation/BottomTabsPresenter.java#L141).

---
### Environment
* React Native Navigation version: 2.21.0
* React Native version: 0.59.9
* Platform(s) (iOS, Android, or both?): Android
* Device info (Simulator/Device? OS version? Debug/Release?): In every configuration

danielang:
This issue is not resolved!

guyca:
Should be fixed in 57eb0db7da6d69c784a9b0cd5672a563901738a9

PR #5625, opened 2019-11-03
Fixes issue #5198: Initial BottomTabs background color is always white

danielang:
See #5198 for details


guyca:
Should be fixed in 57eb0db7da6d69c784a9b0cd5672a563901738a9

PR #5636, opened 2019-11-07
Added peek-and-pop for stack root

incetarik:
Adds an ability to show preview (peek-and-pop) when presenting over a stack


guyca:
hey @incetarik 
Can you please provide more information about what this PR does?

incetarik:
@guyca Sure, when you set the navigation root as stack, or the center component as stack, such as

_The following example is from my project_,

```
Navigation.setRoot({
  root: {
    sideMenu: {
      center: {
        stack: {
          children: [...]
        },
      },
      right: {
        component: ...
      }
    }
  }
})
```

so that when we try to have peek-and-pop on iOS, as we set things up with `Navigation.TouchablePreview` and binding `onPress/onPressIn` events to a function that pushes component to Navigation stack with the option `preview`. In this setup, peek-and-pop (3D touch) is not working due to the stack controller actually. 

When you see the changes I did, I have just added one condition that if the view controller is a stack view controller, then get the current child (top of the stack) and continue as if it is applied on that component. Hence the preview would work. 

In this way, I could manage to show preview while having the main/root component as stack.

PR #5644, opened 2019-11-11
Playground for windows

Damar95:
Added script for the playground on windows.
On linus like systems we delete the lib/dist folder, i cant make it work with error catching if the folder does not exist.


ItsNoHax:
Why don't you just make the start command Windows friendly?

Damar95:
I dont have a Mac/Linux system and i can't test a possible build script which might be compatible with those plattforms.

PR #5667, opened 2019-11-20
Support centered bottomTab icons using bottomTabs.titleDisplayMode option

yogevbd:
Passing `titleDisplayMode: 'alwaysHide'` will both hide the bottomTab title text and center the bottomTab icon on iOS.
```
options: {
  bottomTabs: {
    titleDisplayMode: 'alwaysHide'
   }
}

### CODE REVIEW

File: lib/ios/UITabBarController+RNNOptions.m [hunk 1]
```diff
@@ -74,6 +75,12 @@ - (void)setTabBarVisible:(BOOL)visible animated:(BOOL)animated {
 	}
 }
 
+- (void)setTabItemImagesCentered:(BOOL)centered {
```
guyca: This setter doesn't set the value, instead it hides an imperative function - seems redundant as it doesn't handle the other case when `centered` is false.

lib/ios/UITabBar+centered.m
```diff
@@ -0,0 +1,53 @@
...
+		}
+		
+		if ([subView isKindOfClass:NSClassFromString(@"_UIBadgeView")]) {
+			UIView* imageView = [subView.superview viewWithTag:IMAGE_VIEW_TAG];
+			subView.frame = CGRectMake(subView.frame.origin.x, (imageView.frame.origin.y + imageView.frame.size.height * BADGE_OFFSET) - subView.frame.size.height / 2, subView.frame.size.width, subView.frame.size.height);
+		}
+	}
+}
+
++ (void)rnn_extendsUITabBarButtonLayoutSubviews {
```
guyca: I might be wrong here, but if all this function does is swizzle `layoutSubviews` - can we rename it to `swizzleLayoutSubviews`?

lib/ios/UITabBar+centered.m
```diff
@@ -0,0 +1,53 @@
...
+	dispatch_once(&onceToken, ^{
+		[[self class] rnn_extendsUITabBarButtonLayoutSubviews];
+	});
+	
+	for (UITabBarItem* item in self.items) {
+		item.title = nil;
+	}
+}
+
+- (void)rnn_UITabBarButton_layoutSubviews {
```
guyca: As this function is only used from the function below - moving it below it will improve readability as we'd be able to read this file top to bottom. Right now we start at the top, continue to the third function then finish at the middle of the file in the second function.

lib/ios/UITabBar+centered.m
```diff
@@ -0,0 +1,53 @@
+#import "UITabBar+centered.h"
```
guyca: These extension classes usually serve to add utility function on top of existing implementations. I personally think it's better to name these classes `...+utils.h`.
If in the future we'll need to add a few more helper functions, unrelated to centering icons we'll be able to addd them to this utils class.

lib/ios/RNNBottomTabsPresenter.m
```diff
@@ -9,6 +9,9 @@ - (void)applyOptionsOnInit:(RNNNavigationOptions *)options {
     UITabBarController *bottomTabs = self.boundViewController;
     RNNNavigationOptions *withDefault = [options withDefault:[self defaultOptions]];
     [bottomTabs setCurrentTabIndex:[withDefault.bottomTabs.currentTabIndex getWithDefaultValue:0]];
+	if ([[withDefault.bottomTabs.titleDisplayMode getWithDefaultValue:@"alwaysShow"] isEqualToString:@"alwaysHide"]) {
+		[bottomTabs setTabItemImagesCentered];
```
guyca: Sorry for nitpicking - but this public method isn't a setter. It should probably be named `centerTabItems` as it does a job - centre both icons and badge (not only images)

PR #5682, opened 2019-11-25
Dismiss presented modals and clear the component store on reloading

yogevbd:
This fix a leak reproduce on reloading react-native on iOS.

### DIFF
--- a/lib/ios/RNNBridgeManager.m
+++ b/lib/ios/RNNBridgeManager.m
@@ -15,6 +15,7 @@
 @property (nonatomic, strong, readwrite) RCTBridge *bridge;
 @property (nonatomic, strong, readwrite) RNNExternalComponentStore *store;
 @property (nonatomic, strong, readwrite) RNNReactComponentRegistry *componentRegistry;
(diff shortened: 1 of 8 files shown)


lib/ios/RNNBridgeManager.m
```diff
@@ -104,7 +109,10 @@ - (void)onJavaScriptLoaded {
 }
 
 - (void)onBridgeWillReload {
-	UIApplication.sharedApplication.delegate.window.rootViewController =  nil;
+	[_overlayManager dismissAllOverlays];
+	[_modalManager dismissAllModalsSynchronosly];
```
guyca: Unless we also have an async method of dismissing all overlays, I think `dismissAllModals` is a more suitable name :)

playground/src/screens/NavigationScreen.js
```diff
@@ -18,7 +18,13 @@ class NavigationScreen  extends React.Component {
     return {
       topBar: {
         title: {
-          text: 'Navigation'
+          component: {
```
guyca: w00ps

PR #5685, opened 2019-11-26
Fix topBar background color on iOS 13

yogevbd:


lib/ios/UINavigationBar+utils.h
```diff
@@ -0,0 +1,9 @@
+#import <UIKit/UIKit.h>
+
+@interface UINavigationBar (utils)
+
+- (void)rnn_setBackgroundColor:(UIColor *)color;
```
guyca: Why do we append the `rnn_` prefix to these methods? Can we avoid prefixing methods?

lib/ios/RNNStackController.m
```diff
@@ -52,35 +51,7 @@ - (UIViewController *)childViewControllerForStatusBarStyle {
...
-		}
-	} else {
-		UIView *transparentView = [self.navigationBar viewWithTag:TOP_BAR_TRANSPARENT_TAG];
-		if (transparentView){
-			[transparentView removeFromSuperview];
-		}
-		
-		self.navigationBar.barTintColor = nil;
-	}
+	[self.navigationBar rnn_setBackgroundColor:backgroundColor];
```
guyca: WOW! I love that you move this entire function to a category!

jfrolich:
Why is this dropping support for a transparent background color?

guyca:
@jfrolich W00ps, that's unintentional. We'll fix asap

PR #5749, opened 2019-12-09
Fix topBarHeight with UIModalPresentationPageSheet

### LINKED ISSUE #5730: Wrong topBarHeight with UIModalPresentationPageSheet
### Issue Description

Starting from iOS 13, `UIModalPresentationPageSheet` is not covering the entire screen, and therefore, the navigation bar doesn't have the same size.

<div>
  
  
</div>

But, when we call `Navigation.constants().topBarHeight`, is still returning `44`.

If we look at the View Hierarchy, it should be `56`.

| iOS 12 | iOS 13 |
| - | - |
|  |  |

https://github.com/wix/react-native-navigation/blob/65118b1ed28dd5a4636dabd5952c515d18a0b802/lib/ios/RNNStackController.m#L21-L23

is still returning `44`, while

https://github.com/wix/react-native-navigation/blob/65118b1ed28dd5a4636dabd5952c515d18a0b802/lib/ios/RNNStackController.m#L33-L35

is returning `UIModalPresentationPageSheet`. 🤔 

---
### Environment
* React Native Navigation version: 4.0.3
* Device info (Simulator/Device? OS version? Debug/Release?): Simulator, iPhone 11 iOS 13.2.1, Debug.

yogevbd:
Closes #5730

### DIFF
--- a/lib/ios/Constants.m
+++ b/lib/ios/Constants.m
@@ -8,7 +8,7 @@
 }
 
 + (CGFloat)topBarHeight {
-	return [UIApplication.sharedApplication.delegate.window.rootViewController getTopBarHeight];
+	return [RCTPresentedViewController() getTopBarHeight];
 }
 
 + (CGFloat)statusBarHeight {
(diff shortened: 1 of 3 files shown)


charpeni:
Thanks! ❤️

PR #5756, opened 2019-12-10
Refactor top bar options

yogevbd:
• Move topBar options logic to designated presenters
• Remove main window background color


lib/ios/TopBarAppearancePresenter.m
```diff
@@ -0,0 +1,79 @@
...
+@synthesize translucent = _translucent;
+
+- (instancetype)initWithNavigationController:(UINavigationController *)boundNavigationController {
+    self = [super initWithNavigationController:boundNavigationController];
+    _appearance = boundNavigationController.navigationBar.standardAppearance ?: [UINavigationBarAppearance new];
+    boundNavigationController.navigationBar.standardAppearance = _appearance;
+    return self;
+}
+
+- (void)updateBackgroundAppearance {
```
guyca: Lets move this function below the setters as they call it. That way we can read the class top down

lib/ios/TopBarAppearancePresenter.m
```diff
@@ -0,0 +1,79 @@
...
+        [_appearance configureWithOpaqueBackground];
+    }
+}
+
+- (void)setTranslucent:(BOOL)translucent {
+    _translucent = translucent;
+    [self updateBackgroundAppearance];
+}
+
+- (void)setTransparent:(BOOL)transparent {
```
guyca: Did we forget to set `_transparent` ?

PR #5784, opened 2019-12-17
Add bottomTabLongPressed event on iOS

N3TC4T:
Would be useful for someone to implement something like this :

1_dDxEuVRuEyXwzyzx2fctvQ


guyca:
Thanks for the PR @N3TC4T We're a bit unavailable this week, but will review it next week 👍

PR #5803, opened 2019-12-25
Add types for children property on TopTabs

### LINKED ISSUE #5570: TopTabs.children type is any[]
https://github.com/wix/react-native-navigation/blob/d6e228128b72c83907f131be3abcf823a5cf0740/lib/src/interfaces/Layout.ts#L132

aalises:
Fixes #5570 

Added type `LayoutTabsChildren[]` on the `children` property of `TopTabs` inside Layout (before was `any[]`. Also unified the naming of the export to be prefixed by `Layout + component name` as the other components exported from the layout.

### DIFF
--- a/lib/src/commands/LayoutTreeParser.ts
+++ b/lib/src/commands/LayoutTreeParser.ts
@@ -2,7 +2,7 @@ import { LayoutType } from './LayoutType';
 import { LayoutNode } from './LayoutTreeCrawler';
 import {
   Layout,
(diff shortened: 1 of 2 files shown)


guyca:
Thanks a lot @aalises. Just keep in mind TopTabs haven't been worked on in a long time and are a mess right now. We do plan on working on them again in the future.. but don't have a concrete timeline.

PR #5821, opened 2020-01-01
[V3] Fixed double click bug for stack

Damar95:
Hey,
this PR fixes a bug which causes a crash if a pushed screen has a ID defined, because its just added once to the stack but twice to the idStack.

I also added a example for it.


Damar95:
Stacks with ID`s getting also added multiple times to Modal Stacks, which resulting in the need to press the hardware back button multiple times to get back.

guyca:
This will be resolved by #5855

PR #5832, opened 2020-01-07
Added modalAttemptedToDismiss event with tests and docs

### LINKED ISSUE #5824: [V4] Support DidAttemptToDismiss for pageSheet modals
Hi, since 4.0.3 there is possibility to set `swipeToDismiss` under modal options. This is really nice addition, but also implementing DidAttemptToDismiss with a callback would be a great addition to this functionality.
I tried myself to implement this but I am not so good at objective C. But this might helped, in `RNNModalManager.m` there should be implementation of `- (void)presentationControllerDidAttemptToDismiss:(UIPresentationController *)presentationController` with getting calling callback passed to options -> modal -> onDismissAttempt or some other property name.

manicantic:
### Android Bottom Navigation Tabs

**_Related Issue_**
Closes #5824

**_Description_**
Since version 4.0.3 modal options are added along with swipeToDismiss flag. This event will trigger only when swipeToDismiss flag is set to false and user attempt to swipe down pageSheet modal. ( Which was use case in out app, so that was a motivation to implement this 😄 ).


guyca:
@manicantic e2e are failing due to changes in Modal screen. Can you please fix the tests so we can review and merge your PR?

yogevbd:
Awesome PR! thanks

PR #5846, opened 2020-01-13
Add componentName in modalDismissed event.

jinshin1013:
### The problem
This PR includes:
- Adding `componentName` to the `ModalDismissed` event.
- Updating the documentation on `ModalDismissed` event.


guyca:
hmm
I'm not sure about this PR, could you please elaborate on why you think it's needed?

jinshin1013:
Sorry I should have elaborated on the reasoning behind this PR. 

So to begin, `componentName` is currently being emitted to some of the navigation event listeners such as `registerComponentDidAppearListener` and `registerComponentDidDisappearListener`. And I've been personally using these listeners to log screen `appeared / disappeared` event to the Analytic tools. 

However we found the necessity in being able to specify `modalDismissed` event. We could get away with `registerComponentDidDisappearListener` but it does not specify whether it was a normal `pop` or `modal dismissal` as we may use one registered screen as 'normal' screen and a modal. 

What's your main concern about including `componentName` in the `modalDismissed` event or do you think it is redundant?

guyca:
@jinshin1013 Thanks for the explanation. The reason we didn't include componentName in the first place is because not every layout has a componentName. External components for example don't  have it. Moreover, it can be inferred from `componentDidAppear` event.

Lets implement this on Android as well so we can merge the PR. Relevant classes:
1. https://github.com/wix/react-native-navigation/blob/master/lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/modal/ModalStack.java#L66
2. https://github.com/wix/react-native-navigation/blob/master/lib/android/app/src/main/java/com/reactnativenavigation/react/events/EventEmitter.java#L72

And you'll probably want to add `public abstract String getCurrentComponentName();` to `ViewController.java`
which will return the `componentName` from `ComponentViewController` and

And in `ParentViewController` add something like
```java
@Override
    public String getCurrentComponentName() {
        return getCurrentChild().getCurrentComponentName();
    }
```

jinshin1013:
@guyca Yeah that makes sense!

> Moreover, it can be inferred from componentDidAppear event. 

Yeah it was more about being able to specify `modal dismissed` which cannot be inferred from `componentDidDisappear` unless you register the component with Modal specific name. 

I'll try to implement this on the Android as soon as I get some free time but thanks for the guidance!

lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/TitleBarButtonController.java
```diff
@@ -87,6 +87,11 @@ public void sendOnNavigationButtonPressed(String buttonId) {
         getView().sendOnNavigationButtonPressed(buttonId);
     }
 
+    @Override
+    public String getCurrentComponentName() {
```
jinshin1013: It was complaining that `getCurrentComponetName` needs to be implemented. Should this return its actual componentName?

lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/TitleBarReactViewController.java
```diff
@@ -44,6 +44,11 @@ public void sendOnNavigationButtonPressed(String buttonId) {
 
     }
 
+    @Override
+    public String getCurrentComponentName() {
```
jinshin1013: It was complaining that `getCurrentComponetName` needs to be implemented. Should this return `null` instead?

lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/externalcomponent/ExternalComponentViewController.java
```diff
@@ -87,6 +87,11 @@ public void applyBottomInset() {
         if (view != null) presenter.applyBottomInset(view, getBottomInset());
     }
 
+    @Override
+    public String getCurrentComponentName() {
```
jinshin1013: You mentioned that external component won't have component name. So should this return `null`?

jinshin1013:
@guyca I've tried to implement the Android part but having no prior exposure to Java made it really difficult. I've tried to follow your instruction as much as possible but Java was complaining about few things. I've got the tests to pass on Android and I've briefly tested on the Playground app and it successfully returned the `componentName` when dismissing modals.

iOS tests are failing with `The sandbox is not in sync with the Podfile.lock.`, I'll look into this soon.

jinshin1013:
iOS E2E test failure doesn't seem to be related to this PR. The test is passing on my local machine but it isn't on the Jenkins.

lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/TitleBarButtonController.java
guyca: Yes

lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/TitleBarReactViewController.java
guyca: Yes

lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/externalcomponent/ExternalComponentViewController.java
guyca: Return `externalComponent.name.get()` which is the name used to register the externalComponent with

guyca:
Hey @jinshin1013, Sorry for the delay. I commented, let me know if you have more questions :)

lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/TitleBarButtonController.java
jinshin1013: Now returning:

```java
@Override
public String getCurrentComponentName() {
  return button.component.name.get();
}
```

lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/TitleBarReactViewController.java
jinshin1013: Now returning:

```java
@Override
public String getCurrentComponentName() {
  return null;
}
```

lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/externalcomponent/ExternalComponentViewController.java
jinshin1013: Now returning:

```java
@Override
public String getCurrentComponentName() {
  return externalComponent.name.get();
}
```

jinshin1013:
@guyca Thanks for the feedback! Could you leave the PR when you get a chance please?

docs/docs/events.md
```diff
@@ -133,17 +133,45 @@ commandCompletedListener.remove();
...
 
 });
 ...
 // Unsubscribe
 modalDismissedListener.remove();
 ```
+|       Parameter         | Description |
+|:--------------------:|:-----|
+|**componentId**| Id of the disappearing component|
+|**componentName**|Registered name used when registering the component with `Navigation.registerComponent()`|
```
guyca: Lets change the description to "Registered name of the disappearing component". WDYT?

lib/src/interfaces/ComponentEvents.ts
```diff
@@ -20,7 +20,7 @@ export interface NavigationButtonPressedEvent extends ComponentEvent {
 }
 
 export interface ModalDismissedEvent extends ComponentEvent {
-  componentId: string;
+  componentName?: string;
```
guyca: Why is componentName optional? Seems like we're always emitting it

docs/docs/events.md
```diff
@@ -133,17 +133,45 @@ commandCompletedListener.remove();
 |**completionTime**|Timestamp when the command, and consecutive animations, completed.|
 
 ## registerModalDismissedListener
-Invoked when modal dismissed.
+
+Called each time this component (Modal) is dismissed.
```
guyca: I find this description a bit misleading as a modal can be dismissed only once. This method is invoked each time a modal (any modal) is dismissed. I think we should keep the current description as the new description is misleading

jinshin1013: Done!

docs/docs/events.md
jinshin1013: Oops. Bad copy and paste skill :( Fixed.

lib/src/interfaces/ComponentEvents.ts
jinshin1013: `dismissAllModals` do not emit `componentName`. I'm not sure whether it should return the name of the top modal component name? Let me know I can have a look.

guyca: [You can get the componentName of the top modal like we get the id of the top modal](https://github.com/wix/react-native-navigation/blob/master/lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/modal/ModalStack.java#L84)

guyca:
Hey @jinshin1013 
Seems like `push a screen to ensure its not there after reload` e2e test fails. Perhaps componentName is nil and you're trying to add it to a dictionary.

jinshin1013:
@guyca Hey yeah I am trying to make sure a parity between iOS and Android so `dismissAllModals` returns `componentName` on iOS too. [Look here](https://github.com/wix/react-native-navigation/pull/5846/files#diff-402a1f2cdd98bfe5ae49b8eecb54f391R381). 

If there is only one modal (stack modal), `UIViewController* lastViewController` is `RNNStackModal` which does not seem to have `name` attribute. 

I can have `dismissAllModals` on iOS to not return `componentName` would there be another way?

jinshin1013:
@guyca I know you are busy preparing V5 but could I get some guidance regarding the problem I mentioned above? Thanks 🙏

jinshin1013:
Alright, I've tried to rebase the branch with the updated master and seems like I screwed up something. I'm going to close this one and open a new one.

yogevbd:
@jinshin1013 Fixed the modals issue, can you please check? 😊

jinshin1013:
@yogevbd Legend! Tested on the Playground, `dismissAllModals` do return `componentName` successfully. Thanks for that fix man :) Will keep my eye on the CI tests and hopefully this PR can get merged soon.

PR #5859, opened 2020-01-20
Fix merge options leaks to next screen in stack

yogevbd:


playground/src/screens/PushedScreen.js
```diff
@@ -47,6 +48,7 @@ class PushedScreen extends React.Component {
 
   render() {
     const stackPosition = this.getStackPosition();
+    console.warn(this.props.previousScreenIds)
```
guyca: 🙈

lib/ios/UIViewController+LayoutProtocol.h
```diff
@@ -16,7 +16,7 @@ typedef void (^RNNReactViewReadyCompletionBlock)(void);
 
 - (void)mergeOptions:(RNNNavigationOptions *)options;
 
-- (void)mergeChildOptions:(RNNNavigationOptions *)options;
```
guyca: Isn't `mergeChildOptions` a better name for this function? It's explicit and it's clear how this function differs from the regular `mergeOptions` function.

PR #5877, opened 2020-01-26
Always set prefersLargeTitles to YES on iOS 11+

RobertPaul01:
If the value of `navigationBar.prefersLargeTitles` is changed while
using largeTitles, there will be a weird navigation animation when going
from a screen that uses largeTitles to a screen that doesn't use
largeTitles.

Instead of turning on/off the largeTitles with prefersLargeTitles, I
found it's better to change the `navigationItem.largeTitleDisplayMode`
in `UIViewController+RNNOptions.m`. Switching this value between
`UINavigationItemLargeTitleDisplayModeAlways` and
`UINavigationItemLargeTitleDisplayModeNever` will allow dynamic setting
of the largeTitles and not cause the animation problems.

Before:

After:

### DIFF
--- a/lib/ios/RNNStackPresenter.m
+++ b/lib/ios/RNNStackPresenter.m
@@ -62,7 +62,7 @@
     [_topBarPresenter applyOptions:withDefault.topBar];
     
     [stack setNavigationBarBlur:[withDefault.topBar.background.blur getWithDefaultValue:NO]];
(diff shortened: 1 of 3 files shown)


RobertPaul01:
Appears to be fixed with https://github.com/wix/react-native-navigation/pull/5984/files

PR #5919, opened 2020-02-10
Fix Bottom Inset when view is in fullscreen mode

ndillon1:
Currently when the status bar is hidden , the view mode is set to `View.SYSTEM_UI_FLAG_FULLSCREEN`. And the topInset is calculate to allow for the top section of the view to be immersive. 

According to google's documentation when the status bar is hidden we should also hide the navigation bar. (https://developer.android.com/training/system-ui/navigation)

This PR sorts to calculate the bottom inset to apply this rule to allow for the fullscreen mode to be fully immersive, rather than just take into account the top inset.

### DIFF
--- a/lib/android/app/src/main/java/com/reactnativenavigation/utils/NavigationBarUtils.java
+++ b/lib/android/app/src/main/java/com/reactnativenavigation/utils/NavigationBarUtils.java
@@ -17,7 +17,7 @@ public class NavigationBarUtils {
         final Resources resources = context.getResources();
         final int orientation = resources.getConfiguration().orientation;
         final int resourceId = resources.getIdentifier(orientation == Configuration.ORIENTATION_PORTRAIT ? "navigation_bar_height" : "navigation_bar_height_landscape", "dimen", "android");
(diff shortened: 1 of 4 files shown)


lib/android/app/src/main/java/com/reactnativenavigation/utils/NavigationBarUtils.java
```diff
@@ -0,0 +1,27 @@
...
+    }
+
+    public static int getNavigationBarHeight(Context context) {
+        if (navigationBarHeight > 0) {
+            return navigationBarHeight;
+        }
+        final Resources resources = context.getResources();
+        final int orientation = resources.getConfiguration().orientation;
+        final int resourceId = resources.getIdentifier(orientation == Configuration.ORIENTATION_PORTRAIT ? "navigation_bar_height" : "navigation_bar_height_landscape", "dimen", "android");
+        if(resourceId > 0) {
```
markhomoki:
```suggestion
        if (resourceId > 0) {
```

vavarun:
Would love to have this feature +1

PR #5926, opened 2020-02-13
Updated Android installation instructions

nikodemrafalski:
Changed step  `7. Update MainApplication.java` snippet to keep react-native auto-linking compatibility.


guyca:
Thanks @nikodemrafalski , I've updated the docs to support v5 and latest RN versions. If there are any more issues with compatibility, please let me know 👍

muhammadwafa:
As I heard that auto linking works in v5, but now v6 and still does not works

PR #5946, opened 2020-02-19
Layout background color

yogevbd:
* Modified `layout.backgroundColor` to control the layout parent background color.
* Add `layout.componentBackgroundColor` to control the component background color. (Until now on iOS, `layout.backgroundColor` controlled the component and not the parent.
* Remove drawBehind deprecation

### DIFF
--- a/lib/ios/RNNBottomTabsPresenter.m
+++ b/lib/ios/RNNBottomTabsPresenter.m
@@ -25,7 +25,7 @@
     [bottomTabs setTabBarHideShadow:[withDefault.bottomTabs.hideShadow getWithDefaultValue:NO]];
     [bottomTabs setTabBarStyle:[RCTConvert UIBarStyle:[withDefault.bottomTabs.barStyle getWithDefaultValue:@"default"]]];
     [bottomTabs setTabBarVisible:[withDefault.bottomTabs.visible getWithDefaultValue:YES] animated:[withDefault.bottomTabs.animate getWithDefaultValue:NO]];
-    [bottomTabs.view setBackgroundColor:withDefault.layout.backgroundColor.get];
+    [bottomTabs.view setBackgroundColor:[withDefault.layout.backgroundColor getWithDefaultValue:nil]];
 }
 
 - (void)mergeOptions:(RNNNavigationOptions *)options resolvedOptions:(RNNNavigationOptions *)currentOptions {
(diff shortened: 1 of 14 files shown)


lib/ios/RNNBottomTabsPresenter.m
```diff
@@ -25,6 +25,7 @@ - (void)applyOptions:(RNNNavigationOptions *)options {
     [bottomTabs setTabBarHideShadow:[withDefault.bottomTabs.hideShadow getWithDefaultValue:NO]];
     [bottomTabs setTabBarStyle:[RCTConvert UIBarStyle:[withDefault.bottomTabs.barStyle getWithDefaultValue:@"default"]]];
     [bottomTabs setTabBarVisible:[withDefault.bottomTabs.visible getWithDefaultValue:YES] animated:[withDefault.bottomTabs.animate getWithDefaultValue:NO]];
+    [bottomTabs.view setBackgroundColor:withDefault.layout.backgroundColor.get];
```
guyca: Shouldn't we apply background color with a default value?

guyca: getWithDefaultValue:nil like in SideMenu

lib/ios/RNNComponentPresenter.m
```diff
@@ -127,15 +127,15 @@ - (void)mergeOptions:(RNNNavigationOptions *)options resolvedOptions:(RNNNavigat
 	if (options.topBar.largeTitle.visible.hasValue) {
 		[viewController setTopBarPrefersLargeTitle:options.topBar.largeTitle.visible.get];
 	}
-
+    
+    if (options.layout.componentBackgroundColor.hasValue) {
+        [viewController setBackgroundColor:options.layout.componentBackgroundColor.get];
```
pie6k: Took me ~1hour after upgrading why my overlays arent transparent. It was espetially missleading as typescript file says

```
/**
     * Set background color only for components, helps reduce overdraw if background color is set in default options.
     * #### (Android specific)
     */
    componentBackgroundColor?: Color;
```

So I was ignoring it. Does it mean `backgroundColor` is now useless in component layout?

andrejcesen: Hey @pie6k, you're pretty much on point. As long as you use component layout on its own (without being wrapped in stack/bottomTabs/...), only componentBackgroundColor takes effect. But if you use `stack` for example, `backgroundColor` still changes the root UINavigationController background color, so this affects things like topBar (if you set its color to 'transparent', then `backgroundColor` will show through).
Hopefully this clears things up :)

PR #5963, opened 2020-02-25
Fix popOver presentation style to match enum.

harveyconnor:


guyca:
Seems like this doesn't compile 🤷‍♂

harveyconnor:
@guyca  Might me the other way round, there's an enum with popOver but in the native code it is popover.

PR #5970, opened 2020-02-26
Detach bottomTabPresenter from basePresenter

yogevbd:
* Detach BottomTabPresenter from BasePresenter
* Detach DotIndicatorPresenter from BottomTabPresenter


lib/ios/BottomTabPresenter.m
```diff
@@ -1,68 +1,62 @@
...
+    if (withDefault.bottomTab.badgeColor.hasValue) {
+        [child setTabBarItemBadgeColor:withDefault.bottomTab.badgeColor.get];
     }
 }
 
-- (void)applyOptionsOnWillMoveToParentViewController:(RNNNavigationOptions *)options {
+- (void)applyOptionsOnWillMoveToParentViewController:(RNNNavigationOptions *)options  child:(UIViewController *)child {
     RNNNavigationOptions * withDefault = [options withDefault:self.defaultOptions];
     
+    if (withDefault.bottomTab.badge.hasValue) {
```
guyca: Generally, `applyOptions` always sets a value - either an actual value defined by the user of the default value. Why isn't it the case here?

lib/ios/BottomTabPresenter.m
```diff
@@ -1,68 +1,62 @@
...
+        [child setTabBarItemBadge:withDefault.bottomTab.badge.get];
+    }
+    
+    if (withDefault.bottomTab.badgeColor.hasValue) {
+        [child setTabBarItemBadgeColor:withDefault.bottomTab.badgeColor.get];
+    }
+    
     if (withDefault.bottomTab.hasValue) {
```
guyca: Now that we've moved BottomTabPresenter to BottomTabsController - `bottomTab.hasValue` should always be true.

lib/ios/BottomTabsAppearancePresenter.m
```diff
@@ -2,6 +2,11 @@
 
 @implementation BottomTabsAppearancePresenter
 
+- (void)bindViewController:(UITabBarController *)boundViewController {
+    [super bindViewController:boundViewController];
+    boundViewController.tabBar.standardAppearance = [UITabBarAppearance new];
```
guyca: This should be moved to Presenter `init`

lib/ios/BottomTabPresenter.m
```diff
@@ -1,68 +1,49 @@
...
 }
 
-- (void)applyOptions:(RNNNavigationOptions *)options {
+- (void)applyOptions:(RNNNavigationOptions *)options child:(UIViewController *)child {
     RNNNavigationOptions * withDefault = [options withDefault:self.defaultOptions];
     
-    if (withDefault.bottomTab.badge.hasValue && [self.boundViewController.parentViewController isKindOfClass:[UITabBarController class]]) {
-        [self.boundViewController setTabBarItemBadge:withDefault.bottomTab.badge.get];
-    }
+    if (withDefault.bottomTab.badge.hasValue)
```
guyca: If applyOptions is called, and badge does not have value -should it be removed? How is this reflected now?

lib/ios/BottomTabPresenter.m
```diff
@@ -1,68 +1,49 @@
...
+        [child setTabBarItemBadge:withDefault.bottomTab.badge.get];
     
-    if (withDefault.bottomTab.badgeColor.hasValue && [self.boundViewController.parentViewController isKindOfClass:[UITabBarController class]]) {
-        [self.boundViewController setTabBarItemBadgeColor:withDefault.bottomTab.badgeColor.get];
-    }
+    if (withDefault.bottomTab.badgeColor.hasValue)
```
guyca: Lets make this conditional a one-liner and remove empty lines around it. This way all lines where options are actually applied are grouped together and mistakes are easy to point out because each line should basically look like all others.

lib/ios/RNNStackController.m
```diff
@@ -30,7 +30,7 @@ - (void)mergeChildOptions:(RNNNavigationOptions *)options child:(UIViewControlle
     if (child.isLastInStack) {
         [self.presenter mergeOptions:options resolvedOptions:self.resolveOptions];
     }
-    [self.parentViewController mergeChildOptions:options child:child];
+    [self.parentViewController mergeChildOptions:options child:self];
```
guyca: @yogevbd we're supposed to pass child here, right?

lib/ios/RNNBottomTabsController.m
```diff
@@ -1,24 +1,47 @@
...
+             dotIndicatorPresenter:(RNNDotIndicatorPresenter *)dotIndicatorPresenter
                       eventEmitter:(RNNEventEmitter *)eventEmitter
               childViewControllers:(NSArray *)childViewControllers
                 bottomTabsAttacher:(BottomTabsBaseAttacher *)bottomTabsAttacher {
-    self = [super initWithLayoutInfo:layoutInfo creator:creator options:options defaultOptions:defaultOptions presenter:presenter eventEmitter:eventEmitter childViewControllers:childViewControllers];
     _bottomTabsAttacher = bottomTabsAttacher;
+    _bottomTabPresenter = bottomTabPresenter;
+    _dotIndicatorPresenter = dotIndicatorPresenter;
+    self = [super initWithLayoutInfo:layoutInfo creator:creator options:options defaultOptions:defaultOptions presenter:presenter eventEmitter:eventEmitter childViewControllers:childViewControllers];
+    if (@available(iOS 13.0, *)) {
```
guyca: Why didn't we add this to `initWithLayoutInfo`? 🤔 😅

PR #5978, opened 2020-02-28
Support tabs without icons on Android

guyca:
This commit introduces parity with iOS which allows the user to show tabs without icons. While there's no real production use case for this feature, it's helpful when getting started with a new project.


bneigher:
@guyca can this be used to create "spacers" between bottomTab items?  Theoretical use case:

suppose we have a bottomTab layout like this:
[ A, B, C, D, E]

could this be used to accomplish this?
[A, B, _, C, D]

or does the _ have to be a screen?

guyca:
@bneigher Not really since the middle tab will still be selectable and require a screen.

Issue #5983, opened 2020-02-29
RN CLI Autolinking error


lqrz:
### Issue Description

I installed the library via `npx react-native link react-native-navigation` (and modifying the `minSdkVersion` in `andoid/build.gradle`).

When I run `npx react-native run-android`, the app is built and works fine, though I get the following error in the terminal:
```
error React Native CLI uses autolinking for native dependencies, but the following modules are linked manually: 
  - react-native-navigation (to unlink run: "react-native unlink react-native-navigation")
```

Since the lib has been linked manually,  I add an entry in the `react-native.config.js` file to prevent react native to try to autolink the lib, as in:
```
module.exports = {
    dependencies: {
        'react-native-navigation': {
            platforms: {
                android: null, // disable Android platform, other platforms will still autolink if provided
            },
        }
    }
  };
```

Now, the CLI error is not longer shown and the app is successfully built, but I get an error in the emulator:
```
TypeError: null is not an object (evaluating "this.nativeCommandsModule.setRoot()").
```

Which comes about at my first `Navigation.setRoot(...);` call:
```
[index.js]

const start = () => {
    Navigation.events().registerAppLaunchedListener(() => {
        registerScreens();
        gotoInitialising();
        // gotoLogin();
    });
};

start();
```
My question is, what extra step should I take to get the lib to work and to not have a React Native CLI error, at the same time.

### Environment
* React Native Navigation version: 6.0.1
* React Native version: 0.61.5
* Platform(s) (iOS, Android, or both?): Android
* Device info (Simulator/Device? OS version? Debug/Release?): Android emulator API 28 - (emulator version 29.2.1-5889189) - Debug build

jinshin1013:
So if the project is a fresh project, after installing the package:
```bash
yarn add react-native-navigation
``` 
All you have to run is `react-native link react-native-navigation` as from `V5` the team added the linking script which will modify all the necessary native changes required. 

That should set you up with the working project. However if you, for some reason, required to do this manually, you can reference my [boilerplate project](https://github.com/jinshin1013/rnn-boilerplate) and look at the `./android` files line by line. You will need to look at these files:

- `./android/settings.gradle`
- `./android/build.gradle`
- `./android/app/build.gradle`
- `./android/src/.../java/your/namespace/MainActivity|MainApplication.java`

Also no need to add `react-native-navigation` in your `react-native.config.js` file.

ida-n:
Hi Jin,

[...]

PR #5989, opened 2020-03-03
Fix autolink script

jinshin1013:
This PR fixes the autolink script including:
- Specifying the `minSdkVersion` for Android. Closes https://github.com/wix/react-native-navigation/issues/5983.
- Removing the RNN Pod added by the react-native link script. 

@guyca I've assumed the `minSdkVersion` will be `19` unless user specified higher. I've got the version from the doc, should this be updated to something higher?


jinshin1013:
Closed due to cleaning up the forked repo. Reopened in https://github.com/wix/react-native-navigation/pull/5990

PR #5990, opened 2020-03-03
Fix autolink script

jinshin1013:
This PR fixes the autolink script including:
- Specifying the minSdkVersion for Android. Closes https://github.com/wix/react-native-navigation/issues/5983.
- Removing the RNN Pod added by the react-native link script.

@guyca  I've assumed the minSdkVersion will be 19 unless user specified higher. I've got the version from the doc, should this be updated to something higher?

### DIFF
--- a/autolink/postlink/gradleLinker.js
+++ b/autolink/postlink/gradleLinker.js
@@ -4,6 +4,8 @@ var fs = require('fs');
 var { warnn, errorn, logn, infon, debugn } = require('./log');
 var { insertString } = require('./stringUtils');
 var DEFAULT_KOTLIN_VERSION = '1.3.61';
+// This should be the minSdkVersion required for RNN.
+var DEFAULT_MIN_SDK_VERSION = 19
 
 class GradleLinker {
   constructor() {
@@ -16,6 +18,7 @@ class GradleLinker {
       var contents = fs.readFileSync(this.gradlePath, 'utf8');
       contents = this._setKotlinVersion(contents);
       contents = this._setKotlinPluginDependency(contents);
+      contents = this._setMinSdkVersion(contents);
       fs.writeFileSync(this.gradlePath, contents);
       infon('Root build.gradle linked successfully!\n');
     } else {
@@ -55,6 +58,22 @@ class GradleLinker {
   }
 
   /**
+   * Check the current minSdkVersion specified and if it's lower than
+   * the required version, set it to the required version otherwise leave as it is.
+   */
+  _setMinSdkVersion(contents) {
+    var minSdkVersion = this._getMinSdkVersion(contents)
+    // If user entered minSdkVersion is lower than the default, set it to default.
+    if (minSdkVersion < DEFAULT_MIN_SDK_VERSION) {
+      debugn(`   Updating minSdkVersion to ${DEFAULT_MIN_SDK_VERSION}`)
+      return contents.replace(/minSdkVersion = \d*/, `minSdkVersion = ${DEFAULT_MIN_SDK_VERSION}`)
+    } 
+
+    debugn(`   Already specified minSdkVersion ${minSdkVersion}`)
+    return contents.replace(/minSdkVersion = \d*/, `minSdkVersion = ${minSdkVersion}`)
+  }
+
+  /**
    * @param { string } contents
    */
   _getKotlinVersion(contents) {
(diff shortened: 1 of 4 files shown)


autolink/postlink/gradleLinker.js
```diff
@@ -69,6 +88,21 @@ class GradleLinker {
     return `"${DEFAULT_KOTLIN_VERSION}"`;
   }
 
+  /**
+   * Get the minSdkVersion value.
+   * @param { string } contents
+   */
+  _getMinSdkVersion(contents) {
+    var minSdkVersion = contents.match(/minSdkVersion = (\d*)/)
```
guyca: We should account for zero or more spaces around to `=` symbol.
```suggestion
    var minSdkVersion = contents.match(/minSdkVersion\s=\s(\d*)/)
```

autolink/postlink/gradleLinker.js
```diff
@@ -54,6 +57,22 @@ class GradleLinker {
...
+  _setMinSdkVersion(contents) {
+    var minSdkVersion = this._getMinSdkVersion(contents)
+    // If user entered minSdkVersion is lower than the default, set it to default.
+    if (minSdkVersion < DEFAULT_MIN_SDK_VERSION) {
+      debugn(`   Updating minSdkVersion to ${DEFAULT_MIN_SDK_VERSION}`)
+      return contents.replace(/minSdkVersion = \d*/, `minSdkVersion = ${DEFAULT_MIN_SDK_VERSION}`)
+    } 
+
+    debugn(`   Already specified minSdkVersion ${minSdkVersion}`)
+    return contents.replace(/minSdkVersion = \d*/, `minSdkVersion = ${minSdkVersion}`)
```
guyca: lets account for white spaces around the equals sign
```suggestion
    return contents.replace(/minSdkVersion\s=\s\d*/, `minSdkVersion = ${minSdkVersion}`)
```

jinshin1013: Nice catch! Had to use `\s{0,}=\s{0,}` instead.

autolink/postlink/gradleLinker.js
jinshin1013: Had to use `\s{0,}=\s{0,}` instead.

jinshin1013:
@guyca Thanks for catching the regex issue, I'm still trying to get myself used to it 😓

> minSdk should be 19 as we don't support 18 and below.

Makes sense for minSdkVersion to be at least 19 and I also reckon most people would use 21 anyway. 

> Can you please explain why the change to podfile is needed?

I've found that whenever I run `react-native link react-native-navigation` it adds the pod definition for react-native-navigation. I'm using `@react-native-community/cli@3.2.0` so I don't think it's the cli problem. You should be able to reproduce this problem with fresh react-native project. Also as you mentioned I reckon it'd be good for the existing users who followed the previous installation step.

guyca:
> I've found that whenever I run `react-native link react-native-navigation` it adds the pod definition for react-native-navigation. 

Isn't this desired? I thought `react-native link` adds all native dependencies to the podfile.

autolink/postlink/gradleLinker.js
```diff
@@ -54,6 +57,22 @@ class GradleLinker {
...
+  /**
+   * Check the current minSdkVersion specified and if it's lower than
+   * the required version, set it to the required version otherwise leave as it is.
+   */
+  _setMinSdkVersion(contents) {
+    var minSdkVersion = this._getMinSdkVersion(contents)
+    // If user entered minSdkVersion is lower than the default, set it to default.
+    if (minSdkVersion < DEFAULT_MIN_SDK_VERSION) {
+      debugn(`   Updating minSdkVersion to ${DEFAULT_MIN_SDK_VERSION}`)
+      return contents.replace(/minSdkVersion\s{0,}=\s{0,}\d*/, `minSdkVersion = ${DEFAULT_MIN_SDK_VERSION}`)
```
guyca: You can use `\s*` to match zero or more occurrences of non-breaking white spaces. Up to you :)

jinshin1013:
> Isn't this desired? I thought react-native link adds all native dependencies to the podfile.

That is true for react-native version below 0.60. But I believe tools should assume users are on the "new" react-native and add somewhere in the documentation for the legacy linking (adding Pod definition). Thoughts?

autolink/postlink/gradleLinker.js
jinshin1013: Ah I see! Didn't even think about that hahaha. I'm off of my laptop so will probably just going to go with the current to save myself changing it :)

guyca: No problem 😁

PR #5999, opened 2020-03-04
Root layouts refactor

guyca:
This commit improves support for react-native-youtube library. This library uses the native Youtube player which requires that no visible views will be laid out on top of the player. This commit set visibility of Modal and Overlay containers to GONE when they are empty.


habovh:
@guyca @yogevbd It seems there still is an issue with RN-youtube and RNN.
Using a simple setup (one component root, no overlay, no extra HOCs, just directly registered component holding a single View and Youtube node from `react-native-youtube`) I'm still having the unauthorised overlay.
I'm more than happy to help debug this however I do not have much native Android knowledge.

RNN: 6.2
react-native-youtube: 2.0
react-native: 0.61.4

PR #6015, opened 2020-03-07
Add Windows compatibility in 4 node scripts.

mayconmesquita:


guyca:
Hey @mayconmesquita 👋 
Thanks for the PR, windows support has been requested by many users! ❤️ 
I would love it if we don't add another dev dependency to the library - can we use the `del` command instead of `rimraf`?

mayconmesquita:
Hi @guyca!
With the new commit, the scripts use 'del' instead 'rimraf'.
Thanks for requesting this! :)

PR #6038, opened 2020-03-14
Configure scrollEdgeAppearance with opaque background

RobertPaul01:
I noticed there is now an animation that occurs when pushing the first
view on a stack. This will happen with both:

`largeTitle: { visible: true }`
and
`largeTitle: { visible: false }`

It seems caused by the setting of the `scrollEdgeAppearance` background
without first configuring it using `configureWithOpaqueBackground`.

I also added checks for iOS 13 since these APIs require iOS 13.

I used this StackOverflow answer for reference: https://stackoverflow.com/a/57152709

### DIFF
--- a/lib/ios/TopBarAppearancePresenter.m
+++ b/lib/ios/TopBarAppearancePresenter.m
@@ -32,20 +32,17 @@
 }
 
 - (void)updateBackgroundAppearance {
-    if (@available(iOS 13.0, *)) {
+    if (self.transparent) {
+        [self.getAppearance configureWithTransparentBackground];
+        [self.getScrollEdgeAppearance configureWithTransparentBackground];
+    } else if (self.backgroundColor) {
         [self.getAppearance configureWithOpaqueBackground];
         [self.getScrollEdgeAppearance configureWithOpaqueBackground];
-
-        if (self.transparent) {
-            [self.getAppearance configureWithTransparentBackground];
-            [self.getScrollEdgeAppearance configureWithTransparentBackground];
-        } else if (self.backgroundColor) {
-            [self.getAppearance setBackgroundColor:self.backgroundColor];
-            [self.getScrollEdgeAppearance setBackgroundColor:self.backgroundColor];
-        } else if (self.translucent) {
-            [self.getAppearance configureWithDefaultBackground];
-            [self.getScrollEdgeAppearance configureWithDefaultBackground];
-        }
+        [self.getAppearance setBackgroundColor:self.backgroundColor];
+        [self.getScrollEdgeAppearance setBackgroundColor:self.backgroundColor];
+    } else if (self.translucent) {
+        [self.getAppearance configureWithDefaultBackground];
+        [self.getScrollEdgeAppearance configureWithDefaultBackground];
     }
 }
 
--- a/lib/ios/TopBarAppearancePresenter.m
+++ b/lib/ios/TopBarAppearancePresenter.m
@@ -32,18 +32,20 @@
 }
 
 - (void)updateBackgroundAppearance {
-    if (self.transparent) {
-        [self.getAppearance configureWithTransparentBackground];
-        [self.getScrollEdgeAppearance configureWithTransparentBackground];
-    } else if (self.backgroundColor) {
-        [self.getAppearance setBackgroundColor:self.backgroundColor];
-        [self.getScrollEdgeAppearance setBackgroundColor:self.backgroundColor];
-    } else if (self.translucent) {
-        [self.getAppearance configureWithDefaultBackground];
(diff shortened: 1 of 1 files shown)


guyca:
Hey @RobertPaul01 Can you please upload gifs of the problem before and after your change so we have a better understanding of what your PR fixes? Thanks.

RobertPaul01:
I found that it requires these options for `topBar` in a `bottomTabs` navigator root for it to happen
```
background: {
  color: ...
},
noBorder: true,
```

The animation looks like this for only the first push in the stack. There's like a flash of the top bar.
before

Here is after:
after

lib/ios/TopBarAppearancePresenter.m
```diff
@@ -32,18 +32,20 @@ - (void)setTransparent:(BOOL)transparent {
...
-        [self.getAppearance configureWithTransparentBackground];
-        [self.getScrollEdgeAppearance configureWithTransparentBackground];
-    } else if (self.backgroundColor) {
-        [self.getAppearance setBackgroundColor:self.backgroundColor];
-        [self.getScrollEdgeAppearance setBackgroundColor:self.backgroundColor];
-    } else if (self.translucent) {
-        [self.getAppearance configureWithDefaultBackground];
-        [self.getScrollEdgeAppearance configureWithDefaultBackground];
-    } else {
+    if (@available(iOS 13.0, *)) {
```
yogevbd: We do not need this here because this class will load on iOS 13 and above anyways. See here: https://github.com/wix/react-native-navigation/blob/master/lib/ios/TopBarPresenterCreator.m

lib/ios/TopBarAppearancePresenter.m
```diff
@@ -32,18 +32,20 @@ - (void)setTransparent:(BOOL)transparent {
...
         [self.getAppearance configureWithOpaqueBackground];
```
yogevbd: I think it make more sense to apply this two lines only when `self.backgroundColor` is available, to line 43

lib/ios/TopBarAppearancePresenter.m
```diff
@@ -32,18 +32,20 @@ - (void)setTransparent:(BOOL)transparent {
...
+            [self.getAppearance configureWithTransparentBackground];
+            [self.getScrollEdgeAppearance configureWithTransparentBackground];
+        } else if (self.backgroundColor) {
+            [self.getAppearance setBackgroundColor:self.backgroundColor];
+            [self.getScrollEdgeAppearance setBackgroundColor:self.backgroundColor];
+        } else if (self.translucent) {
+            [self.getAppearance configureWithDefaultBackground];
+            [self.getScrollEdgeAppearance configureWithDefaultBackground];
+        }
     }
```
yogevbd: I think it make more sense to apply `configureWithOpaqueBackground` only when self.backgroundColor is available. We also do need to do the iOS 13 check because this class will load on iOS 13 and above anyways. See here: https://github.com/wix/react-native-navigation/blob/master/lib/ios/TopBarPresenterCreator.m

```suggestion
    if (self.transparent) {
        [self.getAppearance configureWithTransparentBackground];
        [self.getScrollEdgeAppearance configureWithTransparentBackground];
    } else if (self.backgroundColor) {
        [self.getAppearance configureWithOpaqueBackground];
        [self.getScrollEdgeAppearance configureWithOpaqueBackground];
        [self.getAppearance setBackgroundColor:self.backgroundColor];
        [self.getScrollEdgeAppearance setBackgroundColor:self.backgroundColor];
    } else if (self.translucent) {
        [self.getAppearance configureWithDefaultBackground];
        [self.getScrollEdgeAppearance configureWithDefaultBackground];
    }
```

lib/ios/TopBarAppearancePresenter.m
```diff
@@ -36,14 +36,13 @@ - (void)updateBackgroundAppearance {
...
+        [self.getScrollEdgeAppearance configureWithOpaqueBackground];
         [self.getAppearance setBackgroundColor:self.backgroundColor];
         [self.getScrollEdgeAppearance setBackgroundColor:self.backgroundColor];
     } else if (self.translucent) {
         [self.getAppearance configureWithDefaultBackground];
         [self.getScrollEdgeAppearance configureWithDefaultBackground];
-    } else {
-        [self.getAppearance configureWithOpaqueBackground];
-        [self.getScrollEdgeAppearance configureWithOpaqueBackground];
     }
```
yogevbd:
```suggestion
    }  else {
        [self.getAppearance configureWithOpaqueBackground];
        [self.getScrollEdgeAppearance configureWithOpaqueBackground];
    }
```

yogevbd: We still need to fallback to the default background

PR #6054, opened 2020-03-18
Always resolve dismissAllModals promise

guyca:
When no modals are displayed and dismissAllModals is called, resolve the promise successfully. This regression was introduced in  #5991

### DIFF
--- a/lib/android/app/src/test/java/com/reactnativenavigation/viewcontrollers/modal/ModalStackTest.java
+++ b/lib/android/app/src/test/java/com/reactnativenavigation/viewcontrollers/modal/ModalStackTest.java
@@ -174,7 +174,7 @@ public class ModalStackTest extends BaseTest {
     }
 
     @Test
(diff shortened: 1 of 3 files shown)


lib/android/app/src/test/java/com/reactnativenavigation/viewcontrollers/modal/ModalStackTest.java
```diff
@@ -173,6 +173,13 @@ public void onSuccess(String childId) {
         verifyZeroInteractions(listener);
     }
 
+    @Test
+    public void dismissAllModals_rejectIfEmpty() {
```
ItsNoHax: Bit of a nit pick but maybe change to resolveIfEmpty since you no longer reject

guyca: Good point 😅

PR #6064, opened 2020-03-22
[WIP] Added using-mobx and using-react-context documentation.

jinshin1013:
This PR adds `Using MobX` and `Using React Context API` to the new documentation.

I'm definitely not the best person to write up a copy but hopefully I got the examples right 😅

### DIFF
--- a/website/docs/Installing.mdx
+++ b/website/docs/Installing.mdx
@@ -42,9 +42,11 @@ import App from "./App";
 +   Navigation.setRoot({
 +     root: {
 +       stack: {
-+         component: {
-+           name: 'com.myApp.WelcomeScreen'
-+         }
++         children: [
++           component: {
++             name: 'com.myApp.WelcomeScreen'
++           }
++         ]
 +       }
 +     }
 +  });
--- a/website/docs/using-react-context.mdx
+++ b/website/docs/using-react-context.mdx
@@ -1,0 +1,96 @@
+---
+id: using-react-context
+title: Using React Context API
+sidebar_label: Using React Context API
+---
+
+## React Context API 
+React Context API provides a easy way to pass data through the component tree without having to pass props down 
+manually at every level. You can find more about the Context API in [React documentation](https://reactjs.org/docs/context.html).
+
+You can use the Context API with React Native Navigation. In this example, we are going to create a screen which uses
+the Counter Context.
+
+## Create a Counter context
+***TODO***: Found this on the Playground app, will need to explain why we needed to do this way.
+
+```tsx
+// CounterContext.js
+import React from 'react
+
+const initialCounterState = {
+  count: 0
+}
+
+const counterContextWrapper = (component) => ({
+  ...initialCounterState,
+  increment: () => {
+    initialCounterState.count += 1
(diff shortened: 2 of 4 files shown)


website/docs/using-react-context.mdx
```diff
@@ -0,0 +1,96 @@
...
+
+## React Context API 
+React Context API provides a easy way to pass data through the component tree without having to pass props down 
+manually at every level. You can find more about the Context API in [React documentation](https://reactjs.org/docs/context.html).
+
+You can use the Context API with React Native Navigation. In this example, we are going to create a screen which uses
+the Counter Context.
+
+## Create a Counter context
+***TODO***: Found this on the Playground app, will need to explain why we needed to do this way.
```
jinshin1013: @guyca Would you be able to help me out on this one? The gist of the question is because RNN is not a single root structure, we have to work around to have the native `React Context API` working.

website/docs/Installing.mdx
```diff
@@ -42,9 +42,11 @@ import App from "./App";
...
 +     root: {
 +       stack: {
-+         component: {
-+           name: 'com.myApp.WelcomeScreen'
-+         }
++         children: [
++           component: {
++             name: 'com.myApp.WelcomeScreen'
++           }
++         ]
```
jinshin1013: Stack needs to have `children` unless we wanted to do:

```js
root: {
  component: {... }
}
```

guyca: Great catch! This is my mistake 😅

website/docs/using-react-context.mdx
```diff
@@ -0,0 +1,96 @@
...
+title: Using React Context API
+sidebar_label: Using React Context API
+---
+
+## React Context API 
+React Context API provides a easy way to pass data through the component tree without having to pass props down 
+manually at every level. You can find more about the Context API in [React documentation](https://reactjs.org/docs/context.html).
+
+You can use the Context API with React Native Navigation. In this example, we are going to create a screen which uses
+the Counter Context.
```
jinshin1013: > NOTE for myself, add a info:

As React Native Navigation screens are not part of the same component tree, updating the values in the shared context does not trigger a re-render across all screens. However you can still use the `React.Context` per  RNN screen component tree.

website/docs/using-react-context.mdx
jinshin1013: Will going to be adding this info: https://github.com/wix/react-native-navigation/pull/6064/files#r398980408

> The thing is, I'm not sure why someone would want to trigger a render across all of his screens as that seems very wasteful.

This would be more of a case of global state management where the same states are shared across multiple screens.

website/sidebars.js
```diff
@@ -12,6 +12,8 @@ module.exports = {
       'advance-navigation',
       'screen-lifecycle',
       'passing-data-to-components',
+      'using-react-context',
+      'using-mobx',
```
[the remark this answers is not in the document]
jinshin1013: Yep will do!

jinshin1013:
Closing as #6116 is the new PR to replace this.

PR #6084, opened 2020-03-29
Fix status bar visibility on iOS 13

yogevbd:
This PR fixes `statusBar.visible` option which stopped working on iOS 13.


anantanandgupta:
i guess that there is a similar issue with android ... see https://stackoverflow.com/questions/60893877/react-native-navigation-6-3-2-how-to-hide-top-statusbar-completely-in-androi

Issue #5960, opened 2020-02-24
[Android] TopBar Left/Right custom component buttons 


jinshin1013:
#### Left button
Left custom component button does not appear on Android. Also when coming back from a pushed screen the back button from the popped screen stays rather than showing the set left custom component button.

#### Right button
Right custom component button has no padding around it by default positioning it on the far right of the top bar on Android. Also when multiple right buttons are set, on iOS it automatically adds padding creating the right buttons nicely positioned whereas on Android it adds no padding positioning it right next to each other. 

### Steps to Reproduce / Code Snippets / Screenshots
I've created a reproduction on my [boilerplate](https://github.com/jinshin1013/rnn-boilerplate) under the branch called `reproduction/top-bar-padding`. Just need to checkout the branch and run:

```bash
git checkout reproduction/top-bar-padding

yarn
yarn start

# Open another tab on the terminal
yarn android # assuming you have an android device/emulator opened
```

The gif below demonstrates the positioning of the right custom button and the appearance of the left custom button on Android emulator. It also shows that after popping back to the screen with the custom left button still contain the back button from the popped screen.

![reproduction-gif](https://user-images.githubusercontent.com/17612922/75156297-8d27eb00-575d-11ea-8afa-77aba4ad5211.gif)

The screenshots below represent the positioning inconsistency between the Android and iOS platform. On iOS, paddings are applied to the custom buttons by default making it look nicer than on Android. 

![Desktop](https://user-images.githubusercontent.com/17612922/75156352-ad57aa00-575d-11ea-9786-963b9ba26dbb.png)

---
### Environment
* React Native Navigation version: 6.0.1
* React Native version: 61.5
* Platform(s) (iOS, Android, or both?): Android
* Device info (Simulator/Device? OS version? Debug/Release?): All

AlexeyMakDes:
+1
iOS - ok

guyca:
Hey @jinshin1013 :wave:
Android's Toolbar view doesn't support left buttons. It only supports a single icon left button, and even that is a hack.
This can be implemented by us in various ways, one naive approach which should work is to add another `TitleBar` to the `TopBar`, dedicated to the left buttons.

While I understand the importance of this task, I don't think I'll get to it anytime soon. If anyone would like to contribute it, please contact me on Discord.

guyca:
duplicates #3929

PR #6090, opened 2020-03-31
Refactor right buttons

guyca:
This pr started with an attempt to eliminate button flickering entirely when updating buttons which contain react components via mergeOptions. While the attempt wasn't 100% successful, I dit get some insights in the process.

* It's not possible to add buttons at specific indices. Buttons contain an order property which determines the order of the button in the TopBar. To somewhat overcome this limitation, we can let users control button order via options.

* When a right most component button is replaced with another component button, the rest of the buttons shift to the right since the newly added component isn't measured when it's added to the menu and its width is zero. 
We usually handle this situation with the `waitForRender` option but since buttons are measured with `MeasureSpec.UNSPECIFIED`, their dimensions are zero. Because of this, I've added options to set button dimensions.

* When updating buttons via mergeOptions, if a component button is already added to the menu with the same order we will not remove and added it again. This mitigates flickering in some situations.

* Textual button style properties were applied by traversing the view hierarchy and searching for the TextView corresponding to the button and updating its styles directly.
There was an inherent bug in this logic where if two buttons contained the same text, styles could have been applied to the wrong TextView. We now apply styles directly on the button using spans.


jinshin1013:
@guyca Is there any chance to also have a look at https://github.com/wix/react-native-navigation/issues/5960 🙏

guyca:
Refactor right buttons

This pr started with an attempt to eliminate button flickering entirely when updating buttons which contain react components via mergeOptions. While the attempt wasn't 100% successful, I dit get some insights in the process.

* It's not possible to add buttons at specific indices. Buttons contain an order property which determines the order of the button in the TopBar. To somewhat overcome this limitation, we can let users control button order via options.

* When a right most component button is replaced with another component button, the rest of the buttons shift to the right since the newly added component isn't measured when it's added to the menu and its width is zero. 
We usually handle this situation with the `waitForRender` option but since buttons are measured with `MeasureSpec.UNSPECIFIED`, their dimensions are zero. Because of this, I've added options to set button dimensions.

* When updating buttons via mergeOptions, if a component button is already added to the menu with the same order we will not remove and added it again. This mitigates flickering in some situations.

* Textual button style properties were applied by traversing the view hierarchy and searching for the TextView corresponding to the button and updating its styles directly.
There was an inherent bug in this logic where if two buttons contained the same text, styles could have been applied to the wrong TextView. We now apply styles directly on the button using spans.

PR #6093, opened 2020-03-31
Bottom tabs dont work

eladgel:
You have to use the next commit as a base to see that playground does not work well
 8dad9aed0926e22a00ddda40c90a87d8e18dd385 

image


guyca:
Hey @eladgel 
What do you mean by "using 8dad9ae as base"? That's a very old commit from over six months ago which we obviously don't support.

SInce there are no other reports of "bottom tabs not working" - I'm closing the PR. Seems like whatever issue you're experiencing is caused by a combination of older version of RNN, iOS 13.x/4 and newer Xcode.

PR #6127, opened 2020-04-15
Corrects pop api documentation

perottilds:


perottilds:
I'm dumb, only when after reading it once more I understood it meant the same thing. English is not my mother language, hence the confusion.

PR #6144, opened 2020-04-21
Implement rotate animation for shared element transition

guyca:
This commit adds support for animating rotate/rotationZ animation during shared element transition.

It also...
* Removes ClipBoundsAnimator - Not sure why I reintroduced it
* Sets pivotX and pivotY only if needed
* Removes "Preview" button in Navigation tab from Android


playground/src/screens/sharedElementTransition/CocktailDetailsScreen.js
```diff
@@ -100,6 +100,9 @@ const styles = StyleSheet.create({
     height: SIZE,
     width: SIZE,
     zIndex: 1,
+    // transform: [
+    //   { rotate: '45deg' }
+    // ],
```
yogevbd, line 105:
```suggestion
```

guyca: I actually left that for you, since it seems like rotate animation isn't implemented on iOS. Should we remove the comment?

PR #6147, opened 2020-04-21
Add 'visible' option in TopBar docs

tozaicevas:
I think `visible` option should be under Options/Stack/TopBar. Correct me if I am wrong.


guyca:
Thanks @tozaicevas !

PR #6154, opened 2020-04-22
update Docs setDefaultOptions

nywooz:
setDefaultOptions need to the inside registerAppLaunchedListener for it to work


guyca:
Hey @nywooz, this is actually incorrect. `setDefaultOptions` should be called before root is set. There's no need to call them from the `appLaunched` handler, though that would work just as fine.

PR #6182, opened 2020-05-01
fix: disable swipe gestures when sidemenu is close

mauriciopf:
# Summary

When the side menu is close, either leftSide or rightSide, the user is still able to open the menu with a **swipe gesture**, intercepting the touch event looking if **isDrawerVisible** fixes the issue.

When the drawer is visible we're not going to affect the normal behavior of the gestures.

Already tested for android and working as it suppose to.

| OS      | Implemented |
| ------- | :---------: |
| iOS     |   ❌     |
| Android |    ✅     |

## Checklist

- [ x] I have tested this on a device and a simulator

### DIFF
--- a/lib/android/app/src/main/java/com/reactnativenavigation/views/SideMenu.java
+++ b/lib/android/app/src/main/java/com/reactnativenavigation/views/SideMenu.java
@@ -2,6 +2,8 @@ package com.reactnativenavigation.views;
 
 import android.content.Context;
 import android.util.Log;
+import android.view.MotionEvent;
+import android.view.Gravity;
 
 import androidx.annotation.NonNull;
 import androidx.drawerlayout.widget.DrawerLayout;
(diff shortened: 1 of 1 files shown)


guyca:
Hey @mauriciopf Thanks for the PR!
Can you please explain how I can reproduce the problem in the playground app?

To disable SideMenu from being opened with swipe gesture, set `enabled: false` in [options](https://wix.github.io/react-native-navigation/api/options-sideMenuSide#enabled).

mauriciopf:
@guyca ``enabled: false`` what it does is that it ``setDrawerLockMode`` to ``DrawerLayout.LOCK_MODE_LOCKED_CLOSED`` but this doesnt seem to prevent the user to swipe and open the drawer manually (tested on a samsung galaxy S9). Im new to this repo so how can I run the playground?

mauriciopf:
@guyca just tested...

PR #6205, opened 2020-05-11
Color change with mergeOptions doesn't work

jfrolich:
Platform: android

If you push another screen and then go back the color is changed. (but it's not changed immediately after `mergeOptions`).


jfrolich:
I found a hint to the underlying problem. Buttons only change if you also change the ID.

Issue #6195, opened 2020-05-06
showModal and dismissModal animation on iOS


zabojad:
### Issue Description

I'm trying to change the  default animation used to show and dismiss some specific modal. By default on iOS, it slides from the bottom to the top of the screen. In one particular case, I want it to fade in / fade out.

So here is my code to open the modal:

	react.native.navigation.Navigation.showModal({
            component: {
                name: 'my-modal',
                options: {
                    overlay: {
                        interceptTouchOutside: true,
                    },
                    layout: {
                        backgroundColor: 'transparent',
                        componentBackgroundColor: 'transparent',
					},
    #if android
		            modalPresentationStyle: OverCurrentContext,
    #elseif ios
        		    modalPresentationStyle: OverFullScreen,
    #end
					animations: {
						showModal: {
							alpha: {
								from: .0,
								to: 1.0
							}
						},
						dismissModal: {
							alpha: {
								from: 1.0,
								to: .0
							}
						},
					}
                }
            }
        });

That works as expected on Android. On iOS, the animation setting seems to be ignored.

I've tried to change 

					animations: {
						showModal: {
							alpha: {
								from: .0,
								to: 1.0
							}
						},

by:

					animations: {
						showModal: {
							content: {
							    alpha: {
								from: .0,
								to: 1.0
							    }
							}
						},

But then it is even worse, the default "slide" animation is still there but it seems my component root react comp has been removed and only its children are visible.

Obviously, I am missing something. Could someone please explain me what em I doing wrong with these animation settings ?

---
### Environment
* React Native Navigation version: 3.7.0
* React Native version: 0.61.5
* Platform(s) (iOS, Android, or both?): iOS
* Device info (Simulator/Device? OS version? Debug/Release?): all

zabojad:
Hey @yogevbd ! I know you are one of the iOS guys of the project :). Would you have any suggestion for me for this issue? Thanks in advance.

Andarius:
I have a similar issue where the screen goes black when setting alpha (here on dismiss):

```
Navigation.showModal({
component: {
    name: screen,
    passProps: props,
    options: {
        animations: {
            dismissModal: {
                alpha: {
                    from: 1,
                    to: 0,
                    duration: 200,
                },
            },
        },
        modalPresentationStyle: 'overCurrentContext',
        layout: {
            backgroundColor: 'rgba(0, 0, 0, 0.1)',
[...]

PR #6238, opened 2020-05-21
Fixes black screen when using animation in `dismissModal`

simonmitchell:
Moves viewControllerForKey to use viewForKey with transitions where appropriate as discussed in this Stack Overflow post: https://stackoverflow.com/questions/24338700/from-view-controller-disappears-using-uiviewcontrollercontexttransitioning/25193675#25193675

This mainly resolves an issue when using animations with `dismissModal` that results in all views being removed and a black screen appearing as discussed in #6195 

The remaining `viewControllerForKey` calls should probably be removed too where their `view` property is being used, but this seems to resolve the bug for now at least!


guyca:
@simonmitchell Thanks for the PR! Could you please attach gif's of the problem, before and after the fix so we have a better understanding of the problem?

simonmitchell:
Sure thing! So this is the simple code I'm using:

```
Navigation.showModal({
  stack: {
    children: [
      {
        component: {
          name: 'ForestWatcher.MapWalkthrough',
          options: {
            animations: {
              dismissModal: {
                alpha: {
                  from: 1,
                  to: 0,
                  duration: 300
                }
              },
              showModal: {
                alpha: {
                  from: 0,
                  to: 1,
                  duration: 300
                }
              }
            },
            layout: {
              backgroundColor: 'transparent',
              componentBackgroundColor: 'rgba(0,0,0,0.74)'
            },
            screenBackgroundColor: 'rgba(0,0,0,0.74)',
            modalPresentationStyle: 'overCurrentContext'
          }
        }
      }
    ]
  }
});
```

Before:
Before

After:
After

RuairiOliver:
@simonmitchell thanks for this. I tried this out and the black screen is gone. Sometimes the screen flashes when it dismisses instead of smoothly animating out. Happens fairly regularly for me. 

showModal: {
            waitForRender: true,
            alpha: {
              from: 0,
              to: 1,
              duration: 300
            }
          },
          dismissModal: {
            alpha: {
              from: 1,
              to: 0,
              duration: 300
            }
          }

simonmitchell:
Hi @RuairiOliver I have noticed the same but I believe it’s unrelated to this as I could also observe it before this fix. The code around animating is pretty complex but I may have to look into that too once this work has gone to our project managers!

PR #6246, opened 2020-05-26
fixes badgeColor in ios 13+

DaveAdams88:
The badge colour was not changing from the default red in the tab bar item (when passing "badgeColor" to "bottomTab" options)


kanzitelli:
I have just had the same issue and the provided solution works for me! Thanks @DaveAdams88

PR #6249, opened 2020-05-27
Add NavigationFunctionComponent interface

elizabeth-dev:
- Added NavigationFunctionComponent interface

This interface can be used instead of React.FunctionComponent, fixing Typescript `Property 'options' does not exist on type 'FunctionComponent<{}>'` error when trying to declare static options on a functional component.

### DIFF
--- a/lib/src/interfaces/NavigationFunctionComponent.ts
+++ b/lib/src/interfaces/NavigationFunctionComponent.ts
@@ -1,0 +1,8 @@
+import React from 'react';
+import { NavigationComponentProps } from './NavigationComponentProps';
+import { Options } from './Options';
+
+export interface NavigationFunctionComponent<Props = {}>
+  extends React.FunctionComponent<Props & NavigationComponentProps> {
+    options?: Options;
+}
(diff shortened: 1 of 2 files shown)


elizabeth-dev:
(Commit authored with wrong email, force pushed to fix it. Sorry. https://www.youtube.com/watch?v=9ZlOhSt_qW0)

lib/src/interfaces/NavigationFunctionComponent.ts
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+import { NavigationComponentProps } from './NavigationComponentProps';
+import { Options } from './Options';
+
+export interface NavigationFunctionComponent<Props = {}>
+  extends React.FunctionComponent<Props & NavigationComponentProps> {
+    options?: Options;
```
jinshin1013, line 7: Correct me if I'm wrong but I believe `options` can also be a function with props. In my project, I have it typed as:

```tsx
options?: ((props: Props) => Options) | Options
```

So you could do:

```tsx
const Component: NavigationFunctionComponent<{ name: string }> = () => {}

Component.options = ({ name }) => ({
  topBar: {
    title: {
      text: name
    }
  }
})
```

elizabeth-dev: That's true, as stated in [the docs](https://wix.github.io/react-native-navigation/docs/passing-data-to-components#handling-passprops-in-static-options). I'm not used to defining options that way.

I've checked that `NavigationComponentProps` are not passed to options functions. Interface fixed and rebased onto master.

jinshin1013:
@guyca this looks good to me. What do you think?

guyca:
Yup, looks great! Thanks @elizabeth-dev 👍

PR #6256, opened 2020-05-30
[Docs] Modify functional components event listener example

elizabeth-dev:
Using `Navigation.events().bindComponent()` to register event listeners in functional components leads to typing errors when using Typescript, as bindComponent method requires a `React.Component<any>` as first argument.

With registerComponentListener we achieve the exact same effect, but providing more appropiate typings for for its parameters, using `NavigationComponentListener` for the listener, and making componentId required.


guyca:
Thanks for the PR!!

PR #6269, opened 2020-06-04
Replace deprecated link process

eduardopelitti:
Running `react-native link` will be deprecated in the following major versions of rn-cli.

React Native Navigation recommends the `react-native link` command to install the library, although it already implements Autolinking. These causes the library to make unnecessary modifications to the project it's installing to, which are already handled in Autolinking. 

However, since manual steps need to be completed, we are benefiting from the `"postlink"` hook in the `link` command which allows to run any set of scripts after the link has completed, in our case the node script `./autolink/postlink/run.js`.

I propose using the `bin` field from npm's package.json to perform these tasks from a single command and without using react-native link. I named it "rnn-link" and it can be run via `npx rnn-link` from the root of the project right after `yarn add react-native-navigation`.

I've updated the docs to reflect this, plus some minor changes I feel are necessary, after some work with these scripts and the install process. I still feel some parts are confusing, specially the iOS manual link (the hardcore way).

Feel free to propose any changes!

This was developed with the support of @underscopeio 🥳

### DIFF
--- a/autolink/postlink/run.js
+++ b/autolink/postlink/run.js
@@ -1,3 +1,5 @@
+#!/usr/bin/env node
+
 // @ts-check
 var { infon } = require("./log");
 var postLinkAndroid = require('./postLinkAndroid');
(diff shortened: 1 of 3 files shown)


guyca:
So happy we sorted this out! Thanks @eduardopelitti ! 🙇

Issue #6262, opened 2020-06-03
Unnecessary deep cloning of props causes comparisons with === to fail


HectorRicardo:
### Issue Description

Apparently, when using Navigation.updateProps(), the props passed as parameter are deep cloned. This causes shallow comparisons with '===' to fail inside the children components. I don't expect this behavior, because React does not deep clone properties automatically for you (see example).

### Steps to Reproduce / Code Snippets / Screenshots

Repository: [https://github.com/HectorRicardo/deepCloneBug](https://github.com/HectorRicardo/deepCloneBug)

index.js:
```js
import {Navigation} from 'react-native-navigation';

import App from './App';

Navigation.registerComponent('com.myApp.WelcomeScreen', () => App);

Navigation.events().registerAppLaunchedListener(() => {
  Navigation.setRoot({
    root: {
      stack: {
        children: [
          {
            component: {
              id: 'com.myApp.WelcomeScreen',
              name: 'com.myApp.WelcomeScreen',
              passProps: {
                onSelector2ItemSelect: function onSelector2ItemSelect(item) {
                  Navigation.updateProps('com.myApp.WelcomeScreen', {
                    selectedItemSelector2: item,
                    onSelector2ItemSelect,
                  });
                },
              },
            },
          },
        ],
      },
    },
  });
});
```

App.js
```js
import React, {useState} from 'react';
import {Button, Text, View, StyleSheet} from 'react-native';

const itemsList = [{name: 'item1'}, {name: 'item2'}];

const App = ({selectedItemSelector2, onSelector2ItemSelect}) => {
  const [selectedItemSelector1, setSelectedItemOfSelector1] = useState();
  return (
    <View style={styles.root}>
      <Selector
        description="this selector works"
        items={itemsList}
        selectedItem={selectedItemSelector1}
        onSelectItem={setSelectedItemOfSelector1}
      />
      <Selector
        description="this selector doesn't work"
        items={itemsList}
        selectedItem={selectedItemSelector2}
        onSelectItem={onSelector2ItemSelect}
      />
    </View>
  );
};

const Selector = ({description, items, selectedItem, onSelectItem}) => (
  <View>
    <Text style={styles.description}>{description}</Text>
    <Text style={styles.description}>
      Selected item's name: {selectedItem?.name}
    </Text>
    {items.map(item => (
      <View
        key={item.name}
        style={[
          styles.itemContainer,
          selectedItem === item && styles.selectedItem, // <------ comparison with === fails
[...]

PR #6280, opened 2020-06-09
Fix updateProps creating new props object

jinshin1013:
This PR closes https://github.com/wix/react-native-navigation/issues/6262.


jfrolich:
Be very carefull with this. I am pretty sure this cloneDeep is there for a reason (I did touch this code some time ago but I don't know the specifics anymore).

jfrolich:
Ah yes some operations are mutable such as here: https://github.com/wix/react-native-navigation/blob/9ef61a9a1357312208463ef149a7951931043d1e/lib/src/commands/OptionsProcessor.ts. So you need to clone the tree first so you can mutate the new tree.

jfrolich:
Actually I think you can remove the clone deep specifically for the `updateProps` function, as there doesn't seem to be mutation involved.

jinshin1013:
@jfrolich Hmm yeah, looks like some of the other commands are accessing the cloned object. Yeah I'll change it to only use `clone` for `updateProps`.

jinshin1013:
Hmmm, I'm not sure if it is worth changing the current behaviour as there are other ways for the consumer to be checking the equality of the objects.

jfrolich:
I think for updateProps, not cloning and just passing the object is probably better. It will allow creating a memoized component.

guyca:
@jinshin1013 I tend to agree with @jfrolich on this one. Cloning props here is probably my mistake. We clone *options* passed to other commands but in this case, cloning is actually a mistake as it prevents a basic optimisation with the `===` equality check.

What do you say we reopen the PR and remove all clone/cloneDeep from the `updateProps` command?

PR #6317, opened 2020-06-17
Fix SearchBar false render

### LINKED ISSUE #4443: [v2] searchBar: false render a searchBar on iOS 11+
### Issue Description

`searchBar: false` renders a search bar on iOS 11+.

Create a component with the following definition.

```
{
  component: {
  id: "a.component",
  name: "a.component",
  options: {
    topBar: {
      searchBar: false,
      title: {
        text: "Search Bar is Visible,
      },
    },
  },
}
```

Without `searchBar: false` there is no search bar, with `searchBar: any`, a wild search bar appears.

This is due to https://github.com/wix/react-native-navigation/blob/b0c8113243de2043b98442dd485c04b0422aa213/lib/ios/RNNViewControllerPresenter.m#L31 just checking if there is a value instead of making sure the value is truthy.

I wanted to do a PR, but it's also present here https://github.com/wix/react-native-navigation/blob/b0c8113243de2043b98442dd485c04b0422aa213/lib/ios/RNNViewControllerPresenter.m#L67 and I wasn't sure how to handle removing the search bar.

---
### Environment
* React Native Navigation version: 2.1.3-snapshot.50
* React Native version: 0.56
* Device info (Simulator/Device? OS version? Debug/Release?): iOS 11+

jinshin1013:
This PR fixes #4443 

The problem was that `withDefault.topBar.searchBar.hasValue` was true when setting `topBar.searchBar = false`.

### DIFF
--- a/lib/ios/RNNComponentPresenter.m
+++ b/lib/ios/RNNComponentPresenter.m
@@ -12,15 +12,15 @@
 }
 
 - (instancetype)initWithComponentRegistry:(RNNReactComponentRegistry *)componentRegistry defaultOptions:(RNNNavigationOptions *)defaultOptions {
-	self = [super initWithComponentRegistry:componentRegistry defaultOptions:defaultOptions];
-	_topBarTitlePresenter = [[TopBarTitlePresenter alloc] initWithComponentRegistry:componentRegistry defaultOptions:defaultOptions];
-	return self;
+    self = [super initWithComponentRegistry:componentRegistry defaultOptions:defaultOptions];
+    _topBarTitlePresenter = [[TopBarTitlePresenter alloc] initWithComponentRegistry:componentRegistry defaultOptions:defaultOptions];
+    return self;
 }
 
 - (void)bindViewController:(id)boundViewController {
-	[super bindViewController:boundViewController];
-	[_topBarTitlePresenter bindViewController:boundViewController];
-	_navigationButtons = [[RNNNavigationButtons alloc] initWithViewController:boundViewController componentRegistry:self.componentRegistry];
+    [super bindViewController:boundViewController];
+    [_topBarTitlePresenter bindViewController:boundViewController];
+    _navigationButtons = [[RNNNavigationButtons alloc] initWithViewController:boundViewController componentRegistry:self.componentRegistry];
 }
 
 - (void)componentDidAppear {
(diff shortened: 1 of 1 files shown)


lib/ios/RNNComponentPresenter.m
```diff
@@ -43,37 +43,37 @@ - (void)componentDidDisappear {
...
-	if (withDefault.topBar.searchBar.hasValue) {
-		BOOL hideNavBarOnFocusSearchBar = YES;
-		if (withDefault.topBar.hideNavBarOnFocusSearchBar.hasValue) {
-			hideNavBarOnFocusSearchBar = withDefault.topBar.hideNavBarOnFocusSearchBar.get;
-		}
-		[viewController setSearchBarWithPlaceholder:[withDefault.topBar.searchBarPlaceholder getWithDefaultValue:@""] hideNavBarOnFocusSearchBar:hideNavBarOnFocusSearchBar];
-	}
-	
-	[_topBarTitlePresenter applyOptions:withDefault.topBar];
+    if (withDefault.topBar.searchBar.hasValue && withDefault.topBar.searchBar.get) {
```
jinshin1013, line 68: Is it safe to check it this way? Is there a better way to make sure `topBar.searchBar` option was provided and it's indeed `true`?

lib/ios/RNNComponentPresenter.m
```diff
@@ -92,72 +92,72 @@ - (void)applyOptionsOnInit:(RNNNavigationOptions *)options {
...
-		if (options.topBar.hideNavBarOnFocusSearchBar.hasValue) {
-			hideNavBarOnFocusSearchBar = options.topBar.hideNavBarOnFocusSearchBar.get;
-		}
-		[viewController setSearchBarWithPlaceholder:[options.topBar.searchBarPlaceholder getWithDefaultValue:@""] hideNavBarOnFocusSearchBar:hideNavBarOnFocusSearchBar];
-	}
+    if (options.backgroundImage.hasValue) {
+        [viewController setBackgroundImage:options.backgroundImage.get];
+    }
+    
+    if (options.topBar.searchBar.hasValue && options.topBar.searchBar.get) {
```
jinshin1013, line 102: Is it safe to check it this way? Is there a better way to make sure topBar.searchBar option was provided and it's indeed `true`?

jinshin1013:
@yogevbd Wasn't aware of tht syntax! Looks good to me :)

PR #6323, opened 2020-06-21
Upgrade React Native 0.62.2

jinshin1013:
This PR upgrades the playground app to use the latest react native version. Flipper integration was intentionally unimplemented as the `Podfile` uses `use_framework!` which will conflict with the Flipper integration.


guyca:
Seems like there's a problem with the build flavor on Android, I'll look into it

PR #6325, opened 2020-06-21
Initial Typescript for Playground app

### LINKED ISSUE #6343: Export LayoutOrientation in Options.ts
### LayoutOrientation type isn't exported in Options.ts

Converting the Playground app to TypeScript revealed a types and interfaces which are mistakenly not exported in the Options.ts file.

In the playground app, try to change orientation. TS compiler throws an error

---
### Environment
* Platform(s) (iOS, Android, or both?): TS

jinshin1013:
## The problem

The aim of the PR is NOT to convert the entire Playground app to Typescript but rather to discuss and settle on the code style/convention prior to work on the full sweep. Some of the things that I wanted to get some feedback on are:

**Lint and Prettier**
In this PR, I've replaced `tslint` with `eslint` as `tslint` is being deprecated. I've also added `prettier`. I've only added default rules for now as the existing tslint config were mostly default (I think 🤔 ). But let me know if other rules need to be added.

**Component import/export style**
Currently the Playground uses the following convention to import/export components:

```tsx
// Importing 
const { Button } = require('./Button');

// Exporting
module.exports = Button;
```

Although I understand the intention, I think it's important to make sure the Playground app uses the same convention as what the official React Native documentation displays for better familiarity:

```tsx
// Importing
import { Button } from './Button'; // Named import
// or
import Button from './Button'; // Default import

// Exporting
export { Button } // Named export
// or 
export default Button // Default export
```

Another decision to make is on whether the codebase is going to be using `default export` or `named export`. 

**Function component declaration**
This is probably somewhat controversial (there was big Twitter thread about this 😅 ). There are a number of ways to define a Function component:

```tsx
// 1. Using React.FC
const Button: React.FC<ButtonProps> = (props) => { ... }

// 2. Regular Prop type as parameter
const Button = (props: ButtonProps) => { ... }

// 3. Using Function declaration
function Button(props: ButtonProps) { ... }
```

All three options are valid in their own right and can be used interchangeably but I reckon it'd be good to settle on one convention to keep the codebase consistent throughout.

---

### Decisions made
**Lint and Prettier**
It was decided to use the default `eslint` and `prettier` config and will add different rules as we fit later.

**Component import/export style**
It was decided to use the default import/export style:

```ts
// Importing
import Button from './Button'

// Exporting
export default Button
```

**Function component declaration**
It was decided to use either:

```tsx
const Button = ({ onPress, label }: ButtonProps) => { ... }
function Button({ onPress, label }: ButtonProps) { ... }
```

#### Some TS problems found
- `LayoutOrientation` not being exported in `Options.ts`
  - Closes #6343


guyca:
lint: Going with the default is fine by me. If there's any rule you're used to feel free to enable it. As long as it's enabled on save it's fine by me 😁 
Component import/export style: Lets go with import export which seems to be the standard
Functional component: I like 2 and 3

@yogevbd @ItsNoHax WDYT?

jinshin1013:
> As long as it's enabled on save it's fine by me 😁

I agree! Although it requires some user-side IDE configuration to enable auto-save which hopefully everyone has turned it on 🙂 .

> Lets go with import export which seems to be the standard

Nice! I'll also use `default export` style.

yogevbd:
@guyca @jinshin1013 looks good 👏

ItsNoHax:
Changes look great @jinshin1013 !

Before I give you my :+1:, do you mind sending me the exact VSCode settings you are using so I can test some things? (I wonder if these are also worth mentioning in the contributing docs)

jinshin1013:
@guyca I just can't get the Android E2E to succeed. I couldn't reproduce that problem on my local set up. Would you be able to have a look at it please 🙏

jinshin1013:
There are some options that I've found is not typed/documented but still affect the E2E tests:

```ts
// Equivalent of ['portrait', 'landscape'],
layout.orientation = 'default'

// This option affects 'set empty string badge on a current Tab should clear badge' E2E test.
bottomTab.badge.animateBadge = true
```

guyca:
@jinshin1013 
If `default` is equivalent to `[portrait, landscape]`, why can't we use the latter?

jinshin1013:
@guyca Is `default` not being used anymore? Wasn't sure if leaving `default` there was intentional in the playground app was intentional or not. If it is deprecated, I will change that to `[portrait, landscape]`.

jinshin1013:
@guyca It seemed like `layout.orientation = 'default'` is deprecated and functions the same as setting `['portrait', 'lanscape']`so I've removed testing it as it's a duplicate.

guyca:
w00t merged! 🔥 
Thanks @jinshin1013 and everyone else for helping landing this in master 🙇

PR #6346, opened 2020-06-28
Updated the instructions for tutorial

kasperoo:
Currently the tutorial specifies `Navigation.setDefaultOptions({`to be declared before `registerAppLaunchedListener` snippet. This is incorrect, and should be passed inside that function, but before the `Navigation.setRoot({})`. Otherwise the `Bridge not yet loaded` error will be thrown.

https://stackoverflow.com/questions/56740889/bridge-not-yet-loaded-thrown-when-using-latest-wix-react-native-navigation/56748173?noredirect=1#comment100061919_56748173

### DIFF
--- a/website/docs/docs-basic-navigation.mdx
+++ b/website/docs/docs-basic-navigation.mdx
@@ -221,7 +221,7 @@ Navigation.setDefaultOptions({
 });
 ```
 
-We need to add this snippet before registering the `registerAppLaunchedListener`. This way we ensure our theme is applied when our root is set. Our code should now look like this:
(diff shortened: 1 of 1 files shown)


guyca:
@kasperoo Thanks for the PR!

@yogevbd Calling `setDefaultOptions` before `appLaunched` should be possible. If js code is being executed, that means the bridge is available, no?

yogevbd:
This is a bug, fixed in https://github.com/wix/react-native-navigation/pull/6374 🙏

PR #6350, opened 2020-07-01
Fix TypeScript errors in the Playground app

guyca:
The TS migration has surfaced a lot of errors and missing options. This PR aims to address the wrong and missing option declarations and all compiler issues in the playground app.


package.json
```diff
@@ -67,7 +67,10 @@
     "@babel/plugin-proposal-export-default-from": "7.10.1",
     "@babel/plugin-proposal-export-namespace-from": "7.10.1",
     "@babel/types": "7.6.x",
+    "@react-native-community/blur": "^3.6.0",
```
ItsNoHax, line 70: I don't see this being used?

package.json
```diff
@@ -67,7 +67,10 @@
...
+    "@react-native-community/datetimepicker": "^2.5.0",
```
ItsNoHax, line 71: Not being used?

package.json
```diff
@@ -67,7 +67,10 @@
...
     "@react-native-community/eslint-config": "2.0.0",
+    "@react-native-community/netinfo": "^5.9.4",
```
ItsNoHax, line 73: Not being used

package.json
```diff
@@ -94,8 +97,10 @@
     "prettier": "2.0.5",
     "react": "16.11.0",
     "react-native": "0.62.2",
+    "react-native-gesture-handler": "^1.6.1",
```
ItsNoHax, line 100: not being used

package.json
```diff
@@ -94,8 +97,10 @@
...
     "react-native-keyboard-tracking-view": "5.x.x",
-    "react-native-ui-lib": "3.24.2",
+    "react-native-reanimated": "^1.9.0",
```
ItsNoHax, line 102: not being used

package.json
guyca: It's not - but it's a peer dependency of react-native-ui-lib which we had to upgrade. I hope we can get rid of this soon in their v6

PR #6353, opened 2020-07-01
Remove Playground project exclusion in tsconfig

ItsNoHax:
The goal behind this MR is to make sure that MRs fail CI checks if any typescript errors are detected in the Playground project.
- Added module declaration for 'react-native-ui-lib' since it has no TS typings
- Remove tsconfig.build.json since it's only goal was to exclude Playground project


ItsNoHax:
Closing this as it's not the proper solution since it now compiles the playground project out to lib/dist

PR #6357, opened 2020-07-02
Fix back button testID

yogevbd:
`backButton.testID` option was broken. This PR fix it.


website/api/options-backButton.mdx
```diff
@@ -9,42 +9,55 @@ Controls the back button styling.
...
 | ------- | -------- | -------- |
 | boolean | No       | Both     |
+
+### `testID`
+
+Used to interact with the back button in e2e tests.
+
+| Type    | Required | Platform |
+| ------- | -------- | -------- |
+| boolean | No       | Both     |
```
guyca, line 63: Type is string.
Also, why are there so many formatting changes in the file? Is it because of Prettier?

yogevbd: Yes

PR #6372, opened 2020-07-07
Fix overlay touch detection

### LINKED ISSUE #5976: [v6 Android] bottomTabs elevation blocks any overlayed components
### Issue Description

Seems that in v6, the bottomTabs take a higher elevation preference on android.  Essentially this breaks the ability to have an overlay (bespoke FAB in my case) positioned over the bottomTabs

image

I essentially can't press the big plus button (its an overlay).  

Perhaps this is an unintended regression caused by operator error, in which case I would love to know how to accomplish my goal.  Nonetheless, it only started with v6.

---
### Environment
* React Native Navigation version: v6.0.1
* Platform(s) (iOS, Android, or both?): Android
* Device info (Simulator/Device? OS version? Debug/Release?): Pixel 3 API 29

guyca:
When the Overlay was displayed with `statusBar.drawBehind: false`, the StatusBar height wasn't accounted for when calculating if touch events were inside the view's bounds.

To begin with, StatusBar height had to be accounted for because we used `MotionEvent.rawY` coordinates instead of `MotionEvent.y` coordinate which already take into account the view's position within the hierarchy, which compensates for StatusBar height. 

fixes #5976


bneigher:
🚀

bneigher:
has this been regressed?  I am now seeing this again in my app :/

PR #6392, opened 2020-07-13
Make constants getting synchronous

mrousavy:
## PR

I've been thinking about making the constant getting synchronous. This is especially helpful, because by using the constants we can measure our layout and create styles statically. Similar to `Dimensions.get('window')`, we can get react-native-navigation constants.

Note: before I thoroughly test this I wanted to ask if this even was a great idea (or has been considered before)?

:warning: Warning: As far as I know, this does not work in the Chrome debugger because of a limitation on react-native's side.

## Details

Use `isBlockingSynchronousMethod = true` for Java and `RCT_EXPORT_BLOCKING_SYNCHRONOUS_METHOD` for Objective-C to remove the Promise logic.

### DIFF
--- a/lib/android/app/src/main/java/com/reactnativenavigation/react/NavigationModule.java
+++ b/lib/android/app/src/main/java/com/reactnativenavigation/react/NavigationModule.java
@@ -77,16 +77,16 @@ public class NavigationModule extends ReactContextBaseJavaModule {
     public void getLaunchArgs(String commandId, Promise promise) {
         promise.resolve(LaunchArgsParser.parse(activity()));
     }
-    
-    @ReactMethod
-    public void getNavigationConstants(Promise promise) {
+
+    @ReactMethod(isBlockingSynchronousMethod = true)
+    public WritableMap getNavigationConstants() {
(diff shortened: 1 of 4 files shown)


guyca:
Hey @mrousavy !
This is a great suggestion, but I'm not sure this is how we'd like to move forward with it. The current implementation (promise based) lets developers control how and what is blocked.
The changes you suggest attempt to make the API synchronous by blocking the entire JS execution thread. This basically hides the enherint asynchronously of the api, which I don't see how we can avoid.

However, amazing job looking into this and coming up with an idea for improvement.
Honestly, I think the entire constants api can be removed now that SafeArea is seeing increased adoption.

mrousavy:
@guyca I see, I was thinking we should make this synchronous since the Dimensions API is synchronous as well and it would fit better. Also you could create Styles before the first render. I don't know for how long the threads are going to be blocked though...

Regarding the Safe Area View, you can't get the values using a Safe Area View, right? You can only use a SafeAreaView which will manage everything automatically, but let's say I want to have a custom view that's exactly the height of my Statusbar, I still can't do that?

PR #6393, opened 2020-07-13
Type component options

guyca:
This commit makes static options added to components strongly typed.

I also updated a few screens in the Playground app to reflect this, and also corrected rightButtons and leftButtons option which accepts both single button and an array of buttons.


guyca:
@ItsNoHax agreed. They accept both single button and array due to legacy reasons. Some modules in the Wix app use the old syntax 🤷

PR #6394, opened 2020-07-13
Add Option processors

yogevbd:
This PR adds the ability to intercept and modify options each time they applied from any command.
This feature comes in handy when there are guidelines need to be enforced when multiple developers are working together on different modules (or not).

- Add `Navigation.addOptionProcessor(optionPath: string, processor: (value: T, commandName: string) => T)` - When the specific option path is being processed, the processor function gets called and the processor has the option to manipulate this specific option value by returning a new one. It also receives the specific command name which triggered that option processing.
- Fix wallaby

`optionPath` examples: `topBar.title.color`, `bottomTabs.backgroundColor`.


lib/src/commands/OptionsProcessor.ts
```diff
@@ -12,34 +12,50 @@ import { ColorService } from '../adapters/ColorService';
...
   private processObject(
     objectToProcess: object,
     parentOptions: object,
-    onProcess: (key: string, parentOptions: object) => void
+    onProcess: (key: string, parentOptions: object) => void,
+    commandName: string,
+    path?: string
   ) {
     forEach(objectToProcess, (value, key) => {
+      path = this.resolveObjectPath(key, path);
```
guyca, line 57: We should avoid reassigning parameters and enable the [related eslint rule](https://eslint.org/docs/rules/no-param-reassign).

guyca: Seems like the entire class faults in this matter. Lets keep your current implementation and address this in a different PR.

lib/src/commands/LayoutTreeCrawler.test.ts
```diff
@@ -50,28 +50,9 @@ describe('LayoutTreeCrawler', () => {
...
       type: LayoutType.Component,
       data: { name: 'theComponentName', options: {} },
-      children: []
+      children: [],
     };
-    uut.crawl(node);
-    expect(node.data.options).toEqual({ popGesture: true });
-  });
-
-  it('Components: injects options from original component class static property', () => {
```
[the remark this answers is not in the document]
yogevbd: it already exists in line 40

lib/src/Navigation.ts
```diff
@@ -22,12 +22,13 @@ import { ColorService } from './adapters/ColorService';
...
 import { AppRegistryService } from './adapters/AppRegistryService';
 import { Deprecations } from './commands/Deprecations';
+import { ProcessorSubscription } from './interfaces/ProcessorSubscription';
 
 export class NavigationRoot {
   public readonly TouchablePreview = TouchablePreview;
 
   private readonly store: Store;
-  private readonly optionProcessorsRegistry: OptionProcessorsRegistry;
+  private readonly optionProcessorsRegistry: OptionProcessorsStore;
```
guyca, line 31: rename registry to store

PR #6399, opened 2020-07-14
Turbo Modules support

SudoPlz:
Breaking change:

The reason behind that is, there are no pros in handling js location in react-native-navigation
(this feels outside the scope of that library) and there are definitely cons of doing so.

For starters, it forces us to use [BridgeManager as the responder of RCTBridgeDelegate](https://github.com/wix/react-native-navigation/blob/master/lib/ios/RNNBridgeManager.m#L45), which
used to be fine, but..:

with TurboModules there's now a need of responding to other protocols, such as:
- the `RCTCxxBridgeDelegate` ([see here](https://github.com/facebook/react-native/blob/e549f6984e2125969fd11a5b8f4e2df866ac0495/React/CxxBridge/RCTCxxBridge.mm#L368)),
- and the `RCTTurboModuleManagerDelegate` all of which are declared
via the `RCTBridge` initialisation (which happened and still happens in `RNNBridgeManager`)

Changing the responder of `RCTBridgeDelegate` from `RNNBridgeManager` to the actual `AppDelegate` of
the app being built, allows the end user to integrate with TurboModules and customise their code as much as they
want, plus it provides a future proof  way of making sure RNN doesn't break react-native additions
especially since the turbo modules api is fairly new and likely to change.


lib/ios/ReactNativeNavigation.m
```diff
@@ -58,18 +55,11 @@ + (instancetype) sharedInstance {
...
 - (void)bootstrapWithDelegate:(id<RCTBridgeDelegate>)bridgeDelegate launchOptions:(NSDictionary *)launchOptions {
-    [self bootstrap:nil launchOptions:launchOptions bridgeManagerDelegate:bridgeDelegate];
-}
-
-- (void)bootstrap:(NSURL *)jsCodeLocation launchOptions:(NSDictionary *)launchOptions bridgeManagerDelegate:(id<RCTBridgeDelegate>)delegate {
 	UIWindow* mainWindow = [self initializeKeyWindow];
 	
-	self.bridgeManager = [[RNNBridgeManager alloc] initWithJsCodeLocation:jsCodeLocation launchOptions:launchOptions bridgeManagerDelegate:delegate mainWindow:mainWindow];
+	self.bridgeManager = [[RNNBridgeManager alloc] initWithlaunchOptions:launchOptions andBridgeDelegate:bridgeDelegate mainWindow:mainWindow];
+    [self.bridgeManager initializeBridge];
```
SudoPlz, line 62: Perfect!

SudoPlz:
@guyca do you foresee any changes needed on Android for turbo modules to work?

guyca:
@SudoPlz Seems like no changes are needed on Android.

PR #6407, opened 2020-07-20
Add layout processors

yogevbd:
This PR adds the ability to intercept and modify layout each time they are processed by each command (`setRoot`, `push`, `showModal`, `showOverlay`, `setStackRoot`).

Example usage:
```js
Navigation.addLayoutProcessor((layout, commandName) => {
  if (commandName === 'showModal' && layout.stack) {
    layout.stack.options = {
      topBar: {
        background: {
          color: 'gray',
        },
      },
    };
  }
  return layout;
});
```


lib/src/processors/LayoutProcessor.ts
```diff
@@ -0,0 +1,18 @@
+import { LayoutProcessorsStore } from './LayoutProcessorsStore';
+import { ExternalLayoutProcessor } from '../interfaces/Processors';
```
guyca, line 2: What's the difference between `ExternalLayoutProcessor` and `LayoutProcessor`? Feels like they should be the same, or at least share the same interface.

lib/src/interfaces/Processors.ts
```diff
@@ -0,0 +1,9 @@
+import { Layout } from './Layout';
+
+export interface ExternalLayoutProcessor {
```
guyca, line 3: I find the term `external` in a public facing API very confusing. If we're exporting a single type of OptionsProcessor and LayoutProcessor - the word `external` has no meaning or value to the user. Lets keep it short by removing external, WDYT?

lib/src/commands/Commands.ts
```diff
@@ -36,33 +20,39 @@ export class Commands {
...
+    const processedModals =
+      input.modals && this.layoutProcessor.process(input.modals, CommandName.SetRoot);
+    const modals = map(processedModals, (modal: any) => {
       return this.layoutTreeParser.parse(modal);
     });
 
-    const overlays = map(input.overlays, (overlay) => {
+    const processedOverlays =
+      input.overlays && this.layoutProcessor.process(input.overlays, CommandName.SetRoot);
+    const overlays = map(processedOverlays, (overlay: any) => {
```
guyca, line 40: Like the previous comment, Overlays should be processed inside the map, then we wouldn't need to check for null or undefined.

lib/src/interfaces/CommandName.ts
```diff
@@ -0,0 +1,17 @@
+export enum CommandName {
```
guyca, line 1: Don't we need to export this in `index.ts` ?

lib/src/interfaces/Processors.ts
```diff
@@ -0,0 +1,9 @@
+import { Layout } from './Layout';
+
+export interface LayoutProcessor {
```
guyca, line 3: Don't we need to export this in `index.ts` ?

lib/src/interfaces/Processors.ts
```diff
@@ -0,0 +1,9 @@
...
+  (layout: Layout<{}>, commandName: string): Layout<{}>;
+}
+
+export interface OptionsProcessor<T> {
```
guyca, line 7: Don't we need to export this in `index.ts` ?

PR #6409, opened 2020-07-20
Show deprecation warning while toggling bottomTabs visibility on iOS

yogevbd:
On iOS the system doesn't support dynamically toggling bottomTabs at all so we had to hack through it which resulted in many issues with this feature. We plan on entirely removing it and stick to the framework's guidelines which means that toggling the bottomTabs will not be possible at all and controlling it with static options will be the only option.

### DIFF
--- a/lib/src/commands/Deprecations.ts
+++ b/lib/src/commands/Deprecations.ts
@@ -1,10 +1,22 @@
+import isEqual from 'lodash/isEqual';
+import once from 'lodash/once';
+import { Platform } from 'react-native';
 
 export class Deprecations {
-  public onProcessOptions(_key: string, _parentOptions: Record<string, any>) {
-
+  public onProcessOptions(key: string, parentOptions: Record<string, any>, commandName: string) {
+    if (
+      isEqual(key, 'bottomTabs') &&
+      parentOptions[key].visible !== undefined &&
+      Platform.OS === 'ios' &&
+      commandName === 'mergeOptions'
+    ) {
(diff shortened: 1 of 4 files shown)


lib/src/commands/Deprecations.ts
```diff
@@ -1,10 +1,22 @@
+import isEqual from 'lodash/isEqual';
+import once from 'lodash/once';
+import { Platform } from 'react-native';
 
 export class Deprecations {
-  public onProcessOptions(_key: string, _parentOptions: Record<string, any>) {
-
+  public onProcessOptions(key: string, parentOptions: Record<string, any>, commandName: string) {
+    if (
```
guyca, line 7: This will warn each time visible is defined in options - even if it doesn't actually trigger a toggle which can be very confusing.
What about triggering the console.warn from the animate/toggle method in native? We can emit a "DeprecationEvent" from native to js, handle the event by registering an event listener in this class and console warning in the event handler.

lib/src/commands/Deprecations.ts
```diff
@@ -1,10 +1,25 @@
...
+      isEqual(key, 'bottomTabs') &&
```
guyca, line 8:
```suggestion
      key === 'bottomTabs' &&
```

lib/src/commands/Deprecations.ts
```diff
@@ -1,10 +1,25 @@
+import isEqual from 'lodash/isEqual';
```
guyca, line 1:
```suggestion
```

PR #6425, opened 2020-07-23
Fix shared element transition on pop

yogevbd:
The transition context needs to be different for pop and push, until now we didn't handle that.


guyca:
🔥

PR #6432, opened 2020-07-24
Keep an access to reactInstanceManager in ReactGateWay

magrinj:
This can be needed if we need to access some function of instanceManager from NavigationActivity, like `onConfigurationChanged` by exemple.

### DIFF
--- a/lib/android/app/src/main/java/com/reactnativenavigation/react/ReactGateway.java
+++ b/lib/android/app/src/main/java/com/reactnativenavigation/react/ReactGateway.java
@@ -4,6 +4,7 @@ import android.app.Activity;
 import android.content.Intent;
 
 import com.facebook.react.ReactNativeHost;
+import com.facebook.react.ReactInstanceManager;
 import com.reactnativenavigation.NavigationActivity;
 
 public class ReactGateway {
@@ -59,5 +60,9 @@ public class ReactGateway {
 
     public void onActivityResult(Activity activity, int requestCode, int resultCode, Intent data) {
         host.getReactInstanceManager().onActivityResult(activity, requestCode, resultCode, data);
+    }
+	
+    public ReactInstanceManager getReactInstanceManager() {
+        return host.getReactInstanceManager();
     }
 }


guyca:
@magrinj Why is this needed? The `instanceManager` can be accessed with `mReactNativeHost.getReactInstanceManager()`

magrinj:
@guyca Actually `mReactNativeHost` is only accessible from `MainApplication.java`.
In my case I need an access to `instanceManager` from the `MainActivity.java` file that inherit from `NavigationActivity`.

`NavigationActivity` allow has a private access to the `host` via `ReactGateWay` class, and the instance manager can't be retrieve from there, only function like `onActivityResult` are duplicated an accessible. A direct access to the instance should be better to avoid rewriting an access to all the functions.

In my case I need to change my `onConfigurationChanged` function from `MainActivity.java` due to a problem with `Appearance` that doesn't trigger the listener when the mode change.

So my function with the patch now looks like that and fix my problem: 

```java
  @Override
  public void onConfigurationChanged(Configuration newConfig) {
    super.onConfigurationChanged(newConfig);
    getReactGateway().getReactInstanceManager().onConfigurationChanged(this, newConfig);
  }
```

guyca:
Change `mReactNativeHost` `publish static` and access it from the Activity 👍

magrinj:
@guyca Actually when initialised `mReactNativeHost` need an access to the current class instance, I don't think putting this variable in `public static` will work. 

In a newly generated react-native app,`MainActivity` class inherit from `ReactActivity` and from this extended class you can access to the instance manager this way: `this.getReactInstanceManager().onConfigurationChanged(this, newConfig);`

Keeping a way to access this instance in a similar way of `ReactActivity` should be better.

guyca:
a `public static` field can be accessed from anywhere in the code.

magrinj:
> a `public static` field can be accessed from anywhere in the code.

@guyca I know that, but as I mentioned above, this variable is initialized with the current instance of the MainApplication class. Putting it in static will block that, as "this" will no longer be accessible. Maybe there is another solution.

But, as explained, having something that follow likely the same API, as it's done in a react-native bare project should be better.

magrinj:
@guyca Do you try the solution you've mentioned with a `public static`, if you just take a look on the way it's initialise you'll see what I'm talking about.

```java
  private final ReactNativeHost mReactNativeHost =
      new NavigationReactNativeHost(this) { // <----------------- Here to be initialised this variable need an access to the current class instance, by putting it in public static, this will no longer be accessible.
        @Override
        public boolean getUseDeveloperSupport() {
          return BuildConfig.DEBUG;
        }

        @Override
        protected List<ReactPackage> getPackages() {
          List<ReactPackage> packages = new PackageList(this).getPackages();
          return packages;
        }

        @Override
          return "index";
        }
      };
```

guyca:
```java
public class MainApplication extends NavigationApplication {
    public static ReactNativeHost mReactNativeHost;

    @Override
    public void onCreate() {
        mReactNativeHost = new NavigationReactNativeHost(this) {
        };
        super.onCreate();
    }
}
```

magrinj:
@guyca Thanks, I didn't think about initialising this way, it's been a long time since I've used Java. I'll do it this way and it should work.

My suggestion was more about keeping an access more like `ReactActivity` does, but thanks.

balsloev:
I was having the same issue and fixed is like this: 

In MainActivity.java

```java
@Override
public void onConfigurationChanged(Configuration newConfig) {
    super.onConfigurationChanged(newConfig);
 ((MainApplication)getApplicationContext()).getReactNativeHost().getReactInstanceManager().onConfigurationChanged(this, newConfig);
}
```

PR #6439, opened 2020-07-28
Fix modalPresentationStyle docs

migue1s:
The current default is said to be `modalPresentationStyle: 'fullScreen'` however this is not supported on Android. Making it `none` to reflect this (the actual android code seems to have it be `unspecified`, but it should be falling back to `none` from what I can understand)


migue1s:
I can't seem to add a label, it should be `type: documentation`

PR #4229, opened 2018-10-25
[V2][IOS] Resolve 'push'-command with next componentId for ios

Hesowcharov:
Behavior of `Navigation.push(someComponentId, layout)` between android and ios is different.
In android promise resolves with next componentId (id of component that was pushed), but on ios - componentId on which we pushed.
Android: [https://github.com/wix/react-native-navigation/blob/v2/lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/stack/StackController.java#L145](url)
Motivation of this PR - keep api behavior more consistent.

PR #6445, opened 2020-07-29
Resolve 'push'-command with next componentId for ios

hannojg:
This is based on this PR: https://github.com/wix/react-native-navigation/pull/4229
This PR is very old and has many merge conflicts, to make your lives easier I created this PR. It does exactly the same.
Happy to see this merged! 😊 

>Behavior of Navigation.push(someComponentId, layout) between android and ios is different.
In android, promise resolves with next componentId (id of component that was pushed), but on ios - componentId on which we pushed.
Android: https://github.com/wix/react-native-navigation/blob/v2/lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/stack/StackController.java#L145
The motivation of this PR - keep API behaviour more consistent.


guyca:
Hey @hannojg This will be addressed by #6422 . Thanks for bringing this to our attention 👍

PR #6478, opened 2020-08-19
link update

rf1804:
#6423

### DIFF
--- a/lib/src/Navigation.ts
+++ b/lib/src/Navigation.ts
@@ -148,7 +148,7 @@ export class NavigationRoot {
     reduxStore: any
   ): ComponentProvider {
     console.warn(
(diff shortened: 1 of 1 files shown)


jinshin1013:
@rf1804 Thanks for the PR submission, however the original link is already functioning correctly and I do not see a reason to merge this change.

rf1804:
@jinshin1013 I hope so you must have checked the link, because i've tried/tested on at-least 5 different devices and it didn't worked anywhere

Screenshot 2020-08-19 at 2 06 56 PM

jinshin1013:
@rf1804 mmm I cannot reproduce this on my devices, would you be able to clear cache or open the link in Incognito mode? It sometimes happen with invalid cache or dns.

PR #6482, opened 2020-08-19
Fix hitbox of UIBarButtonItem

### LINKED ISSUE #5000: Increasing Tappable Area of Top Bar Buttons
### Issue Description

When using custom icons for left/right buttons, the tappable area is significantly reduced as compared to native iOS icons.

Is there any way to improve the tappable area of custom buttons to extend to the edge of the screen and more vertically?

Example of the tappable area of native iOS icons created using :

```
rightButtons: [
  {
    id: "cart",
    systemItem: "add"
    color: "gray"
  }
]
```

rnn-system-icon

Versus the tappable area of custom icons created using:

```
rightButtons: [
  {
    id: "cart",
    icon: images.shopping_cart,
    color: "gray"
  }
]
```

rnn-png-icon

### Environment
* React Native Navigation version: 2.17.0
* React Native version: 0.59.4
* Device info (Simulator/Device? OS version? Debug/Release?): iPhone X 12.2 simulator in Xcode 10.1

danilobuerger:
Fixes #5000 

Before:
Screenshot 2020-08-19 at 18 54 21

After:
Screenshot 2020-08-19 at 18 53 11

As can be seen the hitbox increases maintaining the custom inset of 15 left.

This was changed to a button in https://github.com/wix/react-native-navigation/commit/545e5fef5fa570aaa20d95bbb40ed9aed72fc480 while adding support for insets. Since they are set on the image it works either way. Maybe @yogevbd has some more insight.


bneigher:
Yesss!

yogevbd:
The problem here is that it's changing the gap between two or more buttons.. @danilobuerger maybe instead we can set the button height to the `navigationBar` height?
Before the change:
Simulator Screen Shot - iPhone 11 - 2020-12-08 at 13 34 45

And after:

Simulator Screen Shot - iPhone 11 - 2020-12-08 at 13 37 14

danilobuerger:
@yogevbd The space is increased because of the default inset being 15 (as mentioned above). If we default this to 0, the spacing would not change.

guyca:
@yogevbd If we change the [default inset to 0 instead of 15](https://github.com/wix/react-native-navigation/commit/545e5fef5fa570aaa20d95bbb40ed9aed72fc480#diff-d3dce5aece5f91ac2a073b133903d083f7c930e703c795736ade377fa7241acbR198) we could accept this PR, right?

yogevbd:
@guyca Yeah, sounds good.

danilobuerger:
@yogevbd I removed the inset... now the buttons are approx. spaced as before:

Screenshot 2020-12-10 at 12 18 14

danilobuerger:
@yogevbd CI is passing now.

yogevbd:
@danilobuerger I'm testing it on the Wix app and reporting back. 💪

yogevbd:
Unfortunately it does affect the Wix app.
Before the change:
Screen Shot 2020-12-15 at 12 24 17
And after:
Screen Shot 2020-12-15 at 12 24 24

I talked with our UX and the reason they want to decrease the spacing between buttons is that some screens in the Wix app have more than one button and the title isn't presented correctly.
I know the current solution is a hack, @danilobuerger maybe we can find another solution that doesn't break the Wix app? 
One solution I can think about is to construct the button as we used to when iconInsets are different than the default. wdyt?

yogevbd (approved):

PR #6489, opened 2020-08-20
Fix large title styles

### LINKED ISSUE #6337: [iOS] LargeTitle is broken
### Issue Description

It's not possible to apply styles to largeTitle

You can reproduce it in the playground app, try to apply tobBar styles in to SearchScreen largeTitle.

out

styles applied: 
```js
static options() {
    return {
      topBar: {
        noBorder: true, // it doesn't work 
        title: {
          text: "Search2",
          color: "green",  // it doesn't work 
        },
        largeTitle: {
          visible: true,
          color: "red",  // it doesn't work 
        },
        searchBar: true,
        background: {
          translucent: true,
        },
        searchBarPlaceholder: "Start Typing",
        hideNavBarOnFocusSearchBar: false,
      },
    };
}
```

you can see noBorder works only when largeTitle is not visible
fontFamily  it doesn't work either
---
### Environment
* React Native Navigation version: 6.8.0
* React Native version: 0.62.2
* Device info (Simulator/Device? OS version? Debug/Release?): Iphone 11 Pro (13.5) Simulator

yogevbd:
On iOS 13, title attributes should be applied on `scrollEdgeAppearance` as it applied on `standardAppearance`.
@guyca We need to think about adding more options to control those two states.

Closes #6337

### DIFF
--- a/lib/ios/TopBarAppearancePresenter.m
+++ b/lib/ios/TopBarAppearancePresenter.m
@@ -72,8 +72,9 @@
     NSString* fontWeight = [largeTitleOptions.fontWeight getWithDefaultValue:nil];
     NSNumber* fontSize = [largeTitleOptions.fontSize getWithDefaultValue:nil];
     UIColor* fontColor = [largeTitleOptions.color getWithDefaultValue:nil];
-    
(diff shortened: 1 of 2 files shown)


guyca (approved):

chrise86:
Sorry @guyca but this still seems to be broken in RNN `6.12.2`. Setting the following either directly on a screen's options or in the default options does not work:
```
largeTitle: {
  color: 'red', // or whatever colour
}
```

Issue #2318, opened 2017-12-11
Dismiss keyboard when pushing screen not working


jarrx:
I want to dismiss the keyboard when pushing a new screen to navigator, but nothing happens and when I go back the keyboard is still open.
The container of the screen where the keyboard is, is a scrollview with `keyboardShouldPersistTaps='always'`.
So what I'm trying to do is
`Keyboard.dismiss()`
`this.props.navigator.push({
    screen: 'colorPicker',
    title: 'Color Picker'
})`

---
### Environment
* React Native Navigation version: ^1.1.300
* React Native version: 0.49.3
* Platform(s) (iOS, Android, or both?): iOS
* Device info (Simulator/Device? OS version? Debug/Release?): Simulator/Device. iOS version 11.0.3

kjbappsllc:
Maybe you have found the fix to this, however, for future people looking at the same issue as I have been, here is how I fixed it:

`let dismissKeyBoard = () => {
    return new Promise((resolve, reject) => {
      Keyboard.dismiss();
      setTimeout(() => {
        resolve();
      }, 5);
    });
  };`

mattmcdonald-uk:
I've also run into this issue, although when returning to the screen I'm unable to dismiss the keyboard at all. Definitely a bug here.

I'm also using a promise based workaround, though rather than an arbitrary time which may be too high or too low, no way of knowing, I'm waiting for the `keyboardWillChangeFrame` event which appears to be the first to occur when the keyboard dismisses. It doesn't seem to fire on Android though so I'm also using the `keyboardDidHide` event.

The promise will never resolve if the keyboard isn't open, so you'd need some way of tracking that to only use the method if needed.

````es6
new Promise((resolve, reject) => {
  const frameListener = Keyboard.addListener('keyboardWillChangeFrame', () => {
    frameListener.remove()
    didListener.remove()
    resolve()
  })

  const didListener = Keyboard.addListener('keyboardDidHide', () => {
    frameListener.remove()
    didListener.remove()
    resolve()
  })

  Keyboard.dismiss()
})
````

**Comment by jsina** [comment]
I've actually couldn't dismiss the keyboard after go back from pushed screen.

**Comment by raphaelhovsepyan** [comment]
Maybe this workaround can be helpful. Instead of using timeouts you can use `InteractionManager`.
```javascript
Keybaord.dismiss();
InteractionManager.runAfterInteractions(() => {
  // screen push logic
})
```

zabojad:
[...]

PR #6492, opened 2020-08-20
Added ability to blur focus on push for iOS

danilobuerger:
Setting
```
animations: {
  push: {
    blur: true
  }
}
```
blurs the old view controller (by calling `end editing`) when pushing onto it.

Fixes: #2318


danilobuerger:
@guyca @yogevbd Maybe not a good idea to do this on the native side. According to TextInputState:
```
// This class is responsible for coordinating the "focused" state for
// TextInputs. All calls relating to the keyboard should be funneled
// through here.
```

guyca:
😕

PR #6509, opened 2020-08-25
Fix topBar title with subtitle view

yogevbd:
Move topBar title with subtitle view initialisation to when the viewController view is already initialised. Fix's cases where the titleView initialised too soon and therefor not visible because `layoutSubviews` never gets called.


guyca (changes requested):
Looks good! Lets add e2e and we're good to go!

PR #6511, opened 2020-08-25
Fix crash if Shared Element ID is invalid

mrousavy:
Fixes a hard-crash on iOS if the specified Shared Element ID could not be found/invalid nativeID has been set.

Instead, the Shared Element Transition, which could not locate the specified View by it's `nativeID`, will just be ignored.


mrousavy:
I just noticed that when one `nativeID` (either source or target) is invalid/cannot be found, the Android app gets stuck. So the similar issue can be reproduced on Android.

I'm guessing this has something to do with [TransitionSetCreator::createSharedElementTransitions](https://github.com/wix/react-native-navigation/blob/master/lib/android/app/src/main/java/com/reactnativenavigation/views/element/TransitionSetCreator.kt#L24-L36), since that calls `reportTransitionCreated(...)` in **both callbacks**, when the `to` view has been found **and** when the `from` view has been found. Now I know those are `lazyinit`s, but the one of those callbacks never gets called and that's probably the problem here.

yogevbd (approved):

yogevbd:
@mrousavy When defining shared element transition we already wait for render. I think we should also throw an informative exception in such case in addition to the check you've added.

mrousavy:
@yogevbd I see. We could use the LogBox to show a warning

Issue #6519, opened 2020-08-28
[ENHANCEMENT] Animate borderRadius for (Shared) Element Transitions


mrousavy:
### Enhancement Description

It would be awesome to support animations on the `borderRadius` property for (Shared) Element Transitions.

In the [new Shared Element Transition demo](https://github.com/wix/react-native-navigation/pull/6516), the animation looks very jumpy because the borderRadius instantly goes from 20 -> 0:

<img src="https://user-images.githubusercontent.com/15199031/91563477-346d2100-e93f-11ea-90b2-51aa8cddb153.gif" width="60%">

> Also, the image has different widths (and therefore different _aspect ratios_), which also doesn't get animated properly. See https://github.com/wix/react-native-navigation/issues/6505 for more details

### Steps to Reproduce / Code Snippets / Screenshots

1. Create Shared Element Transition for Images
2. Assign source OR destination a `borderRadius`

mrousavy:
After reading the [apple developer doc about UIKit Animations](https://developer.apple.com/library/archive/documentation/WindowsViews/Conceptual/ViewPG_iPhoneOS/AnimatingViews/AnimatingViews.html), I've noticed that `cornerRadius` is not an animatable property. (for UIKit)

**...but**:
> In places where you want to perform more sophisticated animations, or animations not supported by the UIView class, you can use Core Animation and the view’s underlying layer to create the animation. Because view and layer objects are intricately linked together, changes to a view’s layer affect the view itself.

I've also found the following code snippet to successfully animate cornerRadius:

```objc
CABasicAnimation *animation = [CABasicAnimation animationWithKeyPath:@"cornerRadius"];
animation.timingFunction = [CAMediaTimingFunction functionWithName:kCAMediaTimingFunctionLinear];
animation.fromValue = [NSNumber numberWithFloat:10.0f];
animation.toValue = [NSNumber numberWithFloat:0.0f];
animation.duration = 1.0;
[viewToAnimate.layer addAnimation:animation forKey:@"cornerRadius"];
[animation.layer setCornerRadius:0.0];
```

mrousavy:
I got something working 👀

<table>
<tr>
<th>Playground</th>
<th>My app</th>
</tr>
<tr>
<td width="50%"><img src="https://user-images.githubusercontent.com/15199031/91988923-2c8ef180-ed30-11ea-99ae-304c70d50f46.gif"></td>
[...]

PR #6532, opened 2020-09-02
Animate borderRadius for Shared Element Transitions

mrousavy:
See https://github.com/wix/react-native-navigation/issues/6519

Creates a new `CornerRadiusTransition` which allows animating the `cornerRadius` property of a View's layer. The new Transition is added to the `DisplayLinkAnimation` array (in `createAnimations`) when the fromView's cornerRadius property is different than the toView's cornerRadius property.

## TODO

1. **Android support.** Sorry, I'm more of an iOS dev.
2. **Pop support.** I don't know why, but the cornerRadius animation only works on push. When going back in the Cocktail example app, the cornerRadius will not be visibly animated, even though the `CornerRadiusTransition::animateWithProgress` function is called.
3. [optional] **Masked Corner support.** Support cornerRadius only being animated on corners that actually change. Currently all 4 corners are being animated, which is not desirable in some rare occasions. I don't know if the Yoga engine uses the `UIView::layer.maskedCorners` property, but if yes, we just need to mirror that to our SharedElement view. I could look into that.

### DIFF
--- a/lib/ios/AnimatedReactView.m
+++ b/lib/ios/AnimatedReactView.m
@@ -1,9 +1,11 @@
 #import "AnimatedReactView.h"
 #import <React/UIView+React.h>
+#import "UIView+Utils.h"
 
 @implementation AnimatedReactView {
     UIView* _originalParent;
     CGRect _originalFrame;
+    CGFloat _originalCornerRadius;
     UIView* _toElement;
     UIColor* _fromColor;
     NSInteger _zIndex;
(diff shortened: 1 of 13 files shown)


mrousavy:
I've just noticed an issue with this implementation. When the source and destination element sizes/dimensions are different, the `cornerRadius` property is not calculated for the destination element, but rather for the source element. This means, when the source size is small (e.g. `100 x 100`) and has a `borderRadius` of `0`, and the destination size is large (e.g. `300 x 300`) and has a `borderRadius` of `150`, the `layer.cornerRadius` property will be calculated for the `100 x 100` size, instead of the `300 x 300` size. In other words, it looks like this:

ezgif com-video-to-gif

If any iOS/Core Animation wizard has a solution for this, please let me know.

yogevbd (approved):

Issue #6431, opened 2020-07-24
[ENHANCEMENT ✨] Elastic/Bounce interpolation animation for Shared Element Transitions


mrousavy:
### Progress

- [x] iOS: Add simple overshoot interpolation mode
- [ ] iOS: Add damped harmonic spring interpolation mode (with [this config](https://github.com/wix/react-native-navigation/issues/6431#issuecomment-679028328))
- [ ] iOS: Support custom properties for interpolation modes (e.g. `{ tension }` for `'overshoot'`, `{ mass, stiffness, damping }` for `'spring'`)
- [x] Android: Add simple overshoot interpolation mode
- [ ] Android: Add damped harmonic spring interpolation mode (with [this config](https://github.com/wix/react-native-navigation/issues/6431#issuecomment-679028328))
- [ ] Android: Support custom properties for interpolation modes (e.g. `{ tension }` for `'overshoot'`, `{ mass, stiffness, damping }` for `'spring'`)

### Introduction

Since Shared Element Transitions are becoming a really common pattern among modern mobile apps, it's awesome that this library supports those out of the box and natively.

### Problem

One problem with the current implementation is, that there are only accelerate/decelerate interpolation modes for (Shared) Element Transitions, which looks really unrealistic and "stiff". While a full animated API export, like custom interpolation on an `Animated.Value`, would be great, a solution to this seems almost impossible. It would however be awesome to support other pre-defined interpolation modes, such as `'spring'`, `'elastic'`, easing functions or `'bounce'`.

<table>
<tr>
<th>Apple (App of the Day): Bounce</th>
<th>wix/react-native-navigation: decelerate</th>
</tr>
<tr>
<td width="50%"><img src="https://user-images.githubusercontent.com/15199031/87281540-9e4a7c00-c4f3-11ea-9268-a64b48eaaf8d.gif" alt="huh weird, why isnt this showing?"></td>
<td width="50%"><img src="https://user-images.githubusercontent.com/15199031/88392537-734a0d00-cdbc-11ea-9442-97f018240477.gif" alt="huh weird, why isnt this showing?"></td>
</tr>
<tr>
<td width="50%">As you can see, they used a smooth bounce effect for the push and pop animations to make the whole app feel more physical. They also used a blurred background for the drag down gesture, which I've opened a separate issue for at <a href="https://github.com/wix/react-native-navigation/issues/6391">#6391</a>.</td>
[...]

PR #6544, opened 2020-09-08
Add Spring interpolation

mrousavy:
This PR adds a custom interpolation mode: `'spring'` for element transitions, shared element transitions and animations in general, ref https://github.com/wix/react-native-navigation/issues/6431

### iOS

Implemented a custom damped harmonic spring interpolation for iOS using [this grapher file](https://github.com/jenox/UIKit-Playground/tree/master/01-Demystifying-UIKit-Spring-Animations/Resources/DampedHarmonicSpring.gcx) for reference, which uses a custom `springiness` parameter (ζ) and `mass` parameter (m).

### Android

I used [OvershootInterpolator](https://developer.android.com/reference/android/view/animation/OvershootInterpolator) from the Android animation API, which only has a `tension` property to configure `springiness`.

### Demo

<table>
<tr>
<th>Android</th>
<th>iOS</th>
</tr>
<tr>
<td width="50%"
></td>
<td width="50%"></td>
</tr>
</table>

### TODO

1. Due to the differences in the two APIs, we have to find a matching configuration and parse/map that on each platform, so it looks similarily. I'm currently hard-coding the parameters (iOS: `springiness: 0.3`, `mass: 3`, Android: `tension: 1.0`), but they have to be exposed to the JS-API as-well. I'll leave that part up to you at wix, since that changes the public API. (See #6431 for my suggestions regarding the public TS API)
2. On iOS, I'd suggest instead of using single static functions for interpolation, we use an object-oriented approach to create interpolation instances. Then, just use the `interpolationInstance.interpolate(p)` function on that object to actually interpolate. This makes the API more extensible, and also allows for us to "cache" the parameters I'm creating on every call to `RNNSpring()` ([those lines right here](https://github.com/mrousavy/react-native-navigation/blob/3e0f1ea0cea4fc04072ee7cc8ac6ddc793df7a1c/lib/ios/RNNInterpolator.m#L42-L53))


mrousavy:
Also, why does [this property](https://github.com/wix/react-native-navigation/blob/master/lib/src/interfaces/Options.ts#L900) not have other interpolations such as `'linear'`, `'accelerateDecelerate'`, etc.?

yogevbd (approved):

guyca:
> Also, why does [this property](https://github.com/wix/react-native-navigation/blob/master/lib/src/interfaces/Options.ts#L900) not have other interpolations such as `'linear'`, `'accelerateDecelerate'`, etc.?

The TS interface is missing a few interpolation options, nice catch!

lib/ios/RNNInterpolator.m
```diff
@@ -28,13 +28,39 @@ static CGFloat RNNApplyInterpolation(CGFloat p, RNNInterpolationOptions interpol
...
+            return RNNSpring(p, 0.3, 3);
     }
 }
 
 static CGFloat RNNInterpolate(CGFloat from, CGFloat to, CGFloat p, RNNInterpolationOptions interpolation) {
     return from + RNNApplyInterpolation(p, interpolation) * (to - from);
 }
 
+static CGFloat RNNSpring(CGFloat p, CGFloat springiness, CGFloat mass) {
+	// TODO: Cache those allocations
```
guyca, line 42: Why do we need to cache these variables?

guyca (changes requested):

lib/ios/RNNInterpolator.m
mrousavy: Those functions are being executed many times in a short amount of time, that's why my gut was telling me to remove everything that doesn't _have_ to be in there to somewhere else, especially "statics" like those variables to improve performance. They are being created with the values that the user supplies (calculated from `mass` and `springiness`) and won't change from there on forward.

mrousavy: tbh it's just a bit of easy math, some `sqrt`s and `pow`s, nothing fancy like `cos`ines so we're probably fine and maybe the compiler will do some tricks anyways.

guyca (approved):

PR #6546, opened 2020-09-08
Resolve fonts using RCTFont

### LINKED ISSUE #6010: Providing fontWeight and fontFamily breaks custom font family in titles (And possibly other places)
### Issue Description

If you provide both a `fontWeight` and `fontFamily` in styling objects for `topBar.title` and `topBar.largeTitle` then on iOS the system font is used.

- Add a custom font into your application
- Add `fontWeight` and `fontFamily` values to your RNN styling
- Observe custom font face is ignored

The issue lies with this `if` statement in `RNNFontAttributesCreator`

```objc
CGFloat resolvedFontSize = [self resolveFontSize:currentFont fontSize:fontSize];
    
    titleTextAttributes[NSForegroundColorAttributeName] = color;
    
   // If font weight is provided, all other attributes are ignored!
    if (fontWeight) {
        titleTextAttributes[NSFontAttributeName] = [UIFont systemFontOfSize:resolvedFontSize weight:[RCTConvert UIFontWeight:fontWeight]];
    } else if (fontFamily){
        titleTextAttributes[NSFontAttributeName] = [UIFont fontWithName:fontFamily size:resolvedFontSize];
    } else if (fontSize && currentFont) {
        titleTextAttributes[NSFontAttributeName] = [UIFont fontWithDescriptor:currentFont.fontDescriptor size:resolvedFontSize];
    } else if (fontSize) {
        titleTextAttributes[NSFontAttributeName] = [UIFont systemFontOfSize:resolvedFontSize];
    }
```

---
### Environment
* React Native Navigation version: 6.0.1
* Platform(s) (iOS, Android, or both?): iOS (Unsure about Android)
* Device info (Simulator/Device? OS version? Debug/Release?): Simulator iOS 13.0, Debug

yogevbd:
Until now, passing `fontWeight` resulted in using the system fonts. Now it will resolve the font weight with the specified font family.

Closes #6010


lib/src/interfaces/Options.ts
```diff
@@ -5,16 +5,6 @@ import { ImageRequireSource, Insets } from 'react-native';
 // Only assign PlatformColor or DynamicColorIOS as a Color symbol!
 declare type Color = string | symbol;
 type FontFamily = string;
-type FontWeight =
```
guyca, line 8: What's the motivation for removing `FontWeight`? Seems like it's more explicit and provides better auto completion.

guyca (changes requested):

lib/src/interfaces/Options.ts
yogevbd: It's different from one font to another. https://developer.apple.com/fonts/system-fonts/

guyca: Seems like there's a general set of weights and each font can support a different subset. If a font doesn't support a given `fontWeight` then the utility functions in RCTFont use the closest weight.

Since specifying a weight which isn't supported by font doesn't error (instead an approximation is used, which is fine) I feel we should keep the type and have it include all weights supported in [RCTFont](https://github.com/facebook/react-native/blob/d0871d0a9a373e1d3ac35da46c85c0d0e793116d/React/Views/RCTFont.mm#L40). WDYT?
The auto completion is really beneficial in these cases 🙏

guyca (approved):

PR #6557, opened 2020-09-09
Fix searchBarHiddenWhenScrolling when presented on root screen

### LINKED ISSUE #3429: [ios][v2] searchBarHiddenWhenScrolling doesn't work when searchBar is present on root screen
### Issue Description

The searchBarHiddenWhenScrolling style property should make the search bar hide under the top bar when the user scrolls down in the content view. This works when the search bar is present on a child screen of the navigation stack, but when the search bar is present on the root screen of the navigation stack, it doesn't scroll away.

I looked into why this is happening, and it seems like it is due to the UIViewController lifecycle methods being called before the searchController property of the navigationItem is set. That is why this works properly for pushed screens, because in that case the UIViewController lifecycle methods are called after the view controller subviews have been set, thanks to the onReactViewReady call in -[RNNCommandsHandler push:layout:completion:rejection:]. That means one fix for this would be to ensure that the lifecycle methods are called for the root VC only after the react view is ready in setRoot, like we do in push. 

I tried making a change that fixes this issue with an approach similar to how showModal handles this, but I was not totally confident this fix is the right approach. The patch is here: https://gist.github.com/benhiller/5e6fdc8535f63a32680c1c46628d93b4. Let me know if this does seem like a reasonable approach to fix this issue and I can send out a pull request with this change.

1. Apply this patch to the v2 branch: https://gist.github.com/benhiller/dab2beacfb03eb35bd6aa5364e5767cf
2. Tap 'Push SearchBar' on the welcome screen
3. Scroll down, see that the search bar doesn't hide under the TopBar.

GIF demonstrating the issue: https://gfycat.com/FatherlyCarelessDolphin

---
### Environment
* React Native Navigation version: v2 at commit fbf8ea13874983b97c381bc3bac1cfcba9cd5b1c
* React Native version: 0.51.1
* Device info (Simulator/Device? OS version? Debug/Release?): iPhone 6 Simulator, iOS 11.4, Debug

mateioprea:
This one fixes #3429 

searchBarHiddenWhenScrolling was called before setting navigationItem.searchController. This PR fixes this issue.

### DIFF
--- a/lib/ios/RNNComponentPresenter.m
+++ b/lib/ios/RNNComponentPresenter.m
@@ -57,10 +57,12 @@
     
     if ([withDefault.topBar.searchBar getWithDefaultValue:NO]) {
         BOOL hideNavBarOnFocusSearchBar = YES;
+        
+        [viewController setSearchBarWithPlaceholder:[withDefault.topBar.searchBarPlaceholder getWithDefaultValue:@""] hideNavBarOnFocusSearchBar:hideNavBarOnFocusSearchBar];
+        
         if (withDefault.topBar.hideNavBarOnFocusSearchBar.hasValue) {
             hideNavBarOnFocusSearchBar = withDefault.topBar.hideNavBarOnFocusSearchBar.get;
         }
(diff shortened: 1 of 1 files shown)


mateioprea:
@yogevbd maybe you can help me with this reviewing this one? Thank you!

PR #6589, opened 2020-09-15
'spring' -> 'overshoot'

mrousavy:
This PR renames the `'spring'` interpolation mode to `'overshoot'`. 

On Android, an [`OvershootInterpolator`](https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/view/animation/OvershootInterpolator.java) was used, so it made no sense to call it `'spring'`. On iOS, I changed the Spring implementation to be the exact same of the android's [`OvershootInterpolator::getInterpolation`](https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/view/animation/OvershootInterpolator.java#73) so we have the same look and feel on both platforms. 

This means, using the `tension` property we can configure both animations. Now we only need to expose that to the JS API somehow.

If requested, I can add the spring implementation (which uses `mass`, `springiness`, `damping`, ... properties) again under the name `'spring'`, and try to find a way to implement that for Android.

Oh and also, I've added docs for the interpolation modes.

### DIFF
--- a/lib/ios/RNNInterpolator.m
+++ b/lib/ios/RNNInterpolator.m
@@ -28,9 +28,9 @@ static CGFloat RNNApplyInterpolation(CGFloat p, RNNInterpolationOptions interpol
             return RNNLinear(p);
         case RNNInterpolationDecelerate:
             return RNNDecelerate(p);
-        case RNNInterpolationSpring:
-			// TODO: Expose springiness and mass properties to JS-API and make uniform with Android implementation (only has tension property)
-            return RNNSpring(p, 0.3, 3);
+        case RNNInterpolationOvershoot:
+			// TODO: Expose tension property
+            return RNNOvershoot(p, 1);
     }
 }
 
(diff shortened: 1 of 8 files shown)


lib/ios/RNNInterpolator.m
```diff
@@ -28,37 +28,22 @@ static CGFloat RNNApplyInterpolation(CGFloat p, RNNInterpolationOptions interpol
...
-	if (lambda < w0) {
-		return fabs(1 - exp(-lambda * p) * (s0 * cos(wd * p) + ((v0 + s0 * lambda)/wd) * sin(wd * p)));
-	} else if (lambda > w0) {
-		return fabs(1 - exp(-lambda * p) * (((v0+s0 * (lambda + wd))/(2 * wd)) * exp(wd * p) + (s0 - (v0 + s0 * (lambda + wd)) / (2 * wd)) * exp(-wd * p)));
-	} else {
-		return fabs(1 - exp(-lambda * p) * (s0 + (v0 + lambda * s0) * p));
-	}
+static CGFloat RNNOvershoot(CGFloat p, CGFloat tension) {
+	// _o(t) = t * t * ((tension + 1) * t + tension)
+	// o(t) = _o(t - 1) + 1
```
yogevbd, line 43:
```suggestion
```

yogevbd (approved):

yogevbd:
💪💪

yogevbd (approved):

PR #6591, opened 2020-09-15
Configurable interpolation

mrousavy:
Add support to configure interpolation modes.

### Progress

- [x] iOS: Add `RNNInterpolation` interface with properties for each interpolation mode (parse JSON dictionary) as well as interpolation mode name
- [x] Android: Add `Interpolation` class with properties for each interpolation mode (parse JSON dictionary) as well as interpolation mode name
- [x] JS: Add types for interpolation modes with parameters
- [ ] JS: Deprecate old method

### TL;DR

this
```js
interpolation: 'overshoot'
```
becomes this
```js
interpolation: { mode: 'overshoot', tension: 2.3 }
```

### Explanation

Instead of:

```js
Navigation.push(this, {
  component: {
    name: Screens.CocktailDetailsScreen,
    options: {
      animations: {
        push: {
          sharedElementTransitions: [
            {
              fromId: `image${item.id}`,
              toId: `image${item.id}Dest`,
              duration: LONG_DURATION,
              interpolation: 'linear',
            },
          ],
        },
      },
    },
  },
});
```

You now use:

```js
Navigation.push(this, {
  component: {
    name: Screens.CocktailDetailsScreen,
    options: {
      animations: {
        push: {
          sharedElementTransitions: [
            {
              fromId: `image${item.id}`,
              toId: `image${item.id}Dest`,
              duration: LONG_DURATION,
              interpolation: { mode: 'linear' },
            },
          ],
        },
      },
    },
  },
});
```

The benefit here is, that the `'overshoot'` interpolation mode (and possible some future ones too) accept parameters. In the case of the overshoot mode, a `tension` property is accepted.

So now to customize the tension, I can use:

```js
Navigation.push(this, {
  component: {
    name: Screens.CocktailDetailsScreen,
    options: {
      animations: {
        push: {
          sharedElementTransitions: [
            {
              fromId: `image${item.id}`,
              toId: `image${item.id}Dest`,
              duration: LONG_DURATION,
              interpolation: { mode: 'overshoot', tension: 5 },
            },
          ],
        },
      },
    },
  },
});
```

The types are defined as:

```ts
type Interpolation =
  | { mode: 'linear' }
  | { mode: 'accelerateDecelerate' }
  | { mode: 'decelerate' }
  | { mode: 'accelerate' }
  | { mode: 'decelerateAccelerate' }
  | { mode: 'overshoot'; tension: number };
```

#### iOS

1. We need to change the current structure in `RNNInterpolator.h|.m` to use an object oriented approach instead of calling the function each frame (e.g. `RNNSpring(...)`). The reason behind this is that we can then store state properties related to the functions that either don't change with every execution (such as values derived from parameters = improves performance), or need to be stateful (e.g. velocity, last frame time, etc for spring animation)
2. We need to modify the usage of the enum parser (the one that parses `'overshoot'` -> `RNNOvershoot`) to also provide each interpolation class (see step 1) with the properties which will be read once in the `init` call (with default values)

@yogevbd can you help me out here?

#### Android

1. We need to modify the usage of the parser (the one that parses `'overshoot'` -> `new OvershootInterpolator`) to also supply the parsers with the properties which will be read once in the constructor (and default values)

@guyca can you help me out here?

Btw; Why does the screen transitioner only accept `accelerate` and `decelerate` interpolation modes? See: [RNNTransitionStateHolder.m](https://github.com/wix/react-native-navigation/blob/373f20b5800b354a774a0acb47f595ad0aa4f01e/lib/ios/RNNTransitionStateHolder.m#L51-L57) 

Btw 2; I just noticed that my Xcode is configured to use tabs instead of spaces. [those lines](https://github.com/wix/react-native-navigation/blob/3710530e51d815a17d8f723ad5723ad6623c82d0/lib/ios/SharedElementAnimator.m#L60-L66) are messed up, we need to do some chores...


guyca:
@mrousavy Looks great! We really like the new API.
1. Lets rename `mode` to `type`. I feel like the term `type` is less foreign and more frequently used in the RN eco-system
2. Changing interpolation is a breaking change - let's do it in a non-breaking manner. We should have the `Interpolation` type accept both the current string types and the new types you've added. In the OptionsProcessor we should be able to process both type of interpolations where the object notation takes precedence.

WDYT?

mrousavy:
@guyca Sounds good to me. So native side should only accept object notation from now on, right?

guyca:
> @guyca Sounds good to me. So native side should only accept object notation from now on, right?

Yup.

mrousavy:
So I've added some implementations of the interpolators (can be found in [lib/ios/Interpolators](https://github.com/mrousavy/react-native-navigation/tree/configurable-interpolation/lib/ios/Interpolators)). I've included some of the ones you find in the react-native Animated (or reanimated) library such as `back`, `ease` (which is bezier), `linear` and currently working on `spring`. All of those can be initialized with their parameters, e.g. `tension` for `overshoot` and `mass`, `damping`, `stiffness` for `spring`.

So currently the iOS API works in the following way:

1. It receives the animation data in JSON format through the bridge, which looks like this:
	```js
	{
		fromId: 'some_element.from',
		toId: 'some_element.to',
		interpolation: 'overshoot',
		duration: 500,
	}
	```
2. It parses the `interpolation` parameter using a `RCT_ENUM_CONVERTER` ([this one](https://github.com/mrousavy/react-native-navigation/blob/configurable-interpolation/lib/ios/RCTConvert%2BInterpolation.m)) which then returns the interpolation mode in enum type (e.g. `RNNInterpolationLinear`)
3. Using that parsed enum type, it then calls a **static** function for every frame with the new progress value (0-1) to interpolate. Example: [CornerRadiusTransition.m](https://github.com/mrousavy/react-native-navigation/blob/configurable-interpolation/lib/ios/CornerRadiusTransition.m#L14)

Now since I've changed the `Interpolation` types from string to object (to allow parameters, e.g.: `interpolation: 'overshoot'` -> `interpolation: { type: 'overshoot', tension: 1.3 }`) the `RCT_ENUM_CONVERTER` function (step 2) doesn't work correctly anymore. That's what I've been having some issues with right now, as I'm not sure how to target this. Ideally I wanted to create a converter that takes the `interpolation: ` JSON, creates a switch statement for every different interpolation mode to conditionally parse the parameters (e.g. `tension` for `'overshoot'`, `mass, damping, stiffness` for `'spring'`, none for `'linear'`, ..) and then returns an instance of the protocol `Interpolator` (defined in [Interpolator.h](https://github.com/mrousavy/react-native-navigation/blob/configurable-interpolation/lib/ios/Interpolators/Interpolator.h)). This `Interpolator` protocol can then be used instead of the **static** function calls (step 3), so instead of `RNNApplyInterpolation(p)` you run `interpolator.interpolate(p)`. 

I probably need your help @yogevbd since I'm not sure how to create that custom converter, I've tried to play around with `RCT_CUSTOM_CONVERTER` but I had no idea how to use it since there are no docs for that.

> Note: We also shouldn't forget that the interpolators currently only interpolate the `x` value, so we need to calculate from and to values extra.

mrousavy:
For **Android** we're already using an object oriented approach by creating instances of the different interpolators (see [Interpolation.java](https://github.com/mrousavy/react-native-navigation/blob/configurable-interpolation/lib/android/app/src/main/java/com/reactnativenavigation/options/params/Interpolation.java)), we just need to get the arguments from the JSON somehow into there. tbh I don't know anything on the Android side, so I'd happily accept some help from you @guyca

mrousavy:
Native part is done. It parses the new interpolation object and uses the interpolators on interface level (protocol in Objective C)

What's left to do:

* Rebase from master
* Deprecate old JS interpolation API - @guyca can you help me at that? what's the best way to do that? A layout parser?

mrousavy:
Closing because of https://github.com/wix/react-native-navigation/pull/6644

PR #6599, opened 2020-09-17
Damped harmonic spring interpolation

mrousavy:
### Progress

- [x] iOS (`RNNSpring`)
- [x] Android (custom `TimeInterpolator`: `SpringInterpolator.kt`)
- [x] JS API
- [ ] Customizable parameters (`mass`, `damping`, `stiffness`) (waiting for https://github.com/wix/react-native-navigation/pull/6591)

### Description

I've added the damped harmonic oscillator ("spring") function again, this time it has the same behaviour on Android and iOS.

Compared to the Overshoot interpolator, the Spring function is more customizeable and can bounce forward & backwards more than just once.

Currently the properties (`mass`, `damping`, `stiffness`) are static, (`3`, `500`, `1000` respectively), so we should probably wait for https://github.com/wix/react-native-navigation/pull/6591 until this lands in master. Could someone lend me a hand over there? @guyca @yogevbd ...

### Sources

[Here](https://github.com/software-mansion/react-native-reanimated/blob/2576ab80ef8cec6900d054dd080b44234f709a4a/src/reanimated2/animations.js#L135-L207) is the spring function from reanimated. I will take a look at that and check if it differs from my implementation!

Also, note that this function is not exactly the same as [the one on Wikipedia](https://en.wikipedia.org/wiki/Harmonic_oscillator#Damped_harmonic_oscillator). I've used [this medium article](https://medium.com/ios-os-x-development/demystifying-uikit-spring-animations-2bb868446773) as a source for my function (specifically [this grapher file (download)](https://github.com/jenox/UIKit-Playground/tree/master/01-Demystifying-UIKit-Spring-Animations/Resources/DampedHarmonicSpring.gcx)).

Screenshot 2020-09-17 at 16 48 28

> I will compare this to the function on Wikipedia and see if there are any differences.

#### React Native

React Native (RCTAnimated) also has a spring function, which can be found in [Libraries/NativeAnimation/Drivers/RCTSpringAnimation.m](https://github.com/facebook/react-native/blob/aee88b6843cea63d6aa0b5879ad6ef9da4701846/Libraries/NativeAnimation/Drivers/RCTSpringAnimation.m#L108-L206). I've tried using that, but couldn't figure out a way to make this work without copying their source code. I had the crazy idea to use their API instead of our custom one, so instead of using `interpolation: 'spring'`, we could use `interpolation: Easing.elastic(1)` or `interpolation: Easing.ease`.

### Stiffness and Damping

Also note that the stiffness is static, but can be calculated using:
```objc
CGFloat stiffness = mass * pow((2 * M_PI) / (T0), 2);
```
Same for damping:
```objc
CGFloat damping = 2 * (1 - springiness) * sqrt(stiffness * mass);
```

### DIFF
--- a/lib/android/app/src/main/java/com/reactnativenavigation/options/params/Interpolation.java
+++ b/lib/android/app/src/main/java/com/reactnativenavigation/options/params/Interpolation.java
@@ -8,11 +8,14 @@ import android.view.animation.AccelerateInterpolator;
 import android.view.animation.DecelerateInterpolator;
 import android.view.animation.LinearInterpolator;
 
+import com.reactnativenavigation.utils.SpringInterpolator;
+
 public enum Interpolation {
     ACCELERATE,
     DECELERATE,
     ACCELERATE_DECELERATE,
     OVERSHOOT,
+    SPRING,
     DEFAULT,
     NO_VALUE;
 
@@ -27,6 +30,9 @@ public enum Interpolation {
             case OVERSHOOT:
                 // TODO: Expose tension property to JS-API
                 return new OvershootInterpolator(1f);
+            case SPRING:
+                // TODO: Expose mass, damping and stiffness property to JS-API
+                return new SpringInterpolator(3, 500, 1000);
             case DEFAULT:
                 return new LinearInterpolator();
         }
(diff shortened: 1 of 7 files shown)


mrousavy:
Also; I will add functions such as `elastic`, `back` and `bounce` once we got #6591.

> Source code for those: [Easing.js from reanimated](https://github.com/software-mansion/react-native-reanimated/blob/2576ab80ef8cec6900d054dd080b44234f709a4a/src/reanimated2/Easing.js)

mrousavy:
Configureable interpolation (https://github.com/wix/react-native-navigation/commit/d9df9b5db4ee7b209ec3f1891386d0b947178de9) landed on master and will be available in 7.2.0.

This PR uses a different spring interpolation option (Damped Harmonic Spring Animation), so it can be used for future reference. But since the PR is no longer relevant I'm closing it!

PR #6600, opened 2020-09-17
Podspec: Depend on React-Core instead of React

mrousavy:
From now on native modules should only depend on `React-Core` instead of `React`.

This fixes Xcode 12 build errors, see: https://github.com/facebook/react-native/issues/29633

### DIFF
--- a/ReactNativeNavigation.podspec
+++ b/ReactNativeNavigation.podspec
@@ -32,7 +32,7 @@ Pod::Spec.new do |s|
     ss.dependency 'Folly/Fabric'
   end
 
-  s.dependency 'React'
+  s.dependency 'React-Core'
   s.dependency 'React-RCTImage'
   s.dependency 'React-RCTText'
   s.frameworks = 'UIKit'


danilobuerger:
If we change this to `React-Core` we should make sure that we are not missing other dependencies, as the old React pulled in the following dependencies:
```
  s.dependency "React-Core", version
  s.dependency "React-Core/DevSupport", version
  s.dependency "React-Core/RCTWebSocket", version
  s.dependency "React-RCTActionSheet", version
  s.dependency "React-RCTAnimation", version
  s.dependency "React-RCTBlob", version
  s.dependency "React-RCTImage", version
  s.dependency "React-RCTLinking", version
  s.dependency "React-RCTNetwork", version
  s.dependency "React-RCTSettings", version
  s.dependency "React-RCTText", version
  s.dependency "React-RCTVibration", version
```

yogevbd:
@mrousavy do we also need to have backward compatibility here?

yogevbd (approved):

PR #6603, opened 2020-09-22
Better resolving shared element corner radius value

### LINKED ISSUE #6596: FastImage support
### Issue Description

Currently the library (mostly the animations, like Shared Element Transitions) assumes that all images are presented in a `RCTImageView` (see [this line](https://github.com/wix/react-native-navigation/blob/3710530e51d815a17d8f723ad5723ad6623c82d0/lib/ios/Utils/UIView%2BUtils.m#L15-L16)). Unfortunately the default react Image component doesn't quite provide the features I need, therefore I switched to the FastImage component from **react-native-fast-image**.

While animations work somewhat as expected (even `resizeMode` transitions correctly, so it looks pretty smooth), I noticed that the borderRadius (`cornerRadius`) animation doesn't work.

#### RCTImageView

ezgif com-video-to-gif-2

#### FFFastImageView

ezgif com-video-to-gif

### Solution

I've tried to also return `ViewTypeImage` for Fast Images, like in [this diff](https://github.com/wix/react-native-navigation/compare/master...mrousavy:fast-image-support#diff-6466f809aba0f1bd8f89c0ac2f62c0cb), but that didn't magically affect cornerRadius :(

FastImage "natively supports borderRadius", but I couldn't figure out how to apply a cornerRadius to a FFFastImageView (which btw is only a `SDWebImage` on iOS)

### Environment
* React Native Navigation version: 7.0.0-snapshot.1216
* Platform(s) (iOS, Android, or both?): iOS, assuming it doesn't magically work on android either

yogevbd:
Until now we resolved the corner radius of shared element view from the view itself, but it is possible that corner radius are applied on this view from a wrapper parent so now we will also try to resolve it from any parent view that have the exact same bounds.

Closes #6596

### DIFF
--- a/lib/ios/AnimatedReactView.m
+++ b/lib/ios/AnimatedReactView.m
@@ -46,6 +46,7 @@
     _originalFrame = _reactView.frame;
     _originalTransform = element.layer.transform;
     _originalLayoutBounds = element.layer.bounds;
+    self.contentMode = element.contentMode;
     self.frame = self.location.fromFrame;
     _originalParent = _reactView.superview;
     _originalCornerRadius = element.layer.cornerRadius;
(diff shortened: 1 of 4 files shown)


mrousavy:
works like a charm! nice job 🎉

mrousavy (approved):

PR #6604, opened 2020-09-22
Fix unclickable button in TopBar

guyca:
Not much to say... button was unclickable while it should be  ¯\_(ツ)_/¯

### DIFF
--- a/playground/src/screens/CustomTopBar.tsx
+++ b/playground/src/screens/CustomTopBar.tsx
@@ -13,7 +13,7 @@ export default class CustomTopBar extends React.Component<Props> {
     return (
       <View collapsable={false} style={styles.container}>
         <TouchableOpacity
-          disabled={!this.props.clickable}
+          disabled={!!this.props.clickable}
           onPress={() => Alert.alert(this.props.title, 'Thanks for that :)')}
         >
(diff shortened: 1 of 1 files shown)


danilobuerger:
I don't think this fix is correct. The logic before was:
If something is `clickable` then its not `disabled`.
Now the logic is:
If something is `clickable` then its not not `disabled`. 
The former seems to be semantically correct.

Looking through any calling code, it appears that `clickable` is either missing (so false) or set to false. In my opinion, we should instead change that.

See 
* https://github.com/wix/react-native-navigation/blob/a66e8e93b456ec911aed7a5a49c14ea0a689c0b9/playground/src/screens/ButtonsScreen.tsx#L34-L41
* https://github.com/wix/react-native-navigation/blob/5f6853f6fb3695decf7a9aa2673cba6edbf11c96/playground/src/screens/OptionsScreen.tsx#L135-L141

guyca:
Oh no. Maybe working when I have a 38.8 fever wasn't such a good idea 😭

danilobuerger:
@guyca any news on this one? Should we close it?

guyca:
I'll get to it this week! 🙏

PR #6610, opened 2020-09-25
Support FastImage in Shared Element Transition

guyca:
This PR aims to add support for [FastImage](https://github.com/DylanVann/react-native-fast-image) in Shared Element Transition.

This PR turned out to be quite gigantic, I'll sum up the changes below:
* Wait for appearing screen to be fully rendered before preparing the shared element transition. Up until now shared views were detected immediately after they were created by the RN and the transition kicked off when all views were found. This caused issues as the views were not measured and laid out yet and therefore could not be animated.
* Bug fix - Animate Image bounds during the transition (Only the image matrix was animated until now 🤦‍♂️)
* Animate `FastImage` border radius
* Disable `React.Image` border radius animator in SET. Due to limitations of Fresco - animating corners of React.Image is overly complex to the point of impossible.
* Support animating images from a pressed/scaled down state


mrousavy (approved):
Tested changes in my codebase with smooth results. FastImage (Glide) borderRadius animations on Android work perfectly fine. Only thing left is borderRadius for Views. 😄

PR #6614, opened 2020-09-28
Include all ancestors scale when determining an element's initial scale

guyca:
This is needed when an entire screen is scaled down, typically with a gesture, and when the gesture reaches
a certain threshold the screen is popped.


mrousavy (approved):
Scale computation works perfectly, even in deeply nested components. Not sure if this is a special case or a side effect of this PR, but borderRadius animation doesn't work when I use a scale down to dismiss gesture...

Issue #6391, opened 2020-07-13
Shared Element Transition for Overlays and Modals, similar to Apple's app of the day


mrousavy:
### Issue Description

I think it would be great to have "overlay"-like views (Overlays and Modals) also be able to run shared element transitions in order to achieve some cool UI/UX. 

For example, looking at the App Store's app of the day animation, we have a Shared Element with a bounce/spring effect (which would also be cool as an `interpolation` option btw! 😉) which transitions into a subview over the parent's view where the parent is blurred out. The user can easily drag this view down which reveals the background view:

<img src="https://user-images.githubusercontent.com/15199031/87281540-9e4a7c00-c4f3-11ea-9268-a64b48eaaf8d.gif" alt="App store demo gif" height=600 />

I was able to achieve roughly the same (no "springy" effect unfortunately) in `react-navigation` using the `react-navigation-shared-element` library by defining a `StackCardStyleInterpolator`: 

<img src="https://user-images.githubusercontent.com/15199031/87282060-27fa4980-c4f4-11ea-9bf7-38de97a64cb1.gif" alt="react-navigation demo gif" height=600 />

Unfortunately with `wix/react-native-navigation` I can't set the details view as an overlay to the parent view and add a blur over it (which I can fade out on transition progress), it's always the white background:

<img src="https://user-images.githubusercontent.com/15199031/87282185-4f511680-c4f4-11ea-9e59-7f4ffb66dfe2.gif" alt="wix/react-native-navigation demo gif" height=600 />

> Also the scroll view behaves weirdly when I use my pan gesture handler to `transform: { scale }` the view down.. is this a separate issue?

So unless I'm missing something, that's currently not possible in wix navigation, that's why I'm opening this enhancement issue.

### Steps to Reproduce / Code Snippets / Screenshots

<table>
<tr>
  <th>App Store</th>
  <th>react-navigation</th>
  <th>wix/react-native-navigation</th>
</tr>
<tr>
  <td><img src="https://user-images.githubusercontent.com/15199031/87281540-9e4a7c00-c4f3-11ea-9268-a64b48eaaf8d.gif" alt="App store demo gif" height=400 /></td>
  <td><img src="https://user-images.githubusercontent.com/15199031/87282060-27fa4980-c4f4-11ea-9bf7-38de97a64cb1.gif" alt="react-navigation demo gif" height=400 /></td>
[...]

PR #6615, opened 2020-09-28
Add `blurParent` option

mrousavy:
### What

This adds support for a prop called `blurParent` which you can enable on a screen you're about to push. Once the new screen is pushed, the parent screen will be rendered beneath the child screen, and a blur-overlay (`UIBlurEffect`) is rendered above it.

> Requires `componentBackgroundColor` and `backgroundColor` to be transparent and a custom `push`/`pop` animation, like a simple fade-in animation, to be set (makes no sense in the default slide from right side animation)

### Why

I'm trying to implement Shared Element Animations that are as smooth as the Apple App of the Day Animation, see https://github.com/wix/react-native-navigation/issues/6391.

### Android

The blurParent effect only makes sense on iOS, since BlurViews aren't a common pattern on Android. Also, I have no idea how to do that in android. Would be great if we could see the same prop in Android, but with a dimmed overlay instead of blurred overlay.

### DIFF
--- a/lib/ios/RNNNavigationOptions.h
+++ b/lib/ios/RNNNavigationOptions.h
@@ -36,6 +36,7 @@ extern const NSInteger BLUR_TOPBAR_TAG;
 @property (nonatomic, strong) WindowOptions* window;
 
 @property (nonatomic, strong) Bool* popGesture;
+@property (nonatomic, strong) Bool* blurParent;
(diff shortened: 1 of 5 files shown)


mrousavy:
Going to close this as this doesn't really make sense anymore, now that we're working on adding SETs for Modals.

PR #6616, opened 2020-09-28
Remove redundant awaitPost before popping screen with Shared Element Transition

guyca:
This was originally added to workaround an issue with flickering images. We've since added a better workaround so this can be removed.


mrousavy (approved):
Works perfectly 👌

PR #6622, opened 2020-09-29
A different approach to animating View scale

guyca:
The current approach was to animate the views scaleX and scaleY properties. This worked pretty well as long as border radius was defined.
When border radius was defined, animating the scale properties distorted the views appearance as the rounded corners were scale unproportionally.

This PR changes things by animating the view's background drawable's dimensions and border radius.


mrousavy (dismissed):
~Ayyy 🎉 Works like a charm~
EDIT: View bounds are wrong on `pop()`

PR #6563, opened 2020-09-10
Fix animations last tick

yogevbd:
Closes #6521

PR #6625, opened 2020-09-30
Shared Element Transitions for Modals!

mrousavy:
This PR adds support for running Shared Element Transitions with `showModal` and `dismissModal` operations. By design, this also includes any animations (including element transitions, content animations, etc.) that are included in Stack operations (`push`, `pop` and `setStackRoot`)

Currently WIP, and iOS only


mrousavy:
Does it make more sense to give the `showModal` and `dismissModal` commands their own transitioning delegate back instead of re-using the `StackTransitioningDelegate` from `push`/`pop` commands? I can't seem to get the transitioning delegate to run it's animations, not sure why..

mrousavy:
At the current state the Shared Element Transitions work for Modals on iOS. 🥳  I've noticed two "regressions" though:

1. The Shared Elements don't start at a scaled down state (applied through transform), so instead a short flicker is noticeable.
2. The FastImage `borderRadius` transitions don't seem to work anymore.

Both of those are not always happening, and it seems like they're mutually exclusive bugs, in other words; If **1.** works, **2.** doesn't work and vice versa.

@guyca I'll take a look if I can implement Android myself, if not, I'll have to wait for you to come back from vacation/guild week. 🏖️

guyca:
@mrousavy This sounds amazing! Can't wait to get back from my staycation! 💪

mrousavy:
I can't work on the Android implementation right now because I got some refactoring to do for my app. I'm already using modals for SETs in my app, so Android just doesn't work atm, meaning it's pretty high priority for me so I'll take another look at it once I got my refactoring finished 😄 

About the two iOS issues; Looks like we already had something similar before (https://github.com/wix/react-native-navigation/pull/6603, https://github.com/wix/react-native-navigation/pull/6563, ...)

yogevbd:
@mrousavy I do think we should wait for a major release if it contains a breaking change, even though it's a minor change.. @guyca wdyt?

lib/src/commands/OptionsProcessor.test.ts
```diff
@@ -43,14 +43,14 @@ describe('navigation options', () => {
       blurOnUnmount: false,
       popGesture: false,
       modalPresentationStyle: OptionsModalPresentationStyle.fullScreen,
-      animations: { dismissModal: { alpha: { from: 0, to: 1 } } },
+      animations: { dismissModal: { content: { alpha: { from: 0, to: 1 } } } },
```
guyca, line 46: @mrousavy Modals don't have a content property. When declaring modal animations, the entire view of the modal is animated.

content is used to declare animations that operate on the views pushed into a stack.

We should keep the current API unchanged.

guyca (changes requested):

guyca:
Nice catch @yogevbd !
@mrousavy I commented on the api change 👍

lib/src/interfaces/Options.ts
```diff
@@ -1137,11 +1137,11 @@ export interface AnimationOptions {
   /**
    * Configure what animates when modal is shown
    */
-  showModal?: ViewAnimationOptions;
+  showModal?: StackAnimationOptions;
```
guyca, line 1140: @mrousavy This change should be reverted as well 👍

playground/src/screens/sharedElementCarDealer/CarDetailsScreen.tsx
```diff
@@ -26,7 +26,7 @@ const CarDetailsScreen: NavigationFunctionComponent<Props> = ({ car, componentId
   const onClosePressed = useCallback(() => {
     if (isClosing.current === true) return;
     isClosing.current = true;
-    Navigation.pop(componentId);
+    Navigation.dismissModal(componentId);
```
guyca, line 29: You've changed the cars example to use modals? Nice!

guyca (changes requested):

yogevbd (approved):

PR #6628, opened 2020-10-01
Fix initial shared element corner radius glitch

yogevbd:
For a split second, the shared element view isn't initialised with the correct corner radius.

Screen Shot 2020-10-01 at 15 57 32


mrousavy:
Also noticed that + the image not starting at a scaled down state when using drag to dismiss gesture. I thought I was going crazy after all those SET issues/prs lol

Will review & test tonight

mrousavy (approved):
Works perfectly

PR #6644, opened 2020-10-03
Configurable Interpolation!

mrousavy:
> Creating this as a new PR because I suck at merging.

This is https://github.com/wix/react-native-navigation/pull/6591 but with a master rebase.

### Progress

- [x] iOS: Add `RNNInterpolation` interface with properties for each interpolation mode (parse JSON dictionary) as well as interpolation mode name
- [x] Android: Add `Interpolation` class with properties for each interpolation mode (parse JSON dictionary) as well as interpolation mode name
- [x] JS: Add types for interpolation modes with parameters
- [x] JS: Deprecate old method and convert to new one from JS (Layout parser? @mateioprea could you help me here?)
- [x] Add docs

### TL;DR

this
```js
interpolation: 'spring'
```
becomes this
```js
interpolation: { mode: 'spring', mass: 3, stiffness: 200 }
```


mrousavy:
While the interpolation modes fully work on iOS and Android, it looks like there is something off on iOS.

* [Android Demo](https://i.imgur.com/2N9VbCq.mp4)
* [iOS Demo](https://i.imgur.com/PaaIl3g.mp4)

It definitely looks like the iOS implementation calculates the end-value wrong. Not yet sure if this is my fault (interpolation) or not. @yogevbd could you please take a look at that once you have some time?

mrousavy:
All done, will do some final testing to see if everything still works, then I'll convert this to a PR

mrousavy:
@guyca @yogevbd I've tested it a bit using the playground, and it looks like everything's working smoothly. (With exception of the comment above with the two demos, where it looks a bit jumpy. I still don't think that this is the fault of the spring interpolator, since I've capped the fraction in the `RNNInterpolate()` function at 1 so it cannot overshoot, but it still does jump a bit at the end. Maybe some `to`-coordinate calculation is wrong on iOS?)

Anyways, the deprecation warning is working, you can configure the interpolations (You can even configure the accelerate and decelerate interpolators), and I've added docs. If anyone of you two finds some time, that would be great since it is quite a big PR. Until then, have a nice guild week 👋

yogevbd (approved):

yogevbd (approved):

PR #6645, opened 2020-10-04
fix: ios topbar flicker on route pop

### LINKED ISSUE #6608: topBar flickers when popping screens on IOS12
### Issue Description

This does not reproduce on IOS13/14, but on IOS12 if the topBar has a non-default background color when using bottom tabs, popping screens causes the topBar to flash in the default color.

Repository with a repro is here: https://github.com/zzorba/RnnHeaderFlicker/tree/master

The steps seem to be

1) Use BottomTabs
2) Use `Navigation.setDefaultOptions` to change `topBar: { background: { color: 'red' } }`
3) Push a screen in one of the tabs, and then pop it.

---
### Environment
* React Native Navigation version: 7.0.0
* Platform(s): iOS
* Device info: Reproduces on both device and simulator, but seems to only happen on IOS12. I have not tested lower versions of iOS as the simulator is no longer available.

reime005:
On iOS devices lower than iOS 13, the top bar flickered (turned white) while transitioning back to the last screen. This should fix it. Tested on different simulators and devices with iOS versions 11, 12, 13 and 14.

Closes #6608

### DIFF
--- a/lib/ios/TopBarPresenter.m
+++ b/lib/ios/TopBarPresenter.m
@@ -83,7 +83,6 @@
         self.navigationController.navigationBar.translucent = YES;
     } else {
         self.navigationController.navigationBar.translucent = NO;
-        self.navigationController.navigationBar.barTintColor = nil;
     }
 }
 


danilobuerger:
@reime005 Thanks for the PR! Did you test setting a background color, so the following code path:
```
        self.navigationController.navigationBar.translucent = NO;
        self.navigationController.navigationBar.barTintColor = _backgroundColor;
```
and then setting (mergeOptions) no background color and not translucent. Will this remove the tint color?

reime005:
@danilobuerger that did not work for me

PR #6649, opened 2020-10-06
Create WindowOptions type

cHaLkdusT:
Setting default root window view's background color. Fixes linter warning when setting window's view background color

Sample usage:
```
    window: {
      backgroundColor: '#25C2A0',
    },
});
```

Screenshot 2020-10-06 at 19 51 04

### DIFF
--- a/lib/src/interfaces/Options.ts
+++ b/lib/src/interfaces/Options.ts
@@ -1139,9 +1139,12 @@ export interface NavigationBarOptions {
 }
 
 /**
- * Configure iOS' default background color of View
+ * Used for configuring and controlling the main window in iOS
  */
 export interface WindowOptions {
+  /**
+   * Configure the background color of the application's main window.
+   */
   backgroundColor?: Color;
 }
 
@@ -1248,7 +1251,7 @@ setRoot: {
    */
   rootBackgroundImage?: ImageResource;
   /**
-   * Background color for the screen
+   * Provides a way to configure the overall presentation of your application's main user interface
    * #### (iOS specific)
    */
   window?: WindowOptions;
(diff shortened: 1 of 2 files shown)


lib/src/interfaces/Options.ts
```diff
@@ -1240,6 +1247,11 @@ setRoot: {
    * #### (iOS specific)
    */
   rootBackgroundImage?: ImageResource;
+  /**
+   * Background color for the screen
+   * #### (iOS specific)
+   */
```
danilobuerger, line 1253: The comment should be more generalised as `WindowOptions` can potentially include more than just the background color.

lib/src/interfaces/Options.ts
```diff
@@ -1138,6 +1138,13 @@ export interface NavigationBarOptions {
   visible?: boolean;
 }
 
+/**
+ * Configure iOS' default background color of View
+ */
```
danilobuerger, line 1143: The comment should be more generalised as `WindowOptions` can potentially include more than just the background color.

lib/src/interfaces/Options.ts
```diff
@@ -1138,6 +1138,13 @@ export interface NavigationBarOptions {
...
+export interface WindowOptions {
+  backgroundColor?: Color;
```
danilobuerger, line 1145: Here a comment specifically regarding `backgroundColor` would be nice.

danilobuerger (changes requested):
Hi @cHaLkdusT 👋 Thanks for the PR, we really appreciate it! Please review the requested changes.

cHaLkdusT:
Hi @danilobuerger, thanks for your time reviewing this PR. I have now updated the documentation comments. Let me know if it is okay.

danilobuerger (approved):

PR #6655, opened 2020-10-08
Update options object documentation

cHaLkdusT:
Updated and aligned code documentation. Added undocumented properties (eg. WindowsOptions)


cHaLkdusT:
This is still an active PR.

PR #6586, opened 2020-09-15
Deprecate searchBar params in favour of a new standalone searchBar options

mateioprea:
We were discussing about moving the search bar params on iOS into a new options object. This will replace the old configuration in favour of the following:

```javascript
options: {
  topBar: {
    searchBar: {
      visible: boolean,
      hideOnScroll: boolean,
      hideTopBarOnFocus: boolean,
      obscuresBackgroundDuringPresentation: boolean,
      backgroundColor: color,
      tintColor: color,
      placeholder: string
    }
  }
}
```

closes #4093

PR #6657, opened 2020-10-09
Make processSearchBar more type safe, rename hiddenWhenScrolling to hideOnScroll

safaiyeh:
SearchBarOptions type uses `hideOnScroll` while the implementation used `hiddenWhenScrolling` updating the code to match this


mateioprea (approved):
Everything looks good! Thanks @safaiyeh! 👍

danilobuerger (changes requested):
As far as I can tell `hiddenWhenScrolling` was introduced with #6586 and released with 7.1.0. Renaming it would break semver if we don't bump the major version with the next release which seems unlikely for a rename. I would suggest splitting this up and just getting more type safety on `hiddenWhenScrolling` without actually renaming it. Then maybe a second PR for the rename that we could tag for v8 and merge at a later date.

danilobuerger (approved):
@safaiyeh pointed out on discord that `hideOnScroll` is actually the documented and typed name, also intended by description of #6586. Therefor we could treat this as a bug that the documented `hideOnScroll` is not being applied.

danilobuerger:
While going through it one more time I noticed that the whole `processSearchBar` is quite unreadable. So I decided on improving the readability and discovered a new bug. In order to fix that bug and increase readability, I am closing this in favour of #6676. Thanks @safaiyeh for all the effort!

PR #6674, opened 2020-10-15
Fix link script on iOS

yogevbd:
Sets the minimum iOS version to iOS 11.0 which is the minimum version required by the library.


danilobuerger (approved):
Looks good, although as previously discussed could have side effects if the podfile was previously modified. This could alter the platform of other targets or downgrade the version if previously set higher.

yogevbd:
@danilobuerger nice catch! fixed.

danilobuerger (approved):

danilobuerger:
Looks good, didn't test though.

yogevbd:
@danilobuerger tested it in multiple cases so I think we're good to go.

PR #6676, opened 2020-10-16
Process processed value instead of old one, bring type safety to processSearchBar

danilobuerger:
Inspired by #6657 and big thanks to @safaiyeh, I tried to make `processSearchBar` more readable and encountered a couple of bugs, fixed by this PR:

## Process processed value instead of old one

This fixes two cases:

1) A processor changes a values type to be an object or array. In that case we would previously not recurse down as the old value was still of another type.

2) A processor changes a field nested inside an object or array value. When recursing down, the options[key] would no longer match the value.

##  Bring type safety to processSearchBar …

Fixes an error where hideOnScroll was not being applied.

Also makes the whole function a lot more readable

## Reviewers

@mateioprea I hope you have some time to review and test this as you probably have the most experience and test cases with the search bar feature.
@guyca Could you please review the new general logic in `processObject` as outlined above?


guyca:
It seems like we've introduced a regression in 7.1.0 where the `hideOnScroll` option stopped working. Our test coverage for the SearchBar options processing is probably lacking. @danilobuerger Do you mind adding te missing tests so this doesn't break in the future?

danilobuerger:
@guyca tests added

guyca (approved):

PR #6702, opened 2020-10-20
Fix waitForRender + mergeOptions not working when view controller is in transit

### LINKED ISSUE #6085: TopBar buttons not rendered when waitForRender is set to true on iOS
### Issue Description

Seems like `mergeOptions` does not work well with `waitForRender: true`.

**Without `showModal: {  waitForRender: true }`**

image

**With `showModal: {  waitForRender: true }`**

image

1. Set default options:

```
    animations: {
      push: {
        waitForRender: true,
      },
    },
  });
```

2. Open any modal page with top bar buttons added by `mergeOptions`.

---
### Environment
* React Native Navigation version: 6.3.2
* Device info (Simulator/Device? OS version? Debug/Release?): Simulator, 13.2.2

danilobuerger:
When `waitForRender` is set to `true`, calling `Navigation.mergeOptions` early (for example in `componentDidMount`) will lead to the options not being applied as the backing view controller has not yet been placed in the window hierarchy. This PR aims to keep track of any pending view controllers and opens them up to being found via their component id. Thus enabling `Navigation.mergeOptions` to work on view controllers in transit.

The pending view controllers are now being kept in `RNNLayoutManager`. For that reason, the layout manager was converted to be instantiable.

A `NSHashTable` was chosen as the backing storage to support keeping the view controllers as a weak reference.

Fixes #6085

### DIFF
--- a/lib/ios/RNNBridgeManager.h
+++ b/lib/ios/RNNBridgeManager.h
@@ -19,4 +19,6 @@ typedef UIViewController * (^RNNExternalViewCreator)(NSDictionary *props,
 
 @property(readonly, nonatomic, strong) RCTBridge *bridge;
 
+- (UIViewController *)findComponentForId:(NSString *)componentId;
+
 @end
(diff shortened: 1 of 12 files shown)


danilobuerger:
@yogevbd this finally passes CI

PR #6705, opened 2020-10-21
[iOS] adds support for column-based SplitViews on iOS 14+

jpdriver:
[UISplitViewController](https://developer.apple.com/documentation/uikit/uisplitviewcontroller) was extensively rewritten as part of iOS 14.

It now supports either a two or three multi-column layout in place of the classic master-detail style.

rendered2x-1591053027

This PR extends `SplitView` to additionally support the updated implementation.

I have tried to ensure that there will be no breaking changes, and any existing apps should not behave any differently (even on iOS 14).

However -- these new options would now be available for those that want them (myself included).

---

# Examples

## "Classic" Behaviour
- still works the same as it does today
```
Navigation.setRoot({
  root: {
    splitView: {
      id: 'SPLIT_VIEW',
      master: {
        component: {
          id: 'MASTER_SCREEN',
          name: 'MasterScreen',
        },
      },
      detail: {
        component: {
          id: 'DETAIL_SCREEN',
          name: 'DetailScreen',
        },
      },
    },
  },
})
```
Screenshot 2020-10-21 at 10 59 44

## Three Columns, `twoBesideDisplay`
- uses `twoBesideSecondary` displayMode to show all columns by default
- primary column can be shown / hidden with toggle (provided automatically by OS)
```
Navigation.setRoot({
  root: {
    splitView: {
      id: 'SPLIT_VIEW',
      primary: {
        component: {
          id: 'PRIMARY_SCREEN',
          name: 'PrimaryScreen',
          options: {
            topBar: {
              title: {
                text: 'Primary',
              },
            },
          },
        },
      },
      supplementary: {
        component: {
          id: 'SUPPPLEMENTARY_SCREEN',
          name: 'SupplementaryScreen',
          options: {
            topBar: {
              title: {
                text: 'Supplementary',
              },
            },
          },
        },
      },
      secondary: {
        component: {
          id: 'SECONDARY_SCREEN',
          name: 'SecondaryScreen',
          options: {
            topBar: {
              title: {
                text: 'Secondary',
              },
            },
          },
        },
      },
      options: {
        splitView: {
          displayMode: 'twoBesideSecondary',
        },
      },
    },
  },
}),
```
Screenshot 2020-10-21 at 10 53 29

Oct-21-2020 10-54-14

## Three Columns, `secondaryOnly`
- primary and supplementary columns are hidden by default
```
      options: {
        splitView: {
          displayMode: 'secondaryOnly',
        },
      },
```

Oct-21-2020 10-57-55


lib/src/commands/LayoutTreeParser.ts
```diff
@@ -126,14 +126,29 @@ export class LayoutTreeParser {
...
+      return {
+        id: api.id || this.uniqueIdProvider.generate(LayoutType.SplitView),
+        type: LayoutType.SplitView,
+        data: { options: api.options },
+        children: master && detail ? [master, detail] : [],
+      };
+    } else {
+      // api is ModernLayoutSplitView
+      const primary = this.parse(api.primary);
+      const supplementary = api.supplementary ? this.parse(api.secondary) : undefined;
```
jpdriver, line 143:
```suggestion
      const supplementary = api.supplementary ? this.parse(api.supplementary) : undefined;
```

danilobuerger:
Hi @jpdriver 👋 Thanks for this awesome PR! I didn't do a code review yet, but a couple of points:

* It would be awesome if you could add a new screen to the playground app to showcase all the features of this.

* We don't want to have two ways to define a split view. So either we keep `master` & `detail` and add a `supplementary`, or we deprecate `master` & `detail` in js land (see f.x. https://github.com/wix/react-native-navigation/blob/master/lib/src/commands/OptionsProcessor.ts#L142 for the recent search bar deprecations and how to map them to the new names), add `primary`, `secondary` and `supplementary` and use them exclusively on the native side.

* For iOS < 14 we can then just ignore `supplementary` and for iOS 14 we can either add a `options.splitView.style` param next to the `options.splitView.displayMode` config options or just don't support the classic mode (is there any point in keeping it, does it look so different?).

Btw, we are also all available on Discord if there is any in depth discussion to be had.

jpdriver:
hey @danilobuerger 👋🏻

- there's already a Playground screen for the Master-Detail version (the Cocktail list). Would you like this new screen to replace that, or do it as a separate addition specifically showing 3x columns?

- i think for clarity we shouldn't mix the naming conventions (i.e. we shouldn't have `master`, `supplementary`, `detail`). i can re-work this so that behind the scenes everything is treated as `primary`, `supplementary`, and `secondary`.

- i'm don't know how many RNN apps are actually using the existing Master-Detail SplitView; but in the interest of flexibility it might be nice to keep some mechanism to continue using either one as required.. i love the idea of going with `options.splitView.style` so I'll update to use that instead

danilobuerger:
> * there's already a Playground screen for the Master-Detail version (the Cocktail list). Would you like this new screen to replace that, or do it as a separate addition specifically showing 3x columns?

Just some additions to showcase the new features (like supplementary, display mode and style)

> * i think for clarity we shouldn't mix the naming conventions (i.e. we shouldn't have `master`, `supplementary`, `detail`). i can re-work this so that behind the scenes everything is treated as `primary`, `supplementary`, and `secondary`.

Sure go with the new naming convention and lets deprecate the old one. See the example for the search bar that I mentioned above on how to convert the values in js and throw a deprecation warning.

> * i'm don't know how many RNN apps are actually using the existing Master-Detail SplitView; but in the interest of flexibility it might be nice to keep some mechanism to continue using either one as required.. i love the idea of going with `options.splitView.style` so I'll update to use that instead

Awesome! The following logic would be nice: 
* If no style is set in native, default to the new style (either two or three columns depending on which are set)
* In the options processor for the deprecation (js):
** if the user has set master / detail, also add the style prop (if not already set) and default to classic
** if the user has set primary / secondary do nothing

yogevbd (approved):

lprhodes:
I'd love to see this updated and merged, what ended up happening with it  @danilobuerger?

danilobuerger:
@lprhodes I have no idea, I am not a maintainer of this repo :)

jpdriver:
for what it's worth, i'm happy to dust this PR off and keep it ready to go -- but i completely understand that it likely can't come in until the next major..

coderkrishnabhai:
> @lprhodes I have no idea, I am not a maintainer of this repo :)

good

PR #6708, opened 2020-10-22
Fix ReactImageView sporadic flicker in pop during Shared Element Transition

guyca:
When animating shared elements in pop, the appearing screen needs a bit of time to settle in and for the React images to be displayed.
Seems like a good old post does the job in this case. This optimization is done only for ReactImages. For FastImage, we continue checking their measured bounds as before.


mrousavy:
Looks a bit weird atm:

### Image

https://i.imgur.com/2gf1gj8.mp4


### FastImage

https://i.imgur.com/kfIt7o2.mp4

guyca:
@mrousavy If you're referring to the border-radius issue observed when `<Image>` is used then that's unrelated to this PR and is a different issue I'll address in a separate PR.

What other issues do you observe?

mrousavy:
I don't think there's an issue with the borderRadius, it's just that the image doesn't start from a scaled down state. It does start at correct `x` and `y`, but it looks like the parent scale transforms are not applied to the `Image`, only to `FastImage` (which is just a View).

Do you want me to create a separate issue for that?

Other than that, everything works perfectly 🎉

guyca:
Oh gotcha! That's indeed a regression introduced in this PR. It's just hard to decipher gifs and figure out what the bug is. Thanks for the explanation.
No need for a separate issue, I'll fix it in this PR 👍

PR #6709, opened 2020-10-23
Fix default alpha animation being reversed

mrousavy:
If you comment out [these](https://github.com/wix/react-native-navigation/blob/master/playground/src/screens/sharedElementTransition/CocktailsListScreen.tsx#L51-L57) and [these](https://github.com/wix/react-native-navigation/blob/master/playground/src/screens/sharedElementTransition/CocktailsListScreen.tsx#L93-L99) lines in the playground demo, then test the Cocktails screen and select your favourite gin cocktail (@danilobuerger !!!!), you can notice that on popping the alpha animation is reversed. You'd expect it to go from visible (opacity of `1`) to invisible (opacity of `0`), but instead it jumps to `0` and then goes to `1` making it look really weird.

This PR changes this default behavior to reverse the default `FadeAnimation` on pop.


guyca (approved):
APPROVED X 1,000,000 👑

PR #6713, opened 2020-10-24
Add scroll edge appearance to iOS + docs

### LINKED ISSUE #6565: Support different topBar appearances on iOS
### Issue Description

Currently `topBar` configuration on iOS only supports `largeTitle`, but not different appearances. For example, I want to have a large title and a blurred background, but the background should be hidden when large title is visible.

For example:

```js
topBar: {
	largeTitle: {
		visible: true
	},
	drawBehind: true,
	background: {
		translucent: true
	}
}
```

This works, but the large title also has a blurred background when not scrolled down yet. But in many apps and native views on iOS, the behavior is like this:

- When large title is visible, the top bar does not have a background
- As soon as the view is scrolled and large title is converted to a normal title, the blurred background appears.

You can see this behavior in WhatsApp and Settings on iOS. Or for example GitHub app which looks like this:

IMG_7769

And when scrolled:
IMG_7770

Currently, RNN does this:

image

As you can see, background is visible even when not scrolled.

---
### Environment
* React Native Navigation version: 6.12.2

mateioprea:
Fixes #6565


julian-baumann:
Hey great work man. I've been waiting for this option. I recently tried to implement it by myself, but I know way too less objective c. Could you also add the "noBorder" option to it? That would be great.

mateioprea:
@julian-baumann it already works if you pass noBorder: true to topBar options. https://www.dropbox.com/s/j938kqz4xzral8z/Screen%20Recording%202020-10-24%20at%2019.38.14.mov?dl=0

julian-baumann:
Bit hard to tell in that video, but doesn't the border disappear also on the default header and not only on the large one, by setting this option?

mateioprea:
Yup, you're right. It's a combination of options. 
```javascript
topBar: {
        testID: STACK_SCREEN_HEADER,
        background: {
          color: "transparent",
        },
        scrollEdgeAppearance: {
          background: {
            translucent: true
          },
        },
        largeTitle: {
          visible: true,
        },
        noBorder: true,
        title: {
          text: 'Stack',
        },
      },
```


https://www.dropbox.com/s/n5ngixgf552pxzz/Screen%20Recording%202020-10-24%20at%2020.42.59.mov?dl=0

julian-baumann:
Nice, but is it also possible to hide the border only on the large title?

mateioprea:
Quick note here: 

- after this merge, if it wasn't for the "active: true/false" option, old apps would see a difference in their design. While this option might not be pleasing to have in the options object, since iOS allows us to customize scroll edge background color, i think that the "enabled: bool" will allow users to: 
	- define simple customizations using background topBar option (define a uniform background color for topBar + scroll edge together)
	- define complex customization with 2 different colors for scroll edge and top bar

@guyca / @danilobuerger what do you think?

yogevbd (approved):

Issue #6647, opened 2020-10-06
Shared Element Transitions for Modals


mrousavy:
### Intro

Currently Shared Element Transitions (SETs) only work for Screens by using `push` and `pop` commands. This is a bit weird, since there's a lot of issues with using custom screen animations, such as the top bar flickering and the bottom tabs sliding out. With Modals, we don't have those issues anymore while providing extra flexibility, such as showing the parent screen beneath, blurring the parent screen, etc.

### TODO

1. [x] Also create Shared Element Transitions for `showModal` and `dismissModal` commands for **iOS** (almost done in #6625 thanks to @yogevbd)
    - [ ] Fix new animations not being applied when passing them using `mergeOptions` in `dismissModal`
    - [x] Fix initial scale/transforms not being applied causing noticeable flickering (see [this demo](https://i.imgur.com/JHcOhy2.mp4))
    - [ ] Fix `borderRadius` animations not being applied (also see [this demo](https://i.imgur.com/JHcOhy2.mp4))
    - [x] Fix last tick in SET being a bit too far, maybe the initial `from` scale/transform is being applied but has been reset to `1` in the meantime (see [this demo](https://i.imgur.com/utPJeE6.mp4))
2. [x] Also create Shared Element Transitions for `showModal` and `dismissModal` commands for **Android** (done in #6716)
3. [x] Update JS API (types) to add support for `sharedElements` object in `showModal` and `dismissModal` (done in #6625)

### Demo

Here's a smooth demo to showcase what it could look like:

<div align="center">
     <img src="https://user-images.githubusercontent.com/15199031/87282060-27fa4980-c4f4-11ea-9bf7-38de97a64cb1.gif" width="50%">
</div>

mrousavy:
Here are some demos of the current implementation:

1. [imgur](https://i.imgur.com/bYa91Sa.mp4)
2. [imgur](https://i.imgur.com/9gJZXpQ.mp4)

What do you think @guyca @yogevbd ? ;)

In some cases, the drag down animation still doesn't work 100%, since it shortly flickers back to it's original position, instead of the scaled down one. Other than that, it's perfectly smooth on iOS.

That leaves us with TODOs: 

* [ ] Android
* [ ] that one flicker bug on iOS
* [ ] and `mergeOptions` not working for modal dismissal. (I'll look into that one)

guyca:
hmmm how often does the drag-flicker reproduce?

mrousavy:
[...]

PR #6714, opened 2020-10-25
Resolve parent cornerRadius only when clipped

yogevbd:
Addresses https://github.com/wix/react-native-navigation/issues/6647


mrousavy (approved):
awesome! 🎉

PR #6718, opened 2020-10-26
Fix scrollEdgeAppearance docs

sallar:
Seems like the docs for `scrollEdgeAppearance` don't match the actual code. Fixed it here.

cc @mateioprea


mateioprea (approved):
Looks good

PR #6732, opened 2020-11-02
Fix incorrect border radius end value

guyca:
While animating ReactImageView in shared element transition, we're changing the bounds of the destination image to max(sourceImage.bounds, destinationImage.bounds) and apply a clipping Rect.
This apparently messes up border-radius as Fresco isn't aware of the clipping Rect and draws the corners according to the actual image bounds.

This commit splits the clip bounds animator into two animators, one for Image and another for FastImage.
In the ReactImageView animator, we update both clip bounds and the drawable bounds

This was noticeable when the default resize mode (cover) was used.

**Note: this or doesn't add support for animating border-radius in ReactImageView. This will be added hopefully in a separate PR**

| Before                                                                                                                         | After                                                                                                                          |
|--------------------------------------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------------------------------------|
| Screenshot_1604324543 | Screenshot_1604324565 |


mrousavy (approved):
tested, borderRadius works 💪  (only scale isn't applied correctly, but that's another PR)

PR #6738, opened 2020-11-03
Convert CommonJS require to ES6 import syntax

cHaLkdusT:
Convert CommonJS require to ES6 import syntax to have consistent sample code snippets


guyca:
Thanks!

guyca (approved):

PR #6755, opened 2020-11-09
Fix right buttons lifecycle bug

mateioprea:
This PR fixes a small edge case. Previously currentButtons() was using 
```arrayByAddingObjectsFromArray:``` but this is not actually correct if the receiving array is empty. 

Quick explanation:

```arrayByAddingObjectsFromArray``` works by appending objects from another array to a receiver array. If the leftButtonItems was empty (the receiver), arrayByAddingObjectsFromArray will actually point to an empty object and will end up with **not adding** the latter array. That will cause ```UIBarButtonItem``` to stop notifying about itself appearing.

### CODE REVIEW

**Review by danilobuerger — DISMISSED** [review]
Looks good. A micro optimisation would be checking for nil instead of allocing a mutable array:
```objc
if (!self.viewController.navigationItem.leftBarButtonItems) {
  return self.viewController.navigationItem.rightBarButtonItems;
}
return [self.viewController.navigationItem.leftBarButtonItems arrayByAddingObjectsFromArray:self.viewController.navigationItem.rightBarButtonItems];
```
but I don't think thats really necessary.

danilobuerger:
@mateioprea @yogevbd Are you guys working on a e2e test for this?

mateioprea:
@danilobuerger sorry, i don't know why I didn't see the mention. An e2e test for this is required indeed, but I don't know what to test for.

guyca:
> @danilobuerger sorry, i don't know why I didn't see the mention. An e2e test for this is required indeed, but I don't know what to test for.

@mateioprea Either a unit test or an e2e checking that the expected lifecycle event is emitted would be great. 
We already have a few e2e tests covering various button related scenarios [here](https://github.com/wix/react-native-navigation/blob/master/e2e/Buttons.test.js).

danilobuerger (approved):

Issue #6773, opened 2020-11-16
[v7] Calling setDefaultOptions with custom title component fails


simonmitchell:
## 🐛 Bug Report

Calling `setDefaultOptions ` at any point with a component-based title on either iOS or Android fails due to `toString()` being called on undefined:

```
TypeError: undefined is not an object (evaluating 'componentName.toString')
ensureClassForName@http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false:100423:49
processComponent@http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false:105151:40
http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false:105082:34
http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false:103276:21
baseForOwn@http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false:103254:29
http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false:103297:24
forEach@http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false:104463:16
processObject@http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false:105071:26
http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false:105096:33
http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false:103276:21
baseForOwn@http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false:103254:29
http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false:103297:24
forEach@http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false:104463:16
processObject@http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false:105071:26
http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false:105096:33
http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false:103276:21
baseForOwn@http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false:103254:29
http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false:103297:24
forEach@http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false:104463:16
processObject@http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false:105071:26
http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false:105096:33
[...]

PR #6774, opened 2020-11-16
Fix/#6773 custom title component

simonmitchell:
Fixes issue #6773 where calling `setDefaultOptions` (And possibly other methods) with a custom component in the title of the navigation (And probably in other places too) would cause an error due to the assumption that any key in the options dictionary with value `component` would itself be a valid object containing a `name` key itself.


danilobuerger (changes requested):
Please get rid of the inner `if` and move the condition to the outer `if`.

danilobuerger:
Hi @simonmitchell. I don't really understand what this is trying to fix. When would `component` not be an object and contain a `name`? As you can see from the typings (left out unrelated fields):
```
export interface OptionsTopBarTitle {
  component?: {
    name: string;
  };
}
```
the `name` is required if the `component` is specified.

simonmitchell:
Hi @danilobuerger, the problem is that the function call is recursive, see the end of the `processObject` function:

```
const processedValue = objectToProcess[key];
if (!isEqual(key, 'passProps') && (isObject(processedValue) || isArray(processedValue))) {
   this.processObject(processedValue, parentOptions, onProcess, commandName, objectPath);
}
```

In our case, this leads to `processComponent` being called with a value that is a react component due to the options object we are passing in:

```
component: {
  name: "navigation.title",
  alignment: "center",
  component: require("components/common/navigation/title")
}
```

simonmitchell:
See issue, can be ignored, this was a coding error from a previous dev on our project that I assumed was an issue with this repo. My apologies!

PR #6781, opened 2020-11-20
Add the ability to override docs version from next docs version

yogevbd:


scripts/documentation.js
```diff
@@ -0,0 +1,33 @@
+const shellUtils = require('shell-utils');
+const exec = shellUtils.exec;
+const fs = require('fs');
+
+export function release(version, override = false) {
+  if (override) _removeDocsVersion(version);
```
guyca, line 6: If version 7.0.0 was already published and we want to update it - do we need to pass `override = true`? Can we detect if 7.0.0 was already published (by reading versions.json) and deleting it before generating it again?

scripts/documentation.js
```diff
@@ -0,0 +1,33 @@
...
+
+function _readDocsVersionsJson() {
+  return JSON.parse(fs.readFileSync(_getDocsVersionsJsonPath()));
+}
+
+function _writeDocsVersionsJson(versionsJson) {
+  fs.writeFileSync(_getDocsVersionsJsonPath(), JSON.stringify(versionsJson, null, 2));
+}
+
+function _getDocsVersionsJsonPath() {
```
guyca, line 27: A private getter to a hardcoded value is basically a const 🙃

guyca (changes requested):

guyca (approved):

Issue #6005, opened 2020-03-05
Bottom tab text cut off/truncated after changing fontWeight to semibold or bold.


drewhamlett:
### Issue Description

After upgrading to 6.1.0 , bottom text is cut off after changing the fontWeight to 'semibold' or 'bold' on the bottomTab property.  'light' or 'regular' doesn't affect the truncation.

### Steps to Reproduce / Code Snippets / Screenshots

![Screen Shot 2020-03-05 at 3 33 49 PM](https://user-images.githubusercontent.com/1056587/76023645-86d10600-5ef7-11ea-9f70-a381b25843be.png)

```js
 bottomTab: {
      fontWeight: 'semibold',
}
```
---
### Environment
* React Native Navigation version: 6.1.0
* React Native version: 0.61.5
* Platform(s) (iOS, Android, or both?): iOS
* Device info (Simulator/Device? OS version? Debug/Release?): Debug

guyca:
Hey @drewhamlett Can you please check if this happens on iOS 12 as well?

drewhamlett:
@guyca This is not happening on iOS 12.4

shanereid:
Same issue. Happens with regular fonts for me too. (ApercuPro I'm using)

RN 0.61.2 + RNN 6.1.0

PR #6784, opened 2020-11-21
Fix bottomTab truncated text on iPad

### LINKED ISSUE #6533: Tab bar text cut off and ellipsized on iPadOS 14
### Issue Description

When rendering a tab bar on an iPad running iPadOS Public Beta 6, text labels for tab bar items are truncated. 

IMG_0292

---
### Environment
* React Native Navigation version: Reproduced on playground app built from `master`
* Platform(s) (iOS, Android, or both?): iOS: iPadOS 14 Public Beta 6
* Device info (Simulator/Device? OS version? Debug/Release?): iPad Pro 11" 2018

yogevbd:
Closes #6533


guyca:
@yogevbd Seems like the two lines this PR deletes were added in b01629c. Are we certain this change won't cause regressions?

We gotta increase our screenshots coverage to cover BottomTabs as well. These iOS issues are madness.

yogevbd:
The issue described in #6005 doesn't reproduce on iOS 14.2.
In this case where it's an iOS new feature bug, I'm not sure how screenshot tests will help but let's think about it..

guyca (approved):

guyca:
We can use screenshot tests to ensure the tab item title isn't truncated in various font configuration.

ryanbourneuk:
Hey @guyca & @yogevbd!

I've just updated to v7.8.1, and I've noticed on iOS 13.3 (iPhone 11 Pro simulator) my tab bar text is now truncated! With the changes in this PR reverted locally, they're fixed again. The iOS 14.2 (iPhone 12 Pro simulator) works with and without the lines being re-added.

I'm happy to `patch-package` this locally (as my client's project does not use text for tabs on iPad), but it may be worth investigating & resolving so both iPad and iPhone can show text.

Happy to help in any way I can, let me know if you need more details.

Thanks! 😄

filipsuk:
I also had to use patch-package to revert this change. iPhone 8 simulator iOS 14.4 (and others) had issues with truncated text in the bottom menu.

PR #6785, opened 2020-11-21
Add componentWillAppear event

yogevbd:
Add `componentWillAppear` event which will notify the component that its view is about to be added to a view hierarchy.
This is equivalent to iOS [viewWillAppear](https://developer.apple.com/documentation/uikit/uiviewcontroller/1621510-viewwillappear) event.


guyca:
@yogevbd looks good! We should emit willAppear for external components as well.

a-eid:
is this PR still being considered, is there any plans to support android as well ?

yogevbd:
@a-eid yeah, we do plan on adding `componentWillAppear` for Android as well

lib/android/app/src/main/java/com/reactnativenavigation/views/component/ComponentLayout.java
```diff
@@ -50,13 +51,20 @@ public void start() {
         reactView.start();
     }
 
-	public void sendComponentStart() {
-		reactView.sendComponentStart(ComponentType.Component);
-	}
+    public void sendComponentWillStart() {
+        if (!willAppearSent)
```
yogevbd, line 55: `willAppear` should be sent each time the component is going to be appeared as `didAppear` does

nelsonprsousa:
@jpgarcia maybe this is something that could be added to rnn-hooks?

PR #6793, opened 2020-11-24
Fix SET frame transition

### LINKED ISSUE #6728: Shared Element Transitions: Animating from scaled down state looks jumpy
## 🐛 Bug Report

I have a view which the user can drag down to dismiss. When dragging down, the view also gets scaled down to indicate that it is being dismissed. Once it reaches a certain threshold, I dismiss the screen (`dismissModal` atm) which causes the Shared Element Transition to run.

Something similar to my use-case can be found in the playground repo.

Unfortunately, _sometimes_ the Shared Element (animating view, in my case a `<FastImage>`) does not start at it's transformed state (scaled down), but rather shortly goes to original position and only then start animating. When running at normal speed, this looks like a very weird flickering, and is definitely noticeable by users.

> I emphasized _sometimes_, because the bug is not reproducable 100% of the time, but rather happens randomly, I'd say about 1 out of 4 times.

Here is a video demo: https://i.imgur.com/4Fp0GAA.mp4

> If you didn't notice it, try playing it frame by frame (e.g. download & play in QuickTime)

### Have you read the [Contributing Guidelines on issues](https://wix.github.io/react-native-navigation/docs/meta-contributing)?

Yes

## To Reproduce

I've tried reproducing it in the playground, but couldn't get it to work. Maybe it's because my views are way more complex and therefore it leads to the animation start being delayed by a frame? I've tried debugging this to find out where the initial values are being set incorrectly, but I couldn't understand the whole codeflow, maybe @yogevbd could help here?

## Expected behavior

Shared Element Animation should transition smoothly starting from scaled down state (all transforms applied)

## Actual Behavior

Shared Element Animation _sometimes_ flickers back to original state then starts from scaled down state

## Your Environment

- React Native Navigation version: master (snapshot)
- React Native version: 0.63.3

## Reproducible Demo

^ above

## Related Issues

* SET for Modals: #6647
* #6729

yogevbd:
Closes #6728


mrousavy (changes requested):
Interesting! Didn't know we only needed the `TransformRectTransition`!

So I've tried it out, and it looks like it does change the behavior of the flickering, but not actually fix it.

Watch this demo: https://i.imgur.com/zbhXcjp.mp4

> (download it and watch it frame by frame)

Before, the Image was the original size instead of scaled down, now it is scaled down twice! You're definitely pulling the right strings @yogevbd, but it seems that somewhere the transform is applied twice 🤔

mrousavy:
<table>
<tr>
<th>Last frame before I call `pop`/`dismissModal`</th>
<th>First frame of Shared Element Transition</th>
<th>Second frame of Shared Element Transition</th>
</tr>
<tr>
<td></td>
<td></td>
<td></td>
</tr>
</table>

mrousavy:
So @yogevbd told me to try if it works if we also remove the `LayerBoundsTransition` (and the `TransformRectTransition`) animators from the `SharedElementAnimator.m`, and that seems to have done the trick. Really happy to see this weird bug fixed!
(Hope you don't mind I pushed those changes to your branch @yogevbd 😬)

So this PR can be merged - unless you want to push the translateX/translateY fix here as well @yogevbd ?

mrousavy (approved):

yogevbd:
@mrousavy let's merge and fix the translation in another PR. 💪

PR #6812, opened 2020-12-02
Add Navigation.dismissAllOverlays api

### LINKED ISSUE #4346: [Feature Request] DismissAllOverlays function
### Feature Request

In our application there are times when we need to force logout the user. To achieve this currently we need to dismissAllModals and then dismiss each overlay by name. Ideally we would be able to dissmissAllOverlays too.

mateioprea:
fixes #4346


lib/ios/RNNOverlayManager.m
```diff
@@ -29,10 +29,11 @@ - (void)dismissOverlay:(UIViewController *)viewController {
     [self detachOverlayWindow:overlayWindow];
 }
 
-- (void)dismissAllOverlays {
+- (void)dismissAllOverlays:(void (^__nullable)(void))completion {
```
danilobuerger, line 32: Why is a completion block needed? The method will block the thread anyway until the overlays are detached, so there is advantage on using a completion block.

lib/ios/RNNBridgeModule.m
```diff
@@ -218,6 +218,18 @@ - (instancetype)initWithCommandsHandler:(RNNCommandsHandler *)commandsHandler {
...
 }
 
+RCT_EXPORT_METHOD(dismissAllOverlays
+                  : (NSString *)commandId componentId
+                  : (RCTPromiseResolveBlock)resolve rejecter
+                  : (RCTPromiseRejectBlock)reject) {
+    RCTExecuteOnMainQueue(^{
+      [self->_commandsHandler dismissAllOverlays:commandId
+                                      completion:^{
+                                        resolve(@(1));
```
danilobuerger, line 228: Keeping with `dismissAllModals` this should resolve to `nil`

lib/ios/RNNOverlayManager.m
mateioprea: Because we need to wait for all of the overlays to be dismissed then we'll call completion() in the dismissAllOverlay method that will let JS land to know that everything is done.

danilobuerger: Yes, but the method `dismissAllOverlays` is synchronous anyway.

lib/ios/RNNBridgeModule.m
mateioprea: done 👍

lib/ios/RNNOverlayManager.m
mateioprea: Oh, you are right. I'll change it.

lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/overlay/OverlayManager.java
```diff
@@ -35,6 +36,14 @@ public void dismiss(ViewGroup overlaysContainer, String componentId, CommandList
         }
     }
 
+    public void dismissAll(ViewGroup overlaysContainer, CommandListener listener) {
+        for (ViewController overlay : overlayRegistry.values()) {
+            if (overlay != null) {
```
guyca, line 41: This null check is redundant as we keep a hard reference to the ViewController.

guyca (changes requested):

danilobuerger (approved):
iOS part looks good, @guyca please review the rest.

guyca:
Amazing work @mateioprea 🔥 
Android looks good, @danilobuerger thanks for reviewing the iOS part! @mateioprea once we update the docs we're good to go!

mateioprea:
@guyca just added the docs. we should be good to go

guyca (approved):

ItsNoHax:
:') thank you guys <3

PR #6829, opened 2020-12-09
Fix iOS hit test in overlays

### LINKED ISSUE #6051: [v6][ios] WKWebView in overlay ignore its touches
### Issue Description

By showing a WKWebView at fullscreen (for example the one from https://github.com/react-native-community/react-native-webview) on iOS in the overlay, the touches pass through the overlay and hit the navigator behind (for example the stack).
It's like the WKWebView is ignoring the touch events if inside the overlay.

The overlay component
```js
const AppHover: FunctionComponent<Props> = (props) => {
  return (
    <WebView
      source={{ uri: 'https://infinite.red' }}
      style={{ flex: 1 }}
    />
  )
}
```
The overlay options
```js
component: {
      name: "AppHover",
      options: {
        overlay: {
          interceptTouchOutside: false,
          handleKeyboardEvents: true
        },
        layout: {
          backgroundColor: "transparent",
          componentBackgroundColor: "transparent"
        }
      }
    }
```

### Environment
* React Native Navigation version: 6.3.0
* Platform(s): iOS
* Device: both emulator and device

alpha0010:
On iOS, `Switch` components do not function in overlays.

Code from @geroale in https://github.com/wix/react-native-navigation/issues/6051#issuecomment-600329866

Fixes #6051


danilobuerger:
I wonder if this relies too much on the implementation detail on how we currently structure the view hierarchy (@guyca, @yogevbd). Maybe there is a better solution on how to determine this?

In any case, there is no reason to use `count`, just access the first subview with `firstObject` instead of the subscript.

guyca:
@danilobuerger Well, since constructing the hierarchy is our responsibility, it seems like something we could rely on... especially in a hit test. How can we improve this conditional?

yogevbd (approved):

PR #6830, opened 2020-12-09
Increase iOS icon nav button hitbox

alpha0010:
Lower impact fix for #5000 (as compared to #6482, which also changes spacing in the nav bar).

### DIFF
--- a/lib/ios/RNNUIBarButtonItem.m
+++ b/lib/ios/RNNUIBarButtonItem.m
@@ -18,7 +18,7 @@
                   action:@selector(onButtonPressed)
         forControlEvents:UIControlEventTouchUpInside];
     [button setImage:iconImage forState:UIControlStateNormal];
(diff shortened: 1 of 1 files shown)


danilobuerger:
I don't like this as it does not increase the hitbox to the side. Also we shouldn't be using magic numbers where possible.

alpha0010:
From my testing, the vertical axis of the hitbox is the most impactful to perceived responsiveness; I did not try to edit horizontal, so as to avoid potential changes to existing icon layouts.

Looking into computing the appropriate number via api...

danilobuerger:
@alpha0010 You don't have to look into it at the moment. @yogevbd is currently testing #6482 with the Wix app. If thats merged, we would close this one.

danilobuerger:
Since we merged #6482, I am going to close this one.

PR #6834, opened 2020-12-12
disabling `rootViewController = nil` in Release Mode

kanzitelli:
In Release mode, when using `expo-updates` (as in [expo-rnn-starter](https://github.com/kanzitelli/expo-rnn-starter)) and want to update an app with the up-to-date bundle when it's active, the app crashes because it "loses" rootViewController in Release Mode. This PR doesn't anyhow affect an app in Debug or Release Mode, it just enables it to work with `expo-updates` and, I guess, as well as for `code-push`.

I have tested it already in production with [this app](https://apps.apple.com/ru/app/id1446775875) which I got released today and everything works good.

### DIFF
--- a/lib/ios/RNNBridgeManager.mm
+++ b/lib/ios/RNNBridgeManager.mm
@@ -134,7 +134,9 @@
     [_overlayManager dismissAllOverlays];
     [_modalManager dismissAllModalsSynchronosly];
     [_componentRegistry clear];
+#ifdef DEBUG
     UIApplication.sharedApplication.delegate.window.rootViewController = nil;
+#endif
 }
 
 @end


danilobuerger:
I don't understand why this would be the case. Do you have a stack trace for the crash?

Also it doesn't make much sense to not-nil the rootViewController when the bridge reloads.

Maybe there is something else going on with your starter? Glancing over it, it seems like you are initing the bridge too (which rnn does too). Maybe that is causing a conflict.

danilobuerger:
Glancing over the source code of expo-updates, it seems likely that you shouldn't be using the methods that you are using in your starter... it seems likely that setting the bridge on expo-updates from rnn and using the rnn bootstrap instead of expo-updates startAndShowLaunchScreen (probably use start instead) will be beneficial.

kanzitelli:
@danilobuerger thanks for the reply! Yes, you are right, I have been initializing two bridges in the current version (I will update soon). I didn't have any conflicts, which is weird, yes 😀  And I didn't use `expo-updates` widely but for the last project, it was necessary for me. So I encountered the issue above and it was caused when I want to use a new bundle during the app launch. And what I have come up with (#ifdef DEBUG), fixed my problem 😀 And I didn't dig into the problem, my bad. I am going to do it now and will keep you updated. Thank you

PR #6835, opened 2020-12-12
Bootstrap with user created bridge

danilobuerger:
To better align with the iOS react-native template, we should bootstrap with a user created bridge as this is created in the template. Bootstrapping rnn will then be migrated to:
```objc
[ReactNativeNavigation bootstrapWithBridge:bridge];
```
The old bootstrap method will be deprecated.

This also means that enabling fabric is passed back to user land on creation of the bridge (see RNTester example).


danilobuerger:
Closing in favor of #7132 .

RomualdPercereau:
The documentation haven't be updated since this modification. Should existing apps be updated ? @danilobuerger

PR #6843, opened 2020-12-16
Fix types passProps does not require componentId

alpha0010:
RNN generates `componentId`.  If using typesafe screen navigation, this results in error:
```ts
Navigation.showOverlay<MyProps>({
  component: {
    name: 'overlay.MyOverlay',
    passProps: { // Error, requires `componentId`.
      foo: 'bar',
    },
  },
});
```

### DIFF
--- a/lib/src/interfaces/Layout.ts
+++ b/lib/src/interfaces/Layout.ts
@@ -16,7 +16,7 @@ export interface LayoutComponent<P = {}> {
   /**
    * Properties to pass down to the component
    */
-  passProps?: P;
+  passProps?: Omit<P, 'componentId'>;
 }
 
 export interface LayoutStackChildren {


jinshin1013:
@alpha0010 Hi there, when you type your screen props, you do not have to manually add `componentId` to it and use built-in type from `react-native-navigation` instead:

```tsx
import { NavigationFunctionComponent } from 'react-native-navigation'

type ScreeenProps = {
  someProp: string
}

const Screen: NavigationFunctionComponent<ScreenProps> = ({ componentId, someProp }) => {
  return ...
}

// When you want to navigate to this
Navigation.showOverlay<ScreenProps>({
  component: {
    name: 'some.screen,
    passProps: {
      someProp: 'bar', // `componentId` is not required.
    },
  },
});
```

PR #6867, opened 2020-12-28
Fix ViewAnimationOptions not merging correctly

mrousavy:
Adds missing enable and waitForRender props for copying. Allows for showModal/dismissModal to run without animation.


danilobuerger (changes requested):
Shouldn't this just call super? `TransitionOptions` already merges those props.

danilobuerger (approved):

PR #6876, opened 2021-01-02
Fix resolving dismissAllModals when there are no presented modals

### LINKED ISSUE #6875: [v7.2+][iOS] dismissAllModals now does not resolve at all if there is no modal
## 🐛 Bug Report

If no modal is currently displayed, `dismissAllModals` does not resolve at all (neither in `then` nor in `catch`). This is due  to [this change](https://github.com/wix/react-native-navigation/pull/6671). 

It happens only on some versions  of iOS. So far, I've checked two versions :
- iOS 12.2 => broken
- iOS  14.3 => it works as expected...
I'm currently downloading other iOS simulator versions to check against them.

IMO, it should at least resolve or throw an error...

## To Reproduce

Call `dismissAllModals` when no modal is currently open. Check if it resolves of throw an error => none of those cases happen, nothing happen at all...

## Expected behavior

`dismissAllModals` should resolve if no modal is currently open.

## Actual Behavior

`dismissAllModals` does not resolve if no modal is currently open.

## Your Environment
- React Native Navigation version: Any version >= 7.2.0
- React Native version: Any
- Platform(s) (iOS, Android, or both?): at least iOS 12.2 and **not** iOS 14.3, 
- Device info (Simulator/Device? OS version? Debug/Release?): all

yogevbd:
Calling `Navigation.dismissAllModals` promise never resolved when no modals are presented. 

Closes #6875

### DIFF
--- a/lib/ios/RNNCommandsHandler.m
+++ b/lib/ios/RNNCommandsHandler.m
@@ -81,7 +81,10 @@ static NSString *const setDefaultOptions = @"setDefaultOptions";
         }
     }
 
-    [_modalManager dismissAllModalsAnimated:NO completion:nil];
+    [_modalManager dismissAllModalsAnimated:NO
+                                 completion:^{
+
+                                 }];
 
     UIViewController *vc = [_controllerFactory createLayout:layout[@"root"]];
     [_layoutManager addPendingViewController:vc];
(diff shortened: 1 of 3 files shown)


danilobuerger:
@yogevbd I don't think this is the right fix. The `CATransaction` handler will not run under certain conditions (fx e2e tests). Thats why I moved them in the first place. A better fix might be:
```objc
if (root.presentedViewController) {
  [root dismissViewControllerAnimated:animated completion:completion];
} else {
  completion();
}
```

yogevbd:
@danilobuerger I see. Will fix thanks!

danilobuerger (dismissed):

PR #6881, opened 2021-01-04
Fix buttons tint color on iOS 12

### LINKED ISSUE #6877: 7.7.0 regression with TopBar button color on IOS 12
## 🐛 Bug Report

When upgrading from version 7.6.0 to 7.7.0, it appears that setting colors for left/right buttons in the TopBar no longer works. I have tried setting them with both the topBar.leftButtonColor/rightButtonColor options as well as the individual color option on each button with the same effect. This doesn't appear to happen on IOS13.

Yes

## To Reproduce

1. Load a nav button icon to use (from react-native-svg or similar)
2. Construct a component in the following manner with a custom button color:
```  component: {
        name: 'My.Screen',
        options: {
          topBar: {
            title: {
              text: `Campaigns`,
            },
            rightButtonColor: '#888888',
            rightButtons: [{
              icon: iconsMap.add,
              id: 'add',
              color: '#888888',
              accessibilityLabel: 'New Campaign',
            }],
          },
        },
      },
    }; 
 ```
3. Observe that the buttons remain 'navigation' blue on iOS 12 simulator/real devices.

## Expected behavior

Expected the buttons change to the specified color.

## Actual Behavior

Buttons remain navigation blue (default)

## Your Environment

- React Native Navigation version: 7.7.0
- Platform: iOS
- Device info: Simulator/Real device running IOS 12

## Reproducible Demo

Can create one if necessary, but seems fairly easy to reproduce with any existing project.

### PR DESCRIPTION
**Pull request by yogevbd**
Fixes a regression created in [this commit](8131ea7650b997cbaefa2665fcad039005ae6955) which caused buttons with icon to never receive a tint color on iOS 12.

Closes #6877

### DIFF
--- a/lib/ios/RNNUIBarButtonItem.m
+++ b/lib/ios/RNNUIBarButtonItem.m
@@ -125,6 +125,7 @@
     self.accessibilityLabel = [buttonOptions.accessibilityLabel withDefault:nil];
     self.enabled = [buttonOptions.enabled withDefault:YES];
     self.accessibilityIdentifier = [buttonOptions.testID withDefault:nil];
+    self.tintColor = [buttonOptions.color withDefault:nil];
     [self applyTitleTextAttributes:buttonOptions];
     [self applyDisabledTitleTextAttributes:buttonOptions];
 }
--- a/playground/ios/NavigationTests/RNNUIBarButtonItemTest.m
+++ b/playground/ios/NavigationTests/RNNUIBarButtonItemTest.m
@@ -55,6 +55,20 @@
     XCTAssertTrue(CGSizeEqualToSize(button.frame.size, size));
 }
 
+- (void)testInitWithIcon_ShouldApplyTintColor {
+    UIColor *buttonColor = UIColor.redColor;
+
+    RNNButtonOptions *buttonOptions = RNNButtonOptions.new;
+    buttonOptions.icon = [Image withValue:UIImage.new];
+    buttonOptions.color = [Color withColor:buttonColor];
+    RNNUIBarButtonItem *barButtonItem =
+        [[RNNUIBarButtonItem alloc] initWithIcon:buttonOptions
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: lib/ios/RNNUIBarButtonItem.m [hunk 1]
```diff
@@ -125,6 +125,7 @@ - (void)applyOptions:(RNNButtonOptions *)buttonOptions {
     self.accessibilityLabel = [buttonOptions.accessibilityLabel withDefault:nil];
     self.enabled = [buttonOptions.enabled withDefault:YES];
     self.accessibilityIdentifier = [buttonOptions.testID withDefault:nil];
+    self.tintColor = [buttonOptions.color withDefault:nil];
```
zzorba, line 128: Seems like this might want to be `getWithDefaultValue:nil` instead of `withDefault:nil` -- ran into a compilation issue without it.

zzorba (changes requested):
Looks like there is a minor compilation issue.

That said, I can confirm that when changed to use `getWithDefaultValue:nil` the behavior is fixed on IOS12 (also confirmed that IOS13 continues to work).

lib/ios/RNNUIBarButtonItem.m
yogevbd: I renamed `getWithDefaultValue` to `withDefault`  [here](4a5b8a4ac44c4a3c6a0c81fa13e8aaf284c4c983). The next RNN version will include this fix.

zzorba: Oh totally missed that, thank you!

zzorba (approved):
Can confirm fix, thank you for the quick change!

zzorba (approved):

radiosilence:
I do not think disabling tinting works now?

tastyqbit:
@yogevbd Is there any update on why disabled tint on topbar buttons has stopped working?

Issue #6892, opened 2021-01-11
iOS 12.x: Navigation bar initial titleTextAttributes and largeTitleTextAttributes values return null


svbutko:
## 🐛 Bug Report

`titleTextAttributes` and `largeTitleTextAttributes` in the respected methods `setTitleAttributes` and `setLargeTitleAttributes` of `TopBarPresenter.m` initially return `NULL` which leads to incorrect `font-family`, `font-weight` and `font-size`

## To Reproduce

1. Create screen with `title` and `largeTitle` in `TopBar`
2. Run on iOS 12.X

## Expected behavior

<img width="432" alt="Screenshot 2021-01-11 at 4 15 51 PM" src="https://user-images.githubusercontent.com/14828004/104195514-63d28500-542b-11eb-9cb0-06cbaaa58e99.png">

## Actual Behavior

<img width="432" alt="Screenshot 2021-01-11 at 1 47 56 PM" src="https://user-images.githubusercontent.com/14828004/104195423-4ac9d400-542b-11eb-8449-d20bf2baa403.png">

Same goes for `title`

## Your Environment

- React Native Navigation version: 7.8.1
- React Native version: 0.63.4
- Platform(s) (iOS, Android, or both?): iOS
- Device info (Simulator/Device? OS version? Debug/Release?): Device and simulator, iOS 12.4, Debug and Release

PR #6893, opened 2021-01-11
Set initial text attributes for largeTitle and title if none are present

svbutko:
Adds initial font attributes to `largeTitleTextAttributes` and `titleTextAttributes` if none are present.
closes: [6892](https://github.com/wix/react-native-navigation/issues/6892)

### DIFF
--- a/lib/ios/TopBarPresenter.m
+++ b/lib/ios/TopBarPresenter.m
@@ -107,6 +107,10 @@
     NSNumber *fontSize = [titleOptions.fontSize withDefault:nil];
     UIColor *fontColor = [titleOptions.color withDefault:nil];
 
+    if (self.navigationController.navigationBar.titleTextAttributes == NULL) {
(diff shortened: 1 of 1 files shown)


svbutko:
In case of `titleTextAttributes` both `font-size` and `font-style` are correct if using `RNNFontAttributesCreator` without any changes but `font-weight` is incorrect (`normal` not expected `bold`).

In case of `largeTitleTextAttributes` it gets a bit difficult since `UIFontTextStyleLargeTitle` returns font without `bold` `font-weight` which again differs from platform. And then if not applied to `fontFamily` and `fontWeight` it gets incorrectly replaced by `RCTFont updateFont` to `SFUIText-Semibold` and `font-weight: normal` rather than `SFUIDisplay-Bold` and `font-weight: bold`

The solution isn't elegant but can be specific to any iOS version if these values will be NULL again.

yogevbd (approved):

PR #6902, opened 2021-01-17
Fix title alignment insets

yogevbd:
Fix's presenting title component with `alignment: 'fill'` wrong insets.


swabbass:
a Fix Here #6903

PR #6910, opened 2021-01-18
Fix icon color being applied to all instances of the same icon in release mode

Andarius:
As @guyca suggested, this fix an issue where the icon color was modified in release mode, because resources are bundled in release. RNN was changing the Drawable's color, but wasn't mutating the drawable.


guyca (approved):

Andarius:
It seems like the build failed because of this

image

PR #6918, opened 2021-01-20
Title bar placement, margins, RTL fix 

swabbass:
Still work in progress:
- [x] Solution, sanity check in the playground.
- [x] refactor old tests + add new ones for new units. (in-progress)
- [x] clean the code.

**Issue:**
We had several reports about the misbehaving toolbar layout, margin issues, RTL and the basic layout.
Since the **`leftButtons`** is coupled to the title itself this caused margin issues in some flows like when there is a custom component in the buttons, and in general, RTL wasn't working as expected.

**Fix:**
- Split ToolBar to 3 sections (**`LeftButtonsBar`**, **`TitleBar`**, **`RightButtonsBar`**)
- Left buttons remains as is in terms of the content, the placement relative to nearby views fixed.
- Title Bar converted from Toolbar (2 textviews/ component + not needed toolbar code) to ConstraintLayout (2 textviews/ component) (need checks for backwards compatibility)
- Right buttons have changed like left buttons, relative placement.
- Fix Rtl remaining RTL issues with aligning custom component.

**Note:** a couple of classes were converted to Kotlin for convenience.


lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/viewcontroller/LayoutDirectionApplier.kt
```diff
@@ -0,0 +1,15 @@
...
+
+class LayoutDirectionApplier {
+    fun apply(root: ViewController<*>, options: Options, instanceManager: ReactInstanceManager) {
+        if (options.layout.direction.hasValue() && instanceManager.currentReactContext != null) {
+            root.activity.window.decorView.layoutDirection = options.layout.direction.get()
+            I18nUtil.getInstance().allowRTL(instanceManager.currentReactContext, options.layout.direction.isRtl)
+            I18nUtil.getInstance().forceRTL(instanceManager.currentReactContext, options.layout.direction.isRtl)
+        }
+    }
+}
```
yogevbd, line 15: Couldn't find any reference to `LayoutDirectionApplier`, do we really need this?

lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/stack/StackPresenter.java
```diff
@@ -186,35 +191,40 @@ private void applyTopBarOptions(Options options, StackController stack, ViewCont
...
+            topBar.setTitleTypeface(typefaceLoader, topBarOptions.title.font);
+            topBar.setTitleAlignment(topBarOptions.title.alignment);
+
+            //check for text subtitle
+            String subTitleText = topBarOptions.subtitle.text.get("");
+            topBar.setSubtitle(subTitleText);
+            topBar.setSubtitleFontSize(topBarOptions.subtitle.fontSize.get(defaultSubtitleFontSize));
+            topBar.setSubtitleColor(topBarOptions.subtitle.color.get(DEFAULT_SUBTITLE_COLOR));
+            topBar.setSubtitleTypeface(typefaceLoader, topBarOptions.subtitle.font);
+            topBar.setSubtitleAlignment(topBarOptions.subtitle.alignment);
```
yogevbd, line 222: Feels like we can split it into two functions, wdyt?

lib/android/app/src/main/java/com/reactnativenavigation/options/TitleOptions.java
```diff
@@ -47,7 +47,14 @@ public static TitleOptions parse(Context context, TypefaceLoader typefaceManager
     public Number topMargin = new NullNumber();
 
     void mergeWith(final TitleOptions other) {
-        if (other.text.hasValue()) text = other.text;
+        if (other.text.hasValue()) {
+            text = other.text;
+            this.component = new ComponentOptions();
```
yogevbd, line 52: why do we need to create `ComponentOptions` instance here? it doesn't feel right

lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/viewcontroller/LayoutDirectionApplier.kt
swabbass: idk if we really need this, it is mainly used to change layout direction in order to make all children in the tree rerender with changed directions, you may look at the blame of the old file (the java one, converted while debugging to kotlin)

lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/stack/StackPresenter.java
swabbass: it can be split for sure, but I think adding functions just for naming (called in one place) is a bit of an overhead to achieve good readability, however, comments will explain a section of the code without changing the code, and in terms of maintainability its the same for functions, if you changed the responsibility of the function you might want to change the name too, hence, changing comment and a function name has the same work cost, and we get the same readability without touching the code flow unless the code is a duplication that needs to be addressed, and in this code, I would suggest using comments instead of functions.

lib/android/app/src/main/java/com/reactnativenavigation/options/TitleOptions.java
swabbass: we need to clear them, RNN throws error invalidate about having component and text set simultaneously, that means one of component and test must be unset (null, clean) at the same time the other has value.

lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/stack/StackPresenter.java
swabbass: even without comments, the code is self-explaining by itself.

yogevbd (approved):

PR #6926, opened 2021-01-23
Shared Element Transition with different resizeModes (iOS)

mrousavy:
Right now, Shared Element Transitions between two Images with different `resizeMode`s only work on Android.

This use case is quite important, since you often run Shared Element Transitions to open a "details page", and a details page often wants to reveal more of the Image which was previously shown with `resizeMode="cover"`.

So the animation is now: `resizeMode="cover"` -> `resizeMode="contain"`, which works smoothly on Android (props to @guyca) but doesn't animate on iOS.

This PR is an attempt to also implement resizeMode animations for iOS.


mrousavy:
https://user-images.githubusercontent.com/15199031/105601643-6cb14800-5d95-11eb-9f09-1d09d50e388d.mp4

I've got the basic idea working. The idea is, that we switch between the resize modes of the element that I have just tapped, in this case it's a `resizeMode="cover"`. That means, the image size goes beyond the view size, so to correctly animate to `resizeMode="contain"` we need to set the `from` view's resize mode to `contain` as well, and then update it's size so that it looks like nothing has changed between the resize mode switch (`cover` -> `contain`). That means adjusting the width or height, whatever side is the one where the image extends beyond.

The following still needs to be done:

1. The code in `AnimatedUIImageView.m` currently modifies the actual view in the view hierarchy (`element`), and that permanently changes it's `.frame` and `.contentMode` (`width`/`height` and `resizeMode` in react) properties, which we don't want. So we probably need to refactor the code out of there, put it in `AnimatedReactView.m` and once we're done with the animation we can reset it to it's original values in the `- (void)reset` function.
2. It only works for `cover` -> `contain` animations for now, I'm working on supporting all possible combinations.

@yogevbd I will need your help with step 1., since I'm not sure how to approach this. I don't know how I can change the view's actual size and resize mode without the `from` screen looking messed up. As you can see in the video, the `from` screen adjusts with the new values and it looks jumpy.

Also I'm not sure how to respect any applied transforms (e.g. scale), because now that jumps as well.

PR #6930, opened 2021-01-26
Fix a leak in shared element animation

### LINKED ISSUE #6929: Video in Modal prevents the view from being unmounted (continues to play in background)
## 🐛 Bug Report

When showing a Modal with a `<Video>` component from react-native-video, the Video prevents the view from being unmounted on dismiss. I noticed this because my `<Video>` was still playing in the background when I dismissed the modal.

It seems like the [video's `removeFromSuperview` function](https://github.com/react-native-video/react-native-video/blob/master/ios/Video/RCTVideo.m#L1597-L1599) is never being called on `dismissModal()`, which prevents it from cleaning up (i.e. pausing/stopping the video). It does however get successfully called on `pop()`.

To double-check, I tested the same with `push` and `pop`, and in that case the view is actually correctly being unmounted (detached/cleaned/whatever).

In other words, once you have a video playing, this code:

```js
useEffect(() => {
  console.log('mounted');
  return () => {
    console.log('unmounted');
  };
}, []);
```

will only print `mounted` to the console, `unmounted` will never be called.

I created a repro for this on the [repro/dismiss-modal-leak](https://github.com/wix/react-native-navigation/tree/repro/dismiss-modal-leak) branch. To test it, open up the playground, navigate to the "Navigation" tab and press the "Shared Element (Car Dealer)" button. I've added a `<Video>` component in the details screens and you can still here audio coming from it once you close the Modal.

No

## To Reproduce

1. Show modal with audible content (e.g. video, audio)
2. Dismiss modal
3. Notice the modal component still being "mounted" in native view hierarchy (video/audio still playing)

## Expected behavior

I expect it to actually unmount like a normal Screen (`push`) does.

## Actual Behavior

It stays mounted and video/audio stays playing in the background

## Your Environment

- React Native Navigation version: 55b1148745d00450e1546d14c0bdf3aa70367728
- React Native version: 0.63
- Device info (Simulator/Device? OS version? Debug/Release?): all

## Reproducible Demo

https://github.com/wix/react-native-navigation/tree/repro/dismiss-modal-leak

yogevbd:
The shared element animation `transitionContext` is strongly held by the `transitionDelegate` which resulted in a leak. 
@mrousavy Awesome finding!

Closes #6929

### DIFF
--- a/lib/ios/TransitionDelegate.m
+++ b/lib/ios/TransitionDelegate.m
@@ -113,4 +113,8 @@
     return self;
 }
 
+- (void)animationEnded:(BOOL)transitionCompleted {
+    _transitionContext = nil;
+}
+
 @end


mrousavy (approved):
works on my end 👍 🔥

PR #6958, opened 2021-02-11
Bottom tabs shadow default visibility

swabbass:
shadow was enabled by default, changed to opt-in, by having valid shadow options in the BottomTabsOptions


lib/android/app/src/main/java/com/reactnativenavigation/views/bottomtabs/ShadowLayout.kt
```diff
@@ -5,6 +5,7 @@ import android.graphics.*
 import android.graphics.Paint.ANTI_ALIAS_FLAG
 import android.view.View
 import android.widget.FrameLayout
+import com.reactnativenavigation.utils.logd
```
yogevbd, line 8:
```suggestion
```

yogevbd (approved):

PR #6987, opened 2021-02-23
Remove react logbox from button and title components

yogevbd:


mrousavy:
By setting opacity to `0` you can still press it though, no?

yogevbd:
@mrousavy Your'e right, will add negative `zPosition` as well

PR #6991, opened 2021-02-25
Fix android top bar not extending behind status bar

MBrianLoc:
Top bar now extends top portion behind status bar when drawbehind is true and translucent is true instead of leaving default white area

Before: 
Screen Shot 2021-02-25 at 6 28 12 AM

After:
Screen Shot 2021-02-25 at 6 26 11 AM


swabbass:
@MBrianLoc thx looks cool.

Can u, kindly, add unit tests/e2e that cover these flows:
- `setRoot` , `push - pop`, `show/dismiss modals` with changing status bar state across changes.
- Animations and shared transitions, the status bar should change immediately in order to now show jump behaviour.

That will prevent future breaks.

yogevbd:
@swabbass can we merge this?

PR #7032, opened 2021-03-14
[Breaking change] Remove layout.direction

yogevbd:
### Breaking change
`layout.direction` is no longer needed as we can use react-native's I18n to detect the direction.

**Examples**

Forcing the layout direction:
```js
I18nManager.forceRTL(true);
```

Syncing the layout direction with the system's direction:
```js
I18nManager.allowRTL(true);
```


swabbass:
@yogevbd awesome, we might need to mention in docs/changes that calling 
```js
I18nManager.allowRTL(true);
```
should be the first thing before the app starts.

PR #7034, opened 2021-03-15
Upgrade playground react-native's version to 0.64

yogevbd:


swabbass:
I Think also in this upgrade we can bump a major, and drop support for old ReactNative versions in android (Flavors), we might support RN 0.63+ version at least in the android size since it's taking a lot of resources maintaining these backwards compatibility behaviours across RN versions.

@guyca wdyt?

yogevbd:
@swabbass we will drop support for RN versions lower than 0.64 in V8 🔥

birkir:
It would make sense to use hermes on android too, right?

swabbass:
> It would make sense to use Hermes on android too, right?
Hey @birkir :] 
The js side of RNN works with Hermes and JSC, it's up to the host app to use Hermes or other variants, we can enable it by default, but it feels like there is no added value to the lib by enabling it in the Playground module of RNN.

birkir:
> > It would make sense to use Hermes on android too, right?
> 
> 
> 
> Hey @birkir :] 
> 
> The js side of RNN works with Hermes and JSC, it's up to the host app to use Hermes or other variants, we can enable it by default, but it feels like there is no added value to the lib by enabling it in the Playground module of RNN. 

Hey, you probably misunderstood me.

This PR enables hermes on iOS, but not Android and I was only commenting on the inconsistency.

I know the package has nothing to do with Hermes outwards.

swabbass:
> > > It would make sense to use Hermes on android too, right?
> > 
> > 
> > Hey @birkir :]
> > The js side of RNN works with Hermes and JSC, it's up to the host app to use Hermes or other variants, we can enable it by default, but it feels like there is no added value to the lib by enabling it in the Playground module of RNN.
> 
> Hey, you probably misunderstood me.
> 
> This PR enables hermes on iOS, but not Android and I was only commenting on the inconsistency in the Playground app.
> 
> I know the package has nothing to do with Hermes outwards.


Hey sorry, I did not get that from your comment but hey, you have a point here, sure.

swabbass:
## Android changes:
- Remove build flavours (since min RN is 64), files moved under the main src directory.
- Add draw over apps permission (`SYSTEM_ALERT_WINDOW`), blocked Playground.
- Hermes is enabled by default, JSC removed.
- Kotlin upgraded to 1.4.x, now, 1.4.x is a min requirement.

cc: @yogevbd

guyca:
@swabbass Regarding Kotlin version - we should use the same Kotlin version React Native uses to ease integration with RNN.

swabbass:
RN65 is released https://reactnative.dev/blog/2021/08/17/version-065, we need to bump it here.

lib/android/app/src/main/java/com/reactnativenavigation/options/parsers/ColorParser.java
```diff
@@ -0,0 +1,32 @@
...
+import android.content.Context;
+
+import com.facebook.react.bridge.ColorPropConverter;
+import com.reactnativenavigation.options.params.Colour;
+import com.reactnativenavigation.options.params.DontApplyColour;
+import com.reactnativenavigation.options.params.NullColor;
+
+import org.json.JSONObject;
+
+public class ColorParser {
```
swabbass, line 12: this file should be removed, since it is removed in main branch

lib/android/gradlew
```diff
@@ -6,20 +6,38 @@
 ##
 ##############################################################################
 
-# Add default JVM options here. You can also use JAVA_OPTS and GRADLE_OPTS to pass JVM options to this script.
-DEFAULT_JVM_OPTS=""
+# Attempt to set APP_HOME
```
swabbass, line 9: this file should not be pushed

lib/android/gradlew.bat
```diff
@@ -8,20 +8,23 @@
...
 
-@rem Add default JVM options here. You can also use JAVA_OPTS and GRADLE_OPTS to pass JVM options to this script.
-set DEFAULT_JVM_OPTS=
-
 set DIRNAME=%~dp0
 if "%DIRNAME%" == "" set DIRNAME=.
 set APP_BASE_NAME=%~n0
 set APP_HOME=%DIRNAME%
 
+@rem Resolve any "." and ".." in APP_HOME to make it shorter.
```
swabbass, line 16: this file should not be pushed

playground/android/gradlew
```diff
@@ -154,19 +154,19 @@ if [ "$cygwin" = "true" -o "$msys" = "true" ] ; then
...
-        (1) set -- "$args0" ;;
-        (2) set -- "$args0" "$args1" ;;
-        (3) set -- "$args0" "$args1" "$args2" ;;
-        (4) set -- "$args0" "$args1" "$args2" "$args3" ;;
-        (5) set -- "$args0" "$args1" "$args2" "$args3" "$args4" ;;
-        (6) set -- "$args0" "$args1" "$args2" "$args3" "$args4" "$args5" ;;
-        (7) set -- "$args0" "$args1" "$args2" "$args3" "$args4" "$args5" "$args6" ;;
-        (8) set -- "$args0" "$args1" "$args2" "$args3" "$args4" "$args5" "$args6" "$args7" ;;
-        (9) set -- "$args0" "$args1" "$args2" "$args3" "$args4" "$args5" "$args6" "$args7" "$args8" ;;
+        0) set -- ;;
```
swabbass, line 160: this file should not be pushed

playground/android/gradlew.bat
```diff
@@ -13,15 +13,18 @@ if "%DIRNAME%" == "" set DIRNAME=.
 set APP_BASE_NAME=%~n0
 set APP_HOME=%DIRNAME%
 
+@rem Resolve any "." and ".." in APP_HOME to make it shorter.
+for %%i in ("%APP_HOME%") do set APP_HOME=%%~fi
+
 @rem Add default JVM options here. You can also use JAVA_OPTS and GRADLE_OPTS to pass JVM options to this script.
-set DEFAULT_JVM_OPTS=
+set DEFAULT_JVM_OPTS="-Xmx64m" "-Xms64m"
```
swabbass, line 20: this file should not be pushed

playground/android/gradlew
svbutko: It's based on RN 0.65.1's default app

svbutko: https://github.com/react-native-community/rn-diff-purge/blob/release/0.65.1/RnDiffApp/android/gradlew

playground/android/gradlew.bat
svbutko: Same thing, it's based on RN 0.65.1's default app
https://github.com/react-native-community/rn-diff-purge/blob/release/0.65.1/RnDiffApp/android/gradlew

lib/android/app/src/main/java/com/reactnativenavigation/options/parsers/ColorParser.java
svbutko: But it's being used in `ThemeColour.kt`, should it be replaced by something else?

svbutko: It's being used in `ThemeColour.kt` but I think it's being taken from `reactNative62` or somewhere else

svbutko: `import com.reactnativenavigation.options.parsers.ColorParser` import from original file

swabbass: In the main branch today it exists in a Kotlin format, but it's under the `reactNative63`, which was removed here, we need to replace it, I'm adding a note here just to remember to do that before merging,`ib/android/app/src/reactNative63/java/com/reactnativenavigation/options/parsers/ColorParser.kt`.

svbutko: I'll change it

svbutko: Done

playground/android/app/build.gradle
```diff
@@ -1,39 +1,32 @@
...
 android {
-    compileSdkVersion 29
-    ndkVersion "20.1.5948944"
-
-    compileOptions {
-        sourceCompatibility JavaVersion.VERSION_1_8
-        targetCompatibility JavaVersion.VERSION_1_8
-    }
+    compileSdkVersion 30
+    ndkVersion "23.0.7599858"
```
swabbass, line 20: we should stick to RN config https://github.com/facebook/react-native/blob/v0.65.1/gradle.properties#L7

playground/android/build.gradle
```diff
@@ -2,7 +2,7 @@
 
 buildscript {
     ext {
-        kotlinVersion = "1.3.72"
+        kotlinVersion = "1.5.31"
```
swabbass, line 5: we need to stick to RN version for now https://github.com/facebook/react-native/blob/v0.65.1/packages/react-native-gradle-plugin/build.gradle.kts#L11

svbutko: `1.5` fixes a lot of issues, and `1.5.30` most importantly adds support to Apple Silicon.

svbutko: I have 3 apps running on `1.5.31` without any issues

https://kotlinlang.org/docs/whatsnew15.html
https://kotlinlang.org/docs/whatsnew1530.html

swabbass: lets check that it compiles and tests are ok and then we are gonna figure out to keep 1.5 or use 1.4.3

playground/android/app/build.gradle
svbutko: Done

svbutko:
Maybe it makes sense to enable new two params in `gradle.properties` of `lib`?

```
android.useAndroidX=true
android.enableJetifier=true
```

Since it uses AndroidX features.

svbutko:
Btw I also think it would be great to fix #7246 while making all the other changes.
I can start changing it.

svbutko:
Good news, everyone!

RN 0.66.0 released:
https://reactnative.dev/blog/2021/10/01/version-066

I'll create a separate PR with update and will merge it here the same way I did with 0.65.1, wdyt?

svbutko:
Updated my apps and tested them on it, works fine

playground/android/build.gradle
svbutko: At least it passes all tests on Android 😄
https://github.com/facebook/react-native/pull/32319

swabbass:
It seems Cool you are doing great :] so Big thanks for such effort.

The e2e fails, on some bundler stuff. 
I'm finishing some stuff with other PR and moving on to this.

playground/android/build.gradle
svbutko: Added another PR to RN regarding NDK: https://github.com/facebook/react-native/pull/32321
It looks ok too, at least according to tests, hope it will be merged soon, it makes bundle smaller

svbutko:
Happy to help, I really like this project and how it evolves

svbutko:
What about dropping support for iOS 11 and iOS 12?
And keeping last 3 major versions (13,14,15)?

It will remove good part of the code and separated tests

svbutko:
https://developer.apple.com/support/app-store/

With iOS 15 release these numbers will go even lower

yogevbd:
@svbutko I need to check with our production team if they plan on removing support for iOS 12 in the Wix app. If so we defiantly can

svbutko:
There are several requirements to upgrade AGP properly to `7.0.2`:
https://github.com/facebook/react-native/pull/32186

svbutko:
And if it's going to be AGP `7.0.2` then I would mention how to install JDK 11:

`brew install --cask adoptopenjdk/openjdk/adoptopenjdk11`

Since most of the users have JDK 8 only installed

swabbass:
> And if it's going to be AGP `7.0.2` then I would mention how to install JDK 11:
> 
> `brew install --cask adoptopenjdk/openjdk/adoptopenjdk11`
> 
> Since most of the users have JDK 8 only installed

yup, we can mention that it is already running e2e locally in a release.

svbutko:
> > > Users can choose to have such `gradle.properties`:
> > > ```
> > > # For adoptopenjdk: brew install --cask adoptopenjdk/openjdk/adoptopenjdk11
> > > #org.gradle.java.home=/Library/Java/JavaVirtualMachines/adoptopenjdk-11.jdk/Contents/Home
> > > 
> > > # For Android Studio embedded JVM version
> > > org.gradle.java.home=/Applications/Android Studio.app/Contents/jre/Contents/Home
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > @svbutko wdyt?
> > 
> > 
> > LGTM, but I think it's worth mentioning Windows users too:
> > `choco install -y nodejs.install openjdk11`
> > I myself on macOS but I know many people use Windows for development
> 
> Showing love for **Windows** devs.

I started working with RN around 5 years ago on Windows (transitioned to macOS in 2019) and I went through all of the possible issues so I know what it's like 😄

PR #7036, opened 2021-03-16
Android mergeOptions subtitle center & margin fix

swabbass:
## Issue:

when setting default alignment in default options it wasn't being respected when `mergeOptions` called with options that do not include alignment, like updating the text or colour, it was resetting the alignment.

## Fix:

- Use the result of default options merged with requested options to update
- Ensure that alignment change should apply to the container when changing subtitle

CC: @danilobuerger


danilobuerger (approved):
Works on my production app. Thanks!

PR #7045, opened 2021-03-21
Support shared element transition for images with different resizeMode

yogevbd:
This PR adds support for animating images with different `resizeMode` which already supported on Android and was missing on iOS.


mrousavy:
Hmm, looks like this breaks the transition with transforms applied (translateX): 

https://user-images.githubusercontent.com/15199031/111959063-476b5c80-8aee-11eb-9790-4c4541e1eb66.MP4

mrousavy (changes requested):
Somehow the playground does not animate correctly:

https://user-images.githubusercontent.com/15199031/112462860-d7650c80-8d61-11eb-89d3-892c53185a2e.mov

yogevbd:
We had a missing treatment for `resizeMode: 'contain'`

https://user-images.githubusercontent.com/10794586/112466023-0aad9880-8d6e-11eb-95d3-a1614347bb2d.mov

guyca:
LGTM!

mrousavy:
I'm still seeing some very weird behaviour regarding translation. Note that my animation worked before this PR (just not with correct resizeMode), and works fine on Android, so it definitely is a change in this PR, I'll try to reproduce again in a few hours

mrousavy (changes requested):
1. Transitioning from `<Video>` -> `<Video>` works, `<FastImage>` -> `<Video>` doesn't work (this might be a common approach, since we don't want to load the entire Video just to display the poster/thumbnail. **This does work on Android**)
2. [the memory leak](https://github.com/wix/react-native-navigation/issues/6929) appeared again  (regression)

I feel like this can be merged if 2. is fixed, but since 1. has actually worked on iOS before with same resizeModes (and does work on Android too), this PR introduces a minor regression

mrousavy (approved):

Issue #7044, opened 2021-03-20
Shared element transition using spring interpolation is jumpy at the end (only Android)


nelsonprsousa:
## 🐛 Bug Report

The spring interpolation is a bit jumpy at the end of the animation. Only on Android (looks okay on iOS).

### Have you read the [Contributing Guidelines on issues](https://wix.github.io/react-native-navigation/docs/meta-contributing/)?
Yes.

## To Reproduce

1. Clone [this repo ](https://github.com/nelsonprsousa/rnn-testing/tree/set-spring-bug)
2. Go to `set-spring-bug` branch
2. yarn
3. yarn android

## Expected behavior
The animation should happens like in iOS (it is a little bit jumpy in Android).

## Actual Behavior
Looks like there are a pixel left to animate
Here's the iOS and Android animation:

https://user-images.githubusercontent.com/8364345/111888416-2dd5f200-89d4-11eb-8eff-45f453c1eed8.mov

## Your Environment

- React Native Navigation version: 7.8.4
- React Native version: 0.63.4
- Platform(s) (iOS, Android, or both?): Android

## Reproducible Demo
1. Clone [this repo ](https://github.com/nelsonprsousa/rnn-testing/tree/set-spring-bug)
2. Go to `set-spring-bug` branch

nelsonprsousa:
I also noticed that if we use `mass: 1` the bug is no longer reproducible. 

Here's a `mass: 50` example:
![Screenshot_1616283251](https://user-images.githubusercontent.com/8364345/111888478-cff5da00-89d4-11eb-87e9-ea2edd09dcb1.png)

guyca:
@swabbass this is probably related to WOARNN-386

swabbass:
> WOARNN-386

I need to try that and see, think it could be

swabbass:
@nelsonprsousa I tried to run your demo, and it just crashes right away when I click on the card, that was on `7.8.4` and on the latest version (7.11.3) it crashes on parsing options, which was fixed here #7046.

so now using master fix.

it seems that the `SpringInterpolator` is mostly inspired by [this](https://github.com/marmelroy/Interpolate/blob/master/Interpolate/SpringInterpolation.swift), when i modified the animation to match what is in the file the bug does not reproduce, seems mass, damping, stiffness affects the end view size some how, I'm looking into it :]

nelsonprsousa:
[...]

PR #7055, opened 2021-03-24
Ensure spring animation always completes

swabbass:
# Issue:

`SpringInterportolator` did not finish when the animation finished, just complied with the forces attached to the view, which caused unwanted padding and jerky animations.

# Fix:

When animation time reaches the end, `1`, just return `1`  as the interpolated value, since it indicates that the interpolated progress is 100%.

Closes : #7044


mrousavy (approved):
I still feel like this is a workaround instead of a fix. This just means that the spring formula is slightly wrong, and with your fix it will now jump to `1f` at the end instead of smoothly animating to it. The larger the translation, the more it jumps.

I guess we should look into the spring interpolation again and maybe rewrite it - I remember back then I had to write it like this because SET interpolations are always time based and a spring interpolation is not a time based animation, so it was hard to find a spring interpolation that could execute under the given time. I tried to use the iOS transition spec as well as the reanimated ones, but none of those worked. 

TL;DR: For now it _works_, we'll have to find a real fix for it soon.

guyca:
I agree with Marc's conclusion. I suggest we evaluate https://developer.android.com/guide/topics/graphics/spring-animation

swabbass:
@mrousavy i agree with you it is a workaround, since you are the one who added SpringInterpolator,  you know from the beginning that SpringInterpolator copied from ios to android without making correct calculations, might lead to this behaviour, in which leads to this workaround because the solution it self does not fit in the android environment, the simplest example just reproduced the bug,  jump in the animation and not being ended,  that means probably the feature weren't checked as as it should've been :).

I'll consider a proper way to use Spring Animations. 👍

mrousavy:
> .. you know from the beginning that SpringInterpolator copied from ios to android without making correct calculations, might lead to this behaviour, ..

That's simply not true. It's the same on iOS as it is on Android, both of those are just a simple spring function which are expected to complete within a given timeframe. The only reason you're seeing this bug is because Android apparently doesn't force-set the value to 100% at the end, while iOS does. (Not just for the Spring interpolation, but apparently for every animation/interpolation).

Also, this problem only occurs with incorrectly tuned spring config values, so e.g. if `mass: 10` and damping is pretty high, this problem can be seen more drastically. In this case the user has full control over the spring configuration, so that he can tune it to nicely finish within the desired timeframe.
I believe `mass: 1` and a bit lower stiffness/damping works perfectly and lets the spring animation finish within `500ms` (push animation duration in my case)

This means technically this wasn't a bug in the first place. Just a platform inconsistency. You can test this if you always return `0.5f` from the linear interpolator. On iOS it is `0` -> `0.5` -> `1.0`, on Android it is `0` -> `0.5`. (Haven't tested that, I might be wrong about this)

swabbass:
> > .. you know from the beginning that SpringInterpolator copied from ios to android without making correct calculations, might lead to this behaviour, ..
> 
> That's simply not true. It's the same on iOS as it is on Android, both of those are just a simple spring function which are expected to complete within a given timeframe. The only reason you're seeing this bug is because Android apparently doesn't force-set the value to 100% at the end, while iOS does. (Not just for the Spring interpolation, but apparently for every animation/interpolation).
> 
> Also, this problem only occurs with incorrectly tuned spring config values, so e.g. if `mass: 10` and damping is pretty high, this problem can be seen more drastically. In this case the user has full control over the spring configuration, so that he can tune it to nicely finish within the desired timeframe.
> I believe `mass: 1` and a bit lower stiffness/damping works perfectly and lets the spring animation finish within `500ms` (push animation duration in my case)
> 
> This means technically this wasn't a bug in the first place. Just a platform inconsistency. You can test this if you always return `0.5f` from the linear interpolator. On iOS it is `0` -> `0.5` -> `1.0`, on Android it is `0` -> `0.5`. (Haven't tested that, I might be wrong about this)

It is, I just googled snippets of the code, that was added in`SpringInterportolator`, it is 95% matching this, https://github.com/marmelroy/Interpolate/blob/master/Interpolate/SpringInterpolation.swift , which is based over https://github.com/robb/RBBAnimationm, you can take a look/

using a code logic from os to other might not work as you can expect.

animations on android are ended to 100%, the interpolator was written there just does not work as other interpolators, in the system, that means the os components are ok, the `SpringInterpolator` implementation is not ok, for sure animations end up normally and get to 100% in android, also how did you manage to get to the conclusion that android won't reach 100%.

the example you described above, is also jumpy on ios, since from interpolation of `0.5` to `1` that means the animation now at the end value, hence, from 50% value to 100%, can you explain to me why there is no jump on ios?  I don't get it how android stock animations just works normally and this interpolator detects that animations on android does not reach 100%.

Like dude, My whole point is, the code is not fit for a solution in android, it can be that this interpolator or system animations are really not working as they should.

mrousavy:
As referenced in [this line](https://github.com/wix/react-native-navigation/blob/master/lib/ios/Interpolators/SpringInterpolator.m#L31), the code is taken from [here](https://github.com/robb/RBBAnimation/blob/master/RBBAnimation/RBBSpringAnimation.m#L60-L107) and modified a bit. 

On both platforms, this is a simple function that receives an input (`progress`, from `0.0` to `1.0`, aka percent), and returns an interpolated output ("pure function"). That means the only factor that influences this function is the input, which is different on Android and on iOS. 

With an input (`t`, or `progress`) of `0.6`, both iOS and Android return the exact same value. (`0.9560279`).

Here are the logs in form of `[input value] -> [return value]`:

Android:

```
0.0   -> 0.0
0.0   -> 0.0
0.132 -> 0.2928254
0.132 -> 0.2928254
0.232 -> 0.564621
0.232 -> 0.564621
0.366 -> 0.799109
0.366 -> 0.799109
0.4   -> 0.83721304
0.4   -> 0.83721304
0.532 -> 0.9305975
0.532 -> 0.9305975
0.6   -> 0.9560279
0.6   -> 0.9560279
0.666 -> 0.97200334
0.666 -> 0.97200334
0.766 -> 0.98605675
0.766 -> 0.98605675
0.832 -> 0.9912613
0.832 -> 0.9912613
0.932 -> 0.995733
0.932 -> 0.995733
0.998 -> 0.99735487
0.998 -> 0.99735487
```

iOS:

```
0.108092 -> 0.221146
0.108092 -> 0.221146
0.108092 -> 0.221146
0.108092 -> 0.221146
0.108092 -> 0.221146
0.137472 -> 0.309183
0.137472 -> 0.309183
0.137472 -> 0.309183
0.137472 -> 0.309183
0.137472 -> 0.309183
0.170760 -> 0.406180
0.170760 -> 0.406180
0.170760 -> 0.406180
0.170760 -> 0.406180
0.170760 -> 0.406180
0.204104 -> 0.496280
0.204104 -> 0.496280
0.204104 -> 0.496280
0.204104 -> 0.496280
0.204104 -> 0.496280
0.237428 -> 0.577117
0.237428 -> 0.577117
0.237428 -> 0.577117
0.237428 -> 0.577117
0.237428 -> 0.577117
0.270622 -> 0.647769
0.270622 -> 0.647769
0.270622 -> 0.647769
0.270622 -> 0.647769
0.270622 -> 0.647769
0.304036 -> 0.709080
0.304036 -> 0.709080
0.304036 -> 0.709080
0.304036 -> 0.709080
0.304036 -> 0.709080
0.337444 -> 0.761187
0.337444 -> 0.761187
0.337444 -> 0.761187
0.337444 -> 0.761187
0.337444 -> 0.761187
0.370862 -> 0.805006
0.370862 -> 0.805006
0.370862 -> 0.805006
0.370862 -> 0.805006
0.370862 -> 0.805006
0.404194 -> 0.841430
0.404194 -> 0.841430
0.404194 -> 0.841430
0.404194 -> 0.841430
0.404194 -> 0.841430
0.437478 -> 0.871528
0.437478 -> 0.871528
0.437478 -> 0.871528
0.437478 -> 0.871528
0.437478 -> 0.871528
0.470790 -> 0.896299
0.470790 -> 0.896299
0.470790 -> 0.896299
0.470790 -> 0.896299
0.470790 -> 0.896299
0.504132 -> 0.916575
0.504132 -> 0.916575
0.504132 -> 0.916575
0.504132 -> 0.916575
0.504132 -> 0.916575
0.537548 -> 0.933109
0.537548 -> 0.933109
0.537548 -> 0.933109
0.537548 -> 0.933109
0.537548 -> 0.933109
0.570832 -> 0.946456
0.570832 -> 0.946456
0.570832 -> 0.946456
0.570832 -> 0.946456
0.570832 -> 0.946456
0.604180 -> 0.957257
0.604180 -> 0.957257
0.604180 -> 0.957257
0.604180 -> 0.957257
0.604180 -> 0.957257
0.637512 -> 0.965948
0.637512 -> 0.965948
0.637512 -> 0.965948
0.637512 -> 0.965948
0.637512 -> 0.965948
0.670860 -> 0.972927
0.670860 -> 0.972927
0.670860 -> 0.972927
0.670860 -> 0.972927
0.670860 -> 0.972927
0.704214 -> 0.978514
0.704214 -> 0.978514
0.704214 -> 0.978514
0.704214 -> 0.978514
0.704214 -> 0.978514
0.737528 -> 0.982971
0.737528 -> 0.982971
0.737528 -> 0.982971
0.737528 -> 0.982971
0.737528 -> 0.982971
0.770898 -> 0.986530
0.770898 -> 0.986530
0.770898 -> 0.986530
0.770898 -> 0.986530
0.770898 -> 0.986530
0.804218 -> 0.989355
0.804218 -> 0.989355
0.804218 -> 0.989355
0.804218 -> 0.989355
0.804218 -> 0.989355
0.837550 -> 0.991600
0.837550 -> 0.991600
0.837550 -> 0.991600
0.837550 -> 0.991600
0.837550 -> 0.991600
0.870912 -> 0.993381
0.870912 -> 0.993381
0.870912 -> 0.993381
0.870912 -> 0.993381
0.870912 -> 0.993381
0.904212 -> 0.994787
0.904212 -> 0.994787
0.904212 -> 0.994787
0.904212 -> 0.994787
0.904212 -> 0.994787
0.937500 -> 0.995899
0.937500 -> 0.995899
0.937500 -> 0.995899
0.937500 -> 0.995899
0.937500 -> 0.995899
0.970854 -> 0.996779
0.970854 -> 0.996779
0.970854 -> 0.996779
0.970854 -> 0.996779
0.970854 -> 0.996779
1.000000 -> 0.997393
1.000000 -> 0.997393
1.000000 -> 0.997393
1.000000 -> 0.997393
1.000000 -> 0.997393
```


Notice how Android does not call the function for `1.0` again. The last value is `0.998`. On the other side, iOS does not call the func for `0.0`. The first value is `0.108092`.

So your statement

> animations on android are ended to 100%

is wrong.

> can you explain to me why there is no jump on ios?

Yes. Again, iOS calls with input `1.0`. Android doesn't.


Now that we know that android doesn't call the interpolator with input `1f` I'm wondering if the fix in this PR even works for every case, since the `if (t == 1f)` would never execute in my case (because `t` only gets to `0.998`) 🤔

mrousavy:
I see two possible solutions for this:

1. Don't do anything at all. The user has full control over the spring config, and we can provide pre-configured spring interpolations the user can use. Everyone can tweak the config to be as light or as stiff as they want.
2. Find a way to run the spring animation without it depending on the timing animation (e.g. the push screen animation). This means, if the spring animation takes really long to complete, the screen might already be pushed but the animation is still playing. This ensures that there will be no jump, and that the spring animation completes. Also the user can push the screen and let the animation finish first before the screen is even fully pushed. Might also be cool for UI/UX.

swabbass:
@mrousavy 

about my statement see this, the android animation progress reaches 1, checked in rnn-test
Screen Shot 2021-03-24 at 15 10 56

mrousavy:
Huh, I must've logged it wrong. My bad.

But my point still stands, you can see that it outputs `0.9973931`, which is not `1`. On iOS, it apparently force-sets the output value to `1` after the animation has been completed, which it does not automatically do on Android (your PR does this now explicitly for the SpringInterpolator)

mrousavy:
The real solution here is to use a Spring Configuration which returns `1` as closely as possible at the end. If you find the holy formula for the spring config we can use that as a default

swabbass:
All fine dude, it happens to me also :].

i see from your logs on ios, that when progress hits `1` it interpolates `0.97` and its expected since they have the same calculations of the interpolation value.

Now about the interpolator itself, both on ios and android will have such jumps if the final interpolation is changed (by the os or manually) from `0.997` to `1.0` which is a `0.003` of progress, which would be seamless, i would try to configure mass, stiffness, damping, so it can jump a bigger step, like at progress 1 it interpolates `0.8` , so we can have 20%, a detectable change, we or the os set the interpolation for 1, and then i assume that it would "jump" on both android and ios.
Screen Shot 2021-03-24 at 15 16 40

swabbass:
my suggestion, which is an effort, not that big, is to use os defined ones, and maybe for spring we use the native os component, but the RNN JS API might change a bit.

PR #7059, opened 2021-03-25
Don't reload when textInput is focused and user is typing rr

guyca:
When a textInput is focused and the developer quickly types a word that has two R's in it - don't reload RN as it's not the developers intention.


swabbass (approved):
looks fine :] THX

ejain:
Looks like this pull request broke reactNative != 63? See #7112.

PR #7074, opened 2021-04-06
Stack declared children crash on duplicate id

swabbass:
# Issue
Calling setRoot with children that have the same ids,  caused pop to crash, not aligned with `Navigation.push` behaviour
```javascript
  setRoot = async () =>
    await Navigation.setRoot({
      root: {
        stack: {
          id: 'stack',
          children: [
            {
              component: {
                id: 'component',
                name: Screens.A,
              },
            },
              {
              component: {
                id: 'component',
                name: Screens.A,
              },
            },
          ],
        },
      },
    });

```

# Fix

Since such behaviour cannot be ignored, pop will always fail if we just ignore the ids, the fix is to be aligned with `push` command.


guyca:
Nice! 🔥

yogevbd (approved):

Issue #7070, opened 2021-04-03
SET dismiss modal regression in latest version (7.12.0)


nelsonprsousa:
## 🐛 Bug Report
While upgrading from 7.8.4 to the latest version 7.12.0, shared element transition no longer works as expected while dismissing modals. Also tested with master (cef992918cce652427d85d6a52e8d9ded65b95c6), same problem.

### Have you read the [Contributing Guidelines on issues](https://wix.github.io/react-native-navigation/docs/meta-contributing/)?
Yes.

## To Reproduce

1. Clone [this repo](https://github.com/nelsonprsousa/rnn-testing/tree/set-regression)
2. Use `set-regression` branch
3. yarn
4. yarn ios / yarn android
5. The last commit is the upgrade to 7.12.0. You can just drop the last commit to go back to 7.8.4, without this problem (reinstall pods and gradlew clean for Android).

## Expected behavior
The background screen should appear as before when we dismiss the modal.

## Actual Behavior
Looks like the screen just appears once the SET is finished.

## Your Environment

- React Native Navigation version: 7.12.0
- React Native version: 0.63.4
- Platform(s) (iOS, Android, or both?): both
- Device info (Simulator/Device? OS version? Debug/Release?): Android and iOS

## Reproducible Demo

7.8.4:

https://user-images.githubusercontent.com/8364345/113493667-6a781200-94d9-11eb-8a50-d68d7b6aaeca.mov

7.12.0:

https://user-images.githubusercontent.com/8364345/113493704-c347aa80-94d9-11eb-8568-974f0d3cffbd.mov

nelsonprsousa:
Also noticed that on Android the SET while showing the modal is no longer working with the latest version.

nelsonprsousa:
Hey @guyca and @swabbass 
I don't think the PR that was merged completely solves this issue. On dismiss model, looks like the background isn't correctly showed. I tried to replicate it with slow motion here (using current snapshot: 7.13.0-snapshot.1515):

https://user-images.githubusercontent.com/8364345/114719786-54671e80-9d2f-11eb-8a6f-3d057b014378.mov

guyca:
@nelsonprsousa The PR fixed an issue on Android while the video shows iOS 🤔

nelsonprsousa:
> @nelsonprsousa The PR fixed an issue on Android while the video shows iOS 🤔
Check the last video on the original video. It's both for Android and iOS.

[...]

PR #7081, opened 2021-04-11
Shared element android fix

swabbass:
# Issue

Showing modal with shared element transition broke, in the opening direction, while it is working in the closing direction, like when opening a card in the Car dealer in the Playground, the shared card won't animate to the new position, fading instead.

https://user-images.githubusercontent.com/7227793/114307698-7cfdd700-9ae9-11eb-8f4c-d4ca4235e9db.mov

# Fix

Check and test that SharedElement transition is playing, matching `push` shared element behaviour, making `showModal` and `push` aligned since push does not suffer such misbehaving in shared element.

https://user-images.githubusercontent.com/7227793/114307791-c2ba9f80-9ae9-11eb-9068-8931a8238994.mov

Closes: #7070

### DIFF
--- a/lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/modal/ModalAnimator.kt
+++ b/lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/modal/ModalAnimator.kt
@@ -4,10 +4,13 @@ import android.animation.Animator
 import android.animation.AnimatorListenerAdapter
 import android.animation.AnimatorSet
 import android.content.Context
+import androidx.annotation.VisibleForTesting
 import com.reactnativenavigation.options.FadeInAnimation
 import com.reactnativenavigation.options.FadeOutAnimation
 import com.reactnativenavigation.options.TransitionAnimationOptions
+import com.reactnativenavigation.options.params.Bool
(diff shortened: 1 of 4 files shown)


lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/modal/ModalAnimator.kt
```diff
@@ -80,7 +105,6 @@ open class ModalAnimator @JvmOverloads constructor(
         set.addListener(object : AnimatorListenerAdapter() {
             private var isCancelled = false
             override fun onAnimationStart(animation: Animator) {
-                runningAnimators[appearing] = animation
```
guyca, line 83: This isn't needed anymore?

guyca (changes requested):

lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/modal/ModalAnimator.kt
swabbass: moved to the `show` function.

guyca (approved):

PR #7089, opened 2021-04-16
docs: Update option-processor rightButtons example

dan-trewin:
Update option-processor-defaults example to work. The current example wouldn't work properly or do anything b/c an array of rightButtons is returned to the processor, not a single button.


guyca (approved):

guyca:
Thanks!

Issue #7037, opened 2021-03-16
Title is not displayed properly.


leralera2201:
If the title is long, it covers  right buttons.
<img width="538" alt="Screenshot 2021-03-16 at 13 24 04" src="https://user-images.githubusercontent.com/58082566/111338710-cfc9a780-867f-11eb-8e90-b3b331a7ea7f.png">

swabbass:
@leralera2201 it seems ugly, but I can explain a bit, there are apps that might want to cut it after a certain count of characters, maybe replace it with "..." or apply any other solution, so it is more like a product thing, and it could be per app, also take into consideration real use, no one would fit such long title, and if there is a case, you can change `fontSize` in the code, and have full control over each title.

yogevbd:
@swabbass it should cut the end of the string.. look like we have a wrong calculation there.

guyca:
@swabbass Notice that Android's behaviour is to collapse the buttons to make room for the title, and only when all buttons are collapsed into the overflow menu icon (three dots button) the title is elipsized.

PR #7096, opened 2021-04-21
Fix title measurements and layout in toolbar

swabbass:
# Issue:

Passing long text in the `topBar.{sub}title.text` field, caused title overflow and misplacement of other views.

Screenshot 2021-03-16 at 13 24 04

See #7037.

# Fix

The way to solve this is to take control of the system and measure/layout the child views as wanted, this is a work in progress, but for now:
- [x] Implement `onMeasure` to set width of each child of toolbar.
- [x] Implement `onLayout` to layout the title according to the measurements.
- [x] Show overflow when no space for buttons.
- [ ] Add testing and checks for more edge cases.

Actual demo of the solution:

https://user-images.githubusercontent.com/7227793/115494247-ee6a2200-a26d-11eb-8668-298ca9ce7b05.mov

NOTE: This is a draft, still a lot of refinements will be pushed gradually.

Closes: #7037.


lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/MainToolBar.kt
```diff
@@ -160,6 +142,69 @@ class MainToolBar(context: Context) : RelativeLayout(context) {
...
+        } else {
+            titleComponent.layout(l + leftButtonsBar.measuredWidth + defaultMargin, t, r - rightButtonsBar.measuredWidth, b)
+        }
+    }
+
+    override fun onMeasure(widthMeasureSpec: Int, heightMeasureSpec: Int) {
+        val mode = MeasureSpec.getMode(widthMeasureSpec)
+        val parentWidth = MeasureSpec.getSize(widthMeasureSpec)
+        val parentHeight = MeasureSpec.getSize(heightMeasureSpec)
+        logd("OnMeasure: parentWidth: $parentWidth, parentHeight: $parentHeight widthMeasureSpec:" +
```
guyca, line 162: Don't forget to remove logs when ready. Perhaps wrap it with [Log.isLoggable](https://developer.android.com/reference/android/util/Log#isLoggable(java.lang.String,%20int)) just in case.

lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/MainToolBar.kt
```diff
@@ -160,6 +142,69 @@ class MainToolBar(context: Context) : RelativeLayout(context) {
...
+                " ${MeasureSpec.toString(widthMeasureSpec)}, heightMeasureSpec: ${MeasureSpec.toString(heightMeasureSpec)}", "MainToolBar")
+        super.onMeasure(widthMeasureSpec, heightMeasureSpec)
+        // exact measurement of the parent
```
guyca, line 165: Why is this comment needed? `mode == MeasureSpec.EXACTLY` isn't too hard to understand.

lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/MainToolBar.kt
```diff
@@ -160,6 +142,69 @@ class MainToolBar(context: Context) : RelativeLayout(context) {
...
+        if (mode == MeasureSpec.EXACTLY) {
+            val titleComponent = getTitleComponent()
+            // measure all 3 child with AT_MOST to get what space they require
```
guyca, line 168:
```suggestion
            // measure all 3 children with AT_MOST to get the width they require
```

lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/MainToolBar.kt
```diff
@@ -160,6 +142,69 @@ class MainToolBar(context: Context) : RelativeLayout(context) {
...
+        if (titleComponentAlignment == Alignment.Center) {
+            val titleLeft = max(l + leftButtonsBar.measuredWidth, (r - l) / 2 - titleComponent.measuredWidth / 2)
+            val titleRight = min(r - rightButtonsBar.measuredWidth, (r - l) / 2 + titleComponent.measuredWidth / 2)
+            titleComponent.layout(titleLeft, t, titleRight, b)
+        } else {
+            titleComponent.layout(l + leftButtonsBar.measuredWidth + defaultMargin, t, r - rightButtonsBar.measuredWidth, b)
+        }
+    }
+
+    override fun onMeasure(widthMeasureSpec: Int, heightMeasureSpec: Int) {
```
guyca, line 158: This function is very long - let's extract the measurement responsibilities to a Measurer class.

lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/MainToolBar.kt
```diff
@@ -160,6 +142,69 @@ class MainToolBar(context: Context) : RelativeLayout(context) {
...
+        if (mode == MeasureSpec.EXACTLY) {
+            val titleComponent = getTitleComponent()
+            // measure all 3 child with AT_MOST to get what space they require
+            rightButtonsBar.measure(MeasureSpec.makeMeasureSpec(parentWidth, MeasureSpec.AT_MOST), heightMeasureSpec)
+            leftButtonsBar.measure(MeasureSpec.makeMeasureSpec(parentWidth, MeasureSpec.AT_MOST), heightMeasureSpec)
+            titleComponent.measure(MeasureSpec.makeMeasureSpec(parentWidth, MeasureSpec.AT_MOST), heightMeasureSpec)
+
+            val titleComponentWidth = titleComponent.measuredWidth
+
+            // hide buttons title bypass the right end, otherwise show
```
guyca, line 175: I'm sorry, but I have no idea what this comment means. Since you're going to extract this logic and then split it into smaller functions that are easier to understand, we won't need this comment after we refactor.

lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/MainToolBar.kt
```diff
@@ -160,6 +142,69 @@ class MainToolBar(context: Context) : RelativeLayout(context) {
...
+            var leftBarWidth = leftButtonsBar.measuredWidth
+            val titleLeft = parentWidth / 2 - (titleComponentWidth / 2)
+            if (titleLeft < this.left + leftBarWidth + defaultMargin) {
```
guyca, line 178: Changing the `showAsAction` property from a function intended to measure views is wrong. This side effect is undesirable as it makes the code very difficult to reason about.

lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/MainToolBar.kt
```diff
@@ -160,6 +142,69 @@ class MainToolBar(context: Context) : RelativeLayout(context) {
...
+                for (item in leftButtonsBar.menu) {
+                    item.setShowAsAction(MenuItem.SHOW_AS_ACTION_NEVER)
+                }
+            }else{
+                for (item in leftButtonsBar.menu) {
+                    item.setShowAsAction(MenuItem.SHOW_AS_ACTION_IF_ROOM)
+                }
+            }
+
+            // hide buttons title bypass the left end, otherwise show
```
guyca, line 189: Duplicate code - duplicate comments 😅

lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/MainToolBar.kt
```diff
@@ -160,6 +142,69 @@ class MainToolBar(context: Context) : RelativeLayout(context) {
...
+                for (item in rightButtonsBar.menu) {
+                    item.setShowAsAction(MenuItem.SHOW_AS_ACTION_NEVER)
+                }
+            }else{
+                for (item in rightButtonsBar.menu) {
+                    item.setShowAsAction(MenuItem.SHOW_AS_ACTION_IF_ROOM)
+                }
+            }
+
+            // calculate and measure the title component for the last time.
```
guyca, line 203: Instead of commenting **what** the code does - extract these to lines to a function and name it `calculateAndMeasureTitleComponent()`.

lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/MainToolBar.kt
```diff
@@ -15,15 +15,25 @@ import com.reactnativenavigation.utils.CompatUtils
 import com.reactnativenavigation.utils.UiUtils
 import com.reactnativenavigation.utils.ViewUtils
 import com.reactnativenavigation.utils.logd
+import kotlin.math.max
+import kotlin.math.min
 
 const val DEFAULT_LEFT_MARGIN = 16
+const val SYSTEM_ACTION_BUTTON_SIZE_DP = 48
 
 class MainToolBar(context: Context) : RelativeLayout(context) {
```
guyca, line 24: Since `MainToolBar` extends `RelativeLayout`, not `Toolbar` - I'd rename the class to something else. It's basically a `TitleAndButtonsContainer`. Its purpose is to layout and measure the title and buttons.

lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/TitleSubTitleLayout.kt
```diff
@@ -15,10 +15,19 @@ import com.reactnativenavigation.options.Alignment
 import com.reactnativenavigation.options.FontOptions
 import com.reactnativenavigation.options.parsers.TypefaceLoader
 
+
 class TitleSubTitleLayout(context: Context) : LinearLayout(context) {
 
-    private val titleTextView = TextView(context).apply { TextViewCompat.setTextAppearance(this, R.style.TitleBarTitle) }
-    private val subTitleTextView = TextView(context).apply { TextViewCompat.setTextAppearance(this, R.style.TitleBarSubtitle) }
+    private val titleTextView = TextView(context).apply {
```
guyca, line 21: Would be nice to declare the layout in XML and move all of these boilerplate to the XML. Then you can set the layout like so:

```kotlin
init {
    LayoutInflater.from(context).inflate(R.layout.title_layout, this)
}
```

guyca (changes requested):

lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/MainToolBar.kt
```diff
@@ -15,15 +15,25 @@ import com.reactnativenavigation.utils.CompatUtils
...
     private var component: View? = null
-    private val centerFrameLayout = FrameLayout(context).apply {
-        id = CompatUtils.generateViewId()
-    }
+
+    private var titleComponentAlignment: Alignment = Alignment.Default
+        set(value) {
+            if (field != value) {
+                field = value
+                requestLayout()
```
guyca, line 32: Nice!

lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/MainToolBar.kt
```diff
@@ -160,6 +142,69 @@ class MainToolBar(context: Context) : RelativeLayout(context) {
...
+        super.onLayout(changed, l, t, r, b)
+        // Bypass regular layout for the titleComponent to make final placements based on the measurement
+        val titleComponent = getTitleComponent()
+        if (titleComponentAlignment == Alignment.Center) {
+            val titleLeft = max(l + leftButtonsBar.measuredWidth, (r - l) / 2 - titleComponent.measuredWidth / 2)
+            val titleRight = min(r - rightButtonsBar.measuredWidth, (r - l) / 2 + titleComponent.measuredWidth / 2)
+            titleComponent.layout(titleLeft, t, titleRight, b)
+        } else {
+            titleComponent.layout(l + leftButtonsBar.measuredWidth + defaultMargin, t, r - rightButtonsBar.measuredWidth, b)
+        }
```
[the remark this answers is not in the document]
swabbass: Agree 100%, for now, the comments are just used for me to track the progress of what was done, in terms of flows, those will be removed, and have a better design

lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/MainToolBar.kt
swabbass: Its not needed, soon will be removed

lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/MainToolBar.kt
swabbass: the comment should be removed and replaced with a function with proper naming :]

lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/MainToolBar.kt
swabbass: Sure

lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/MainToolBar.kt
swabbass: Agree 100%, the decision about to shrink buttons or not is relying on the 3 measurements, maybe the place in `onMeasure` is not the right place, but it's the closest (fastest) place to get those measurements right, we can add the `doOnLayout` block to run after all these calculations are ready, that means, we should trigger that on title changes (text, alignment..), buttons changes ( add/remove/change text/ icon/ custom components).

 I thought about that, this is the closest position to what we need to achieve, WDYT about alternatives?

lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/MainToolBar.kt
swabbass: I left the design as-is for now, to think about the design and enhancing the code, I'm bad at naming stuff so I left them laid like this in the code, while I'm having my time designing it to look right :]

lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/TitleSubTitleLayout.kt
swabbass: we adding one step of parsing the XML to the process, which does not have any added value, but, the style as I see don't have a special thing, I can do it in code.

lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/MainToolBar.kt
guyca: Measuring the title bar and buttons bar should be enough for them to not overlap one another. Changing `showAsAction` should not be needed.

lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/TitleSubTitleLayout.kt
guyca: > which does not have any added value

I wasn't clear enough in my comment, but you should ask yourself if you're rushing into making bold statements. Before stating there's no value, ask what's the added value.

That being said, the idea is to move the boring styling logic out of the way. Handle initial style configurations in XML, keep logic in code.

swabbass: yeah, there is an added value sure, sorry for rushing out for my statement :].

My point is that moving them to XML will group them in one place and separate styling from the view logic, but here we add a step that is parsing XML, which I don't prefer while having a better solution, extract the style to another function out of this file, which its responsibility is to create the styles, wdyt?

guyca: > moving them to XML will group them in one place and separate styling from the view logic

Yes, exactly. Move the boring layout/style/configuration code out of the way. Keep the interesting stuff in code. 

> we add a step that is parsing XML, which I don't prefer 

I don't understand - what's the downside of declaring a 20 LOC XML file and then inflating it? 🤔

swabbass:
It turns out that, changing the right/left toolbar's width won't show the overflow menu button, an issue with the old commits implementation that changed the right/left width and caused wide buttons (text) to be squeezed or overlapped with the title, especially when setting the title to be in the middle of the screen horizontally.

## Take 1:

Fit Left and Right Toolbars around the growing title:
- Measure title, left, and right.
- Change the size of the left and right dynamically.
- Issue: overflow button shows only if the `menu` has an item count that meets the inner criteria that also checks the space available, hence, since no way to show the overflow menu button without changing menu items `showAsAction` or any manipulation to the menu, the left/right bar width change caused in-accessible action items in the best case and iterating over the possibilities that max-es the space for the title, as the worst case.

## Take 2:
- Upgrade `material` and `appcompat` libraries
- Use `MaterialToolbar` which now supports the title in the centre, in `1.4.0-alpha-02`.
- Issue: There literally no way to add view (in a normal way) in the left side of the inner title, that is protected to not being modified out of the library, the best case we drop left actions (which is impossible, at least for now) and life is easy just use one toolbar and that's it, no need to measure even, and in the worst case, extending it and re-layout the whole toolbar, like separating the title to remain at the centre and no overlap with the left bar.
- From the last point, changing the title itself, without changing the toolbars works for the right toolbar. 
- In this implementation the material library leaving the toolbars as is and change title width and location.

Screen Shot 2021-04-27 at 0 43 40


## Take 3:

Fit title between Left and Right toolbars:
- Inspired from this google code [issue](https://github.com/material-components/material-components-android/issues/2011).
- 3 parts container, Left, Title/Subtitle, Right.
- Left and Right `Toolbars` measured at most.
- They can take up the space they need.
- Calculate title width and its layout coordinates.
- Measure title `Exactly` stay at the centre but change the width.
- Calculate overlap from each side of the title and layout it accordingly.

https://user-images.githubusercontent.com/7227793/116177261-2c5ac080-a71c-11eb-849f-0f8221e7ca50.mov




CC: @guyca

lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/TitleSubTitleLayout.kt
swabbass: hmm, the downside is really not that big difference, but, that small difference is: instead of inflating -> setting attributes, using custom view styled for those attributes saves us from adding a Lil step (parsing XML) which is in this case, 3 attributes only, that makes XML parsing not that necessary, when actually we can set those attributes directly.

Besides, I tried the xml to set those attributes and it seems that `TextViewCompat.setTextAppearance(this, R.style.TitleBarTitle)` does not affect `singleLine` or `maxLines` attribute, see:

Screen Shot 2021-04-27 at 6 13 30

lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/TitleComponentMeasurer.kt
```diff
@@ -0,0 +1,32 @@
+package com.reactnativenavigation.views.stack.topbar.titlebar
+
+import kotlin.math.max
+
+fun resolveTitleBoundsLimit(
+        defaultStartMargin: Int,
```
guyca, line 6: Seems like we're always calling this function with `defaultMargin` which is a **private** constant - no need to pass it as a param to the function.

You can pull it out of the class scope and make the conversion without the use of a context like it's done [here](https://github.com/wix/react-native-navigation/blob/master/lib/android/app/src/main/java/com/reactnativenavigation/options/params/DensityPixel.java#L10)

lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/TitleComponentMeasurer.kt
```diff
@@ -0,0 +1,32 @@
...
+        parentWidth: Int,
+        titleWidth: Int,
+        leftBarWidth: Int,
+        rightBarWidth: Int,
+        isCenter: Boolean
+): Pair<Int, Int> {
```
guyca, line 12: Perhaps consider changing the return type to:

```kotlin
typealias RightLimit = Int
typealias StartX = Int

Pair<StartX, RightLimit>
```

To make the code more readable.

guyca (changes requested):

guyca:
@swabbass I'm not sure I understood your last message - but the video looks much better compared to the previous attempts!! Good job! 👏 

1. When there's no enough room for the left buttons, notice how the overflow button appears as the last button (index 2). It should appear on the left, at index 0.
2. Have you confirmed that `showAsAction` works as expected?

lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/stack/topbar/TopBarController.kt
```diff
@@ -15,30 +15,29 @@ import com.reactnativenavigation.viewcontrollers.stack.topbar.button.ButtonContr
...
-import com.reactnativenavigation.views.stack.topbar.titlebar.LeftButtonsBar
-import com.reactnativenavigation.views.stack.topbar.titlebar.RightButtonsBar
+import com.reactnativenavigation.views.stack.topbar.titlebar.RNNToolbar
 
 
 open class TopBarController(private val animator: TopBarAnimator = TopBarAnimator()) {
     lateinit var view: TopBar
-    private lateinit var leftButtonsBar: LeftButtonsBar
-    private lateinit var rightButtonsBar: RightButtonsBar
+    private lateinit var leftToolbar: RNNToolbar
```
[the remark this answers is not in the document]
swabbass: 100%

lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/TopBar.java
```diff
@@ -54,22 +53,22 @@
...
-        return mainToolBar.getLeftButtonsBar().getNavigationIcon();
+        return titleAndButtonsContainer.getLeftToolbar().getNavigationIcon();
     }
 
     public TopBar(final Context context) {
         super(context);
         context.setTheme(R.style.TopBar);
         setId(CompatUtils.generateViewId());
-        this.mainToolBar = new MainToolBar(getContext());
+        this.titleAndButtonsContainer = new TitleAndButtonsContainer(getContext());
```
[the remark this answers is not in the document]
swabbass: yeah, simplicity :]

lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/TitleAndButtonsContainer.kt
```diff
@@ -14,64 +12,53 @@ import com.reactnativenavigation.options.parsers.TypefaceLoader
...
 import com.reactnativenavigation.utils.ViewUtils
-import com.reactnativenavigation.utils.logd
 
 const val DEFAULT_LEFT_MARGIN = 16
 
-class MainToolBar(context: Context) : RelativeLayout(context) {
+
+class TitleAndButtonsContainer(context: Context) : RelativeLayout(context) {
+
+    private val defaultMargin = UiUtils.dpToPx(context, DEFAULT_LEFT_MARGIN)
```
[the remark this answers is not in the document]
swabbass: we do have this in the buttons toolbars,
Screen Shot 2021-04-27 at 20 01 00

 `defaultMargin` is used to add padding at the left side of the title component, now we have 3 components, I have to do it manually because the title is not inside a `Toolbar`

lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/TitleAndButtonsContainer.kt
```diff
@@ -14,64 +12,53 @@ import com.reactnativenavigation.options.parsers.TypefaceLoader
...
+        set(value) {
+            if (field != value) {
+                field = value
+                requestLayout()
+            }
+        }
 
     private val titleSubTitleBar = TitleSubTitleLayout(context).apply {
-        id = CompatUtils.generateViewId()
+        this.id = CompatUtils.generateViewId()
```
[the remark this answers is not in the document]
swabbass: not really!!

lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/TitleComponentMeasurer.kt
swabbass: Awesome!

swabbass:
> @swabbass I'm not sure I understood your last message - but the video looks much better compared to the previous attempts!! Good job! 👏
> 
> 1. When there's no enough room for the left buttons, notice how the overflow button appears as the last button (index 2). It should appear on the left, at index 0.
> 2. Have you confirmed that `showAsAction` works as expected?

Thx dude, your feedback helped a lot :].

I'm adding tests for that, maybe an E2E will cover `showAsAction` but it is in the working.

lib/android/app/src/test/java/com/reactnativenavigation/viewcontrollers/stack/StackPresenterTest.kt
```diff
@@ -408,50 +397,13 @@ class StackPresenterTest : BaseTest() {
...
         uut.mergeOptions(toMerge, parent, child)
-        val title = (topBar.mainToolBar.getTitleComponent() as TitleSubTitleLayout).getTitleTxtView()
+        val title = (topBar.titleAndButtonsContainer.getTitleComponent() as TitleSubTitleLayout).getTitleTxtView()
         assertThat(title).isNotNull()
         assertThat(title.typeface).isEqualTo(SOME_TYPEFACE)
         verify(topBar).setTitleFontSize(9.0)
         verify(topBar).setTitleTextColor(Color.RED)
     }
 
-    @Test
```
swabbass, line 418: this is moved to the container test case

swabbass:
@guyca  we are really close, there are a couple of already known bugs, about the alignment option not being called properly and other minor stuff, for now, I'm adding tests for each and every case possible.

swabbass:
@guyca we learned new stuff here, 
now, the title bar works like a charm, except for one case: RTL & react-component & centre, which leads me to what I learned about `ReactRootView`, in its measure whenever it has changes width, height, or measure spec it will trigger root layout to re-layout and re-measure, which calls `ReactRootView.onLayout/onMeasure` during global layout iteration that will trigger the same global re-layout effect, because in order to make accurate and efficient measurements like native Android ViewGroups (`TitleSubtitleBarLayout`) which we can call `component.measure` first with `AT_MOST` to figure out how much it will take, and then re-measure with `EXACTLY` with the modified width according to the toolbars limitations, hence, calling `ReactRootView.onMeasure` twice will trigger an infinite loop of measuring and layout-ing all the hierarchy, which will hang the main thread.

 `ReactRootView.onMeasure`:
 
Screen Shot 2021-05-04 at 7 10 31

So what was being done here is measuring once:
- `ViewGroup` instead of unnecessary `RelativeLayout` since its measuring twice every child (similar to what my calculations were but measuring the title twice only)
- Calling measure according to an initial limit, with `AT_MOST` to keep a check on the available space between the bars to prevent overlapping.


which worked as expected, except for the RTL case, the react-component get manipulated by the `UIManagerModule` to apply RTL stuff which in turn changes the layout, in a way that in the native side with many overriding tries to prevent that so I can layout where it should be, simply didn't work, see `ReactRootView.onLayout`:
Screen Shot 2021-05-04 at 7 25 09.

This issue explains it:
https://github.com/facebook/react-native/issues/19403

solution to look for (WIP):
https://github.com/facebook/react-native/issues/17968#issuecomment-366184161

guyca:
@swabbass I'm sorry but I didn't read the [last message](https://github.com/wix/react-native-navigation/pull/7096#issuecomment-831676343) - it's too long. Do you need anything from me?

swabbass:
> @swabbass I'm sorry but I didn't read the [last message](https://github.com/wix/react-native-navigation/pull/7096#issuecomment-831676343) - it's too long. Do you need anything from me?

its ok dude :]. I don't need anything, just wanted to share that with you :]

guyca:
@swabbass Amazing findings dude, looking forward to reviewing the final PR! 💪

danilobuerger:
Screenshot_20210505-134153
Sadly still not centered in my case.

swabbass:
> Screenshot_20210505-134153
> Sadly still not centered in my case. Using a string as title.

Thx dude, can you share that string here in order to reproduce it?

danilobuerger:
Sure, I am setting it like so after the component is mounted (as the data is not known at mount):
```
  private setTitle = () => {
      topBar: {
        subtitle: {
          text: undefined,
        },
        title: {
          text: "Birnen Mandel Grießbrei",
        },
      },
    });
  };
```

swabbass:
> Birnen Mandel Grießbrei

I tried to mimic your described behaviour and could not reproduce it, 

https://user-images.githubusercontent.com/7227793/117166277-e44c3580-adce-11eb-9cd6-612a241a2079.mov

I need more info, about what is in the toolbar? RTL or not?

danilobuerger:
@swabbass no worries, I will repro it in the playground and post the diff.

danilobuerger:
@swabbass sorry about the confusion, apparently it didn't apply all the patches the first time. Now it seems to work on my local build. However, if you guys could do a snapshot release, I could also put it through CI :).

guyca (changes requested):

lib/android/app/src/test/java/com/reactnativenavigation/views/TitleAndButtonsContainerTest.kt
```diff
@@ -0,0 +1,471 @@
...
+        activity.window.decorView.layoutDirection = direction
+        activity.setContentView(FrameLayout(activity).apply {
+            layoutDirection = direction
+            addView(uut, ViewGroup.LayoutParams(UUT_WIDTH, UUT_HEIGHT))
+        })
+        idleMainLooper()
+    }
+
+    @Test
+    fun `should maintain stable ids for guest component`() {
```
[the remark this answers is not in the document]
swabbass: it should be setComponent - should not change component id

lib/android/app/src/test/java/com/reactnativenavigation/views/TitleAndButtonsContainerTest.kt
```diff
@@ -0,0 +1,471 @@
...
+import com.reactnativenavigation.options.params.Colour
+import com.reactnativenavigation.options.params.NullColor
+import com.reactnativenavigation.views.stack.topbar.titlebar.ButtonBar
+import com.reactnativenavigation.views.stack.topbar.titlebar.DEFAULT_LEFT_MARGIN_PX
+import com.reactnativenavigation.views.stack.topbar.titlebar.TitleAndButtonsContainer
+import com.reactnativenavigation.views.stack.topbar.titlebar.TitleSubTitleLayout
+import org.assertj.core.api.Assertions
+import org.assertj.core.api.AssertionsForInterfaceTypes.assertThat
+import org.junit.Test
+import org.mockito.Mockito
```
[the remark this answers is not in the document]
swabbass: hmm im using it in other tests where i need mock `object` types in kotlin, here it was auto imported by the IDE, i can switch to that syntax maybe for next tests.

lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/TitleAndButtonsContainer.kt
```diff
@@ -0,0 +1,221 @@
...
+                lp.topMargin = pixelTopMargin
+                this@TitleAndButtonsContainer.layoutParams = lp
+            }
+        }
+    }
+
+    fun clear() {
+        if (this.childCount > 0 && this.component == null) {
+            this.titleSubTitleBar.visibility = View.INVISIBLE
+        }
```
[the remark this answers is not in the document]
swabbass: clear should clear the title now it is more clear to read :]

lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/TitleAndButtonsContainer.kt
```diff
@@ -0,0 +1,221 @@
...
+            this@TitleAndButtonsContainer.layoutParams = lp
+        } else {
+            this.layoutParams = MarginLayoutParams(LayoutParams.WRAP_CONTENT, pixelHeight)
+        }
+    }
+
+    fun setTopMargin(topMargin: Int) {
+        val pixelTopMargin = UiUtils.dpToPx(context, topMargin)
+        if (layoutParams != null) {
+            if (layoutParams is MarginLayoutParams) {
```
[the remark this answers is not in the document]
swabbass: that code was legacy, hmm i did not see that but thanks a lot I'll refactor it to fit yes it needs a refresh according to the new ViewGroup.

lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/TitleAndButtonsContainer.kt
```diff
@@ -0,0 +1,221 @@
...
+
+    fun setSubtitleColor(@ColorInt color: Int) = this.titleSubTitleBar.setSubtitleTextColor(color)
+
+    fun setTitleColor(@ColorInt color: Int) = this.titleSubTitleBar.setTitleTextColor(color)
+
+    fun getTitle(): String = this.titleSubTitleBar.getTitle()
+
+    fun setHeight(height: Int) {
+        val pixelHeight = UiUtils.dpToPx(context, height)
+        if (this.layoutParams != null) {
```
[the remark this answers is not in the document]
swabbass: legacy code, I just transformed an old java class to kotlin and rename it :] , but it should be changed

lib/android/app/src/test/java/com/reactnativenavigation/views/TitleAndButtonsContainerTest.kt
```diff
@@ -0,0 +1,471 @@
...
+    private lateinit var mockLeftBar: ButtonBar;
+    private lateinit var mockRightBar: ButtonBar;
+    private lateinit var mockComponent: View;
+    override fun beforeEach() {
+        super.beforeEach()
+        setup()
+    }
+
+    private fun setup(
+            mockUUT: Boolean = true,
```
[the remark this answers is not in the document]
swabbass: That need to be checked, I'm planning to work on upgrading the library dependencies for v8, so in another task soon I think, the Junit or some other lib will be checked and integrated.

lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/TitleAndButtonsContainer.kt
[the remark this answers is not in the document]
guyca: @swabbass I don't understand your response. Changing visibility in `clear()` is a smell that should be addressed. Changing visibility in a function called clear is an unexpected side-effect.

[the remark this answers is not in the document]
guyca: > clear should clear the title now it is more clear to read :]

@swabbass Perhaps your response is missing a comma? Or maybe even a period?

Perhaps something like this: "Clear should clear the title. Now, it is more clear to read :]"? I think better usage of punctuation marks will make your responses more readable and convey your ideas more clearly.

lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/TopBar.java
```diff
@@ -123,12 +123,28 @@ public void setHeight(int height) {
...
 
     public void setTitleTopMargin(int topMargin) {
-        titleAndButtonsContainer.setTopMargin(topMargin);
+        int marginPx = UiUtils.dpToPx(getContext(), topMargin);
+        ViewGroup.LayoutParams layoutParams = titleAndButtonsContainer.getLayoutParams();
+        MarginLayoutParams marginLp;
+        if (layoutParams instanceof MarginLayoutParams) {
+            marginLp  = (MarginLayoutParams) layoutParams;
+        }else{
+            marginLp = new MarginLayoutParams(layoutParams);
```
guyca, line 142: Why are we creating LayoutParams in a function that sets margin? If `setTitleTopMargin()` is called and there are no layout params we should crash. It's an error, a case that should not happen and should not be supported.

@swabbass Please explain the use case where layout params can be null at this point.
It seems like the `titleAndButtonsContainer` is [created in the constructor](https://github.com/wix/react-native-navigation/blob/5d551cb34ea9d9f607cda888019f3728217d549c/lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/TopBar.java#L71), and immediately added to the hierarchy [here](https://github.com/wix/react-native-navigation/blob/5d551cb34ea9d9f607cda888019f3728217d549c/lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/TopBar.java#L86).

What am I missing? 🤔

lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/TopBar.java
```diff
@@ -123,12 +123,28 @@ public void setHeight(int height) {
         setLayoutParams(lp);
     }
 
+
```
guyca, line 126:
```suggestion
```

Delete accidental double empty line

lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/TopBar.java
```diff
@@ -123,12 +123,28 @@ public void setHeight(int height) {
...
+        ViewGroup.LayoutParams layoutParams = titleAndButtonsContainer.getLayoutParams();
+        MarginLayoutParams marginLp;
+        if (layoutParams instanceof MarginLayoutParams) {
+            marginLp  = (MarginLayoutParams) layoutParams;
+        }else{
+            marginLp = new MarginLayoutParams(layoutParams);
+        }
+        if (marginLp.topMargin != topMargin) {
+            marginLp.topMargin = marginPx;
+            this.titleAndButtonsContainer.setLayoutParams(marginLp);
```
guyca, line 146:
```suggestion
            titleAndButtonsContainer.setLayoutParams(marginLp);
```

guyca (changes requested):

lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/TopBar.java
swabbass: i think I got what you are talking about, essentially no way that LP for `titleAndButtonsContainer` can be null, since it is being set in its parent c'tor, this is not null check, I'm checking for at least `MarginLayoutParams` as the LP for this view, the other options here for LP is to be `ViewGroup.LayoutParams` (no margin options).


I can make it easier and set `MarginLayoutParams` at c'tor. This will make those checks unnecessary since the across the life span of  `titleAndButtonsContainer`, the LP won't change its type.

lib/android/app/src/test/java/com/reactnativenavigation/utils/TitleAndButtonsMeasurer.kt
```diff
@@ -0,0 +1,357 @@
...
+    @Test
+    fun `No Buttons - Aligned start - Title should be at default left margin bar width and right margin`() {
+        val barWidth = 200
+        val leftButtons = 0
+        val rightButtons = 0
+        val isRTL = false
+        val center = false
+        val (left, right) = resolveTitleBoundsLimit(parentWidth, barWidth, leftButtons, rightButtons, center, isRTL)
+
+        assertEquals(DEFAULT_LEFT_MARGIN_PX, left)
```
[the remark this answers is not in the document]
swabbass: wdym ? replace `DEFAULT_LEFT_MARGIN_PX` with `getContentInsetStart`?

lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/TitleAndButtonsContainer.kt
```diff
@@ -0,0 +1,221 @@
...
+    }
+
+    override fun onLayout(changed: Boolean, l: Int, t: Int, r: Int, b: Int) {
+        layoutChildren(l, r, t, b)
+    }
+
+    private fun layoutChildren(l: Int, r: Int, t: Int, b: Int) {
+        val titleComponent = getTitleComponent()
+        val isCenter = titleComponentAlignment == Alignment.Center
+        val parentWidth = r - l
```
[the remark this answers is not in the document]
swabbass: yes, the arguments `l: Int, r: Int, t: Int, b: Int` are related to the container, the parent of the container set the layout bounds for it.
The container here is the parent for the views that we are going to layout, hence its bounds represents its size.

lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/TitleAndButtonsContainer.kt
```diff
@@ -0,0 +1,221 @@
...
+                isCenter,
+                isRTL
+        )
+        val (leftButtonsLeft, leftButtonsRight) = resolveLeftToolbarBounds(parentWidth, leftButtonBar.measuredWidth, isRTL)
+        val (rightButtonsLeft, rightButtonsRight) = resolveRightToolbarBounds(parentWidth, rightButtonBar.measuredWidth,
+                isRTL)
+
+        leftButtonBar.layout(leftButtonsLeft, t, leftButtonsRight, b)
+        rightButtonBar.layout(rightButtonsLeft, t, rightButtonsRight, b)
+        truncateComponentMeasurementIfNeeded(titleRight, titleLeft, titleComponent, isCenter)
```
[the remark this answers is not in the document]
swabbass: not it's not the case, I'm re-measuring the view in order to make inner TextView Truncate the text according to the new measurement, so it does not mean overlap, it should be `measureTextTitleEnsureTruncatedEndIfNeeded`

lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/TitleAndButtonsContainer.kt
```diff
@@ -0,0 +1,221 @@
...
+            val margin = if (isCenter) 0 else 2 * DEFAULT_LEFT_MARGIN_PX
+            titleComponent.measure(MeasureSpec.makeMeasureSpec(titleRight - titleLeft + margin, MeasureSpec.EXACTLY),
+                    MeasureSpec.makeMeasureSpec(titleComponent.measuredHeight, MeasureSpec.EXACTLY))
+        }
+    }
+
+    override fun onMeasure(widthMeasureSpec: Int, heightMeasureSpec: Int) {
+        super.onMeasure(widthMeasureSpec, heightMeasureSpec)
+        val mode = MeasureSpec.getMode(widthMeasureSpec)
+        val parentWidth = MeasureSpec.getSize(widthMeasureSpec)
```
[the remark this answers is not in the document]
swabbass: measuredWidth will do :] thx

lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/TitleAndButtonsContainer.kt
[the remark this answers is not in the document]
swabbass: i made a rename where its more clear

lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/TopBar.java
guyca: @swabbass `titleAndButtonsContainer()` is added to a LinearLayout and `LinearLayout.LayoutParams extends ViewGroup.MarginLayoutParams`.

You should assume `titleAndButtonsContainer.layoutParams` always returns MarginLayoutParams. Any check you add around this logic makes the code confusing and misleading.

lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/TitleAndButtonsContainer.kt
```diff
@@ -0,0 +1,181 @@
...
+
+    private fun measureTextTitleEnsureTruncatedEndIfNeeded(titleRight: TitleRight, titleLeft: TitleLeft, titleComponent: View, isCenter: Boolean) {
+        if (component == null && titleRight - titleLeft != titleComponent.measuredWidth) {
+            val margin = if (isCenter) 0 else 2 * DEFAULT_LEFT_MARGIN_PX
+            titleComponent.measure(MeasureSpec.makeMeasureSpec(titleRight - titleLeft + margin, MeasureSpec.EXACTLY),
+                    MeasureSpec.makeMeasureSpec(titleComponent.measuredHeight, MeasureSpec.EXACTLY))
+        }
+    }
+
+    override fun onMeasure(widthMeasureSpec: Int, heightMeasureSpec: Int) {
```
guyca, line 126: Extracting all of the measurements and layout concerns to dedicated classes would have allowed you to write cleaner and shorter tests.

lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/TitleAndButtonsContainer.kt
```diff
@@ -0,0 +1,181 @@
...
+        this.removeView(this.titleSubTitleBar)
+        this.titleSubTitleBar = layout
+        this.addView(this.titleSubTitleBar, LayoutParams(LayoutParams.WRAP_CONTENT, LayoutParams.MATCH_PARENT))
+    }
+
+    @RestrictTo(RestrictTo.Scope.TESTS)
+    fun setButtonBars(left: ButtonBar, right: ButtonBar) {
+        val leftLp = LayoutParams(leftButtonBar.layoutParams)
+        val rightLp = LayoutParams(rightButtonBar.layoutParams)
+        this.removeView(leftButtonBar)
```
guyca, line 171: What's up with the accessive use of `this`?

guyca (changes requested):

lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/TitleAndButtonsContainer.kt
```diff
@@ -0,0 +1,181 @@
...
+        }
+
+    private var titleSubTitleBar = TitleSubTitleLayout(context)
+    var leftButtonBar = ButtonBar(context)
+        private set
+    var rightButtonBar = ButtonBar(context)
+        private set
+
+    init {
+        this.addView(leftButtonBar, LayoutParams(LayoutParams.WRAP_CONTENT, LayoutParams.MATCH_PARENT))
```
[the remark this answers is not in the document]
swabbass: it is a bad habit of mine, thanks for the note :].

lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/TitleAndButtonsContainer.kt
```diff
@@ -0,0 +1,181 @@
...
+        this.titleSubTitleBar.clear()
+        clearComponent()
+    }
+
+    override fun onLayout(changed: Boolean, l: Int, t: Int, r: Int, b: Int) {
+        val titleComponent = getTitleComponent()
+        val isCenter = titleComponentAlignment == Alignment.Center
+        val containerWidth = r - l
+        val containerHeight = b - t
+        val isRTL = isRTL()
```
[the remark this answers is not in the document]
swabbass: since I'm calling the same function in multiple places in `onLayout` its better to store the same value and make less function calls, still readable.

lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/TitleAndButtonsContainer.kt
swabbass: The tests are short mainly not that long, and the logic for measurements was extracted to a separate high order function where its tests are really short.

lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/TitleAndButtonsContainer.kt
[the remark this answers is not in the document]
guyca: All `isRtl()` does is compare a value. You can safely assume that calling it twice has absolutely zero performance implications.

Applications running RNN are installed on state of the art, 64bit devices with plenty of memory and multiple cores - they can handle comparing to integers.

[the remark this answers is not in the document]
swabbass: I see that this is clearer and consistent with the nearby values I use. 
Since we both know device's capabilities (thanks for your reminders), we can agree that my way and your way are readable, straightforward, and have the same result for each run.
So it is just a matter of preferred style.

lib/android/app/src/main/java/com/reactnativenavigation/utils/ViewExtensions.kt
```diff
@@ -3,7 +3,7 @@ package com.reactnativenavigation.utils
 import android.view.View
 import android.view.ViewManager
 
-fun View.removeFromParent() = (this.parent as? ViewManager)?.removeView(this)
+fun View?.removeFromParent() = (this?.parent as? ViewManager)?.removeView(this)
```
guyca, line 6: I don't understand this change. @swabbass can you please explain it to me? Wouldn't you rather have the null check on the actual receiver? Hiding it makes the code less explicit.

guyca (changes requested):

lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/TitleAndButtonsContainer.kt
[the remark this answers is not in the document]
guyca: I'm glad we agree that invoking `isRtl()` more than once has no performance implications. This leads me to something I noticed in your PRs - at times it seems like a lot of preparation work is done before calling some functions.
Just here, 8 lines are required to prepare various parameters for `resolveHorizontalTitleBoundsLimit()`.

Thankfully, Android Studio supports named arguments so we no longer need to declare parameters in advance for readability. For example, previously passing null and false was very confusing. We had to declare variables for the code to be readable. With named variables, this isn't needed anymore. The IDE provides named arguments when needed as visual aids.
Screen Shot 2021-05-12 at 10 47 06

I strongly feel you should inline these variables. Once you do, you'll notice you can get rid of some of the parameters and calculate them in the functions themselves.

* `isCenter` is inferred from `titleComponentAlignment` which is a public member of `TitleBarAndButtonsContainer`.
* isRtl is now a public extension function of View which in our case is `TitleBarAndButtonsContainer`
* `leftBarWidth` is inferred from `leftButtonBar` which is public in `TitleBarAndButtonsContainer`
* `rightBarWidth` is inferred from `rightButtonBar` which is public in `TitleBarAndButtonsContainer`
* `titleWidth` is inferred from `titleComponent` which is public in  `TitleBarAndButtonsContainer`
* `titleComponent` is the result of the internal function `getTitleComponent()` - it can be made public.

You can pass **`this`**, `r`, `l` as parameters to `resolveHorizontalTitleBoundsLimit`. In `resolveHorizontalTitleBoundsLimit` use the `with(titleAndButtonsContainer)` so you can easily access whatever public members/getters you need.

Instead of 6 parameters, we're down to 3 and we can safely avoid the 8 line preparation work. An additional benefit is that calculations (like `containerWidth` and `containerHeight `) are done closer to where they are used.

lib/android/app/src/main/java/com/reactnativenavigation/utils/ViewExtensions.kt
swabbass: This is mainly changed due to an error that happened while compiling tests that check null views.

Since `View?` can receive null or non-null values and in Kotlin it is treated as a different type,  we can run extensions declared for this type (`View?`)  only, and check the receiver for null and handling it internally, in our code the `?` operator behaves as a null check.

It seems that hiding it would make the code less explicit, but, since the declared view is nullable in java by default and in Kotlin we use `View?`,  it's understandable that accessing it will need a null check, that means it demands devs attention to check for null, what I did for one extension is to make the check internally since the types declared explicitly as nullable, makes the code cleaner with short single liners.

lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/TitleAndButtonsContainer.kt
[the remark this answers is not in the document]
swabbass: You have a good point here dude :] 

IDEs today are really helpful as you described, but GitHub does not have such a feature, now,  what these preps do is making measurement logic independent from any `View` (android specific code) and use just numbers, which means, writing simple and short tests for the calculations without mocking `View` and mock return for each param you listed above. 

That was the only concern about having them like this, so changing it now would take a bit of time re-writing tests and organizing mocks.

lib/android/app/src/main/java/com/reactnativenavigation/utils/ViewExtensions.kt
guyca: But with null coalescing, our code is already cleaner. Even the motivation here is wrong - this change was introduced due to compilation errors in tests. You shouldn't change "production code" because of tests.

Can you call `addView()` on a null ViewGroup? Of course not, it makes no sense. `(null as View?).removeFromParent()` also makes no sense. What's even worst, is that this change might prevent us from detecting actual bugs early on in the development cycle.

For example, if you're trying to remove a view from its parent, but for some reason it's null. Instead of crashing during runtime, your code will work.

swabbass: Production code did not change its the same thing, as it was in java and I moved it to Extension i forgot to add ? and that gave us the error.

Extensions and member functions are totally different in concept, `addView` cant be called on null, but extensions can!, [see](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/is-null-or-empty.html) `String.IsNullOrEmpty()` for nullable types we can use nullable syntax, in Java mindset it is a static function that receives the object with nullable annotation and its called this so we need to check nulls if the value can have null, otherwise it's safe.

About the example, if I'm trying to remove the view from its parent and the view is null then nothing will happen simply the same behaviour that was before, which was checking nulls before removing the view,  [see](https://github.com/wix/react-native-navigation/blob/b53f13c60fe6498ad1690adf1dfd7509f602965c/lib/android/app/src/main/java/com/reactnativenavigation/utils/ViewUtils.java#L118). 

What I'm trying to say here, it's a different syntax but the same thing happens in Java and Kotlin implementations.
Kotlin and Java are different mindset languages that turn out to be JVM languages.

guyca (approved):

PR #7116, opened 2021-05-13
Fix incorrect bounds of FastImage in shared element transition

guyca:
Seems like native images load drawables in two steps. Only in the second step, the image is scaled to fit its bounds. If we return the view before the image has been scaled properly, the shared element transition will use incorrect values when calculating dimensions. This will cause the image not to transition correctly into place.

This change is related to native ImageViews and does not affect ReactImageView.


mrousavy (approved):
works like magic and we didn't notice any regressions in our app. amazing work @guyca ❤️

guyca:
woohooo 🎉

PR #7119, opened 2021-05-13
Fix fast image border animation

hannojg:
### The issue

The `FastImageBorderRadiusAnimator#shouldAnimateProperty` returned false in cases where the `FastImage` component is a child of another component that has a border radius. Example for such a use case:

```jsx
<View style={{ borderRadius: 30, overflow: "hidden" }}>
   <FastImage {...} />
</View>
```

In this case the animation would look like this. You can see that the borders will "make a jump" once the SET is finished:

https://user-images.githubusercontent.com/16821682/118120763-aebbd380-b3f0-11eb-879f-1348b11721a0.mp4

### The fix

I wrote a function that checks recursively:
* If the view is a ReactView -> try to get border radius from this react view
* If the view is a FastImageView -> try to get border radius from this FastImage view
* If it founds a border radius it returns

https://user-images.githubusercontent.com/16821682/118122398-183ce180-b3f3-11eb-8df8-e0f0202a2a3f.mp4

### Open questions

The recursive function fails when it reaches the root with the following error
```
java.lang.ClassCastException: android.view.ViewRootImpl cannot be cast to android.view.ViewGroup
```
currently, I try/catch this. Maybe there is a better way to check this (like `view is Overlay`) ?

### DIFF
--- a/lib/android/app/src/main/java/com/reactnativenavigation/views/element/animators/FastImageBorderRadiusAnimator+utilities.kt
+++ b/lib/android/app/src/main/java/com/reactnativenavigation/views/element/animators/FastImageBorderRadiusAnimator+utilities.kt
@@ -4,8 +4,10 @@ import android.view.View
 import android.view.ViewGroup
 import com.facebook.react.views.view.ReactViewGroup
 import com.reactnativenavigation.R
+import com.reactnativenavigation.react.ReactView
 import com.reactnativenavigation.utils.ViewTags
 import com.reactnativenavigation.utils.borderRadius
+import com.reactnativenavigation.viewcontrollers.viewcontroller.overlay.OverlayLayout
 
 /**
  * Gets the border radius for a react-native-fast-image view that
--- a/lib/android/app/src/main/java/com/reactnativenavigation/views/element/animators/FastImageBorderRadiusAnimator+utilities.kt
+++ b/lib/android/app/src/main/java/com/reactnativenavigation/views/element/animators/FastImageBorderRadiusAnimator+utilities.kt
@@ -41,7 +41,7 @@ private fun getBorderRadius(v: View): Float {
 
 private fun getParent(view: View): ViewGroup? = try {
     when(view.parent) {
-        null -> null
+        null, is ReactView, is OverlayLayout -> null
         else -> ViewTags.get<ViewGroup>(view, R.id.original_parent, view.parent as ViewGroup)
     }
 // TODO: can we handle this better?: java.lang.ClassCastException: android.view.ViewRootImpl cannot be cast to android.view.ViewGroup
--- a/lib/android/app/src/main/java/com/reactnativenavigation/views/element/animators/FastImageBorderRadiusAnimator+utilities.kt
+++ b/lib/android/app/src/main/java/com/reactnativenavigation/views/element/animators/FastImageBorderRadiusAnimator+utilities.kt
@@ -1,0 +1,71 @@
+package com.reactnativenavigation.views.element.animators
+
+import android.view.View
+import android.view.ViewGroup
+import androidx.core.view.children
+import com.facebook.react.views.view.ReactViewBackgroundDrawable
+import com.facebook.react.views.view.ReactViewGroup
+import com.reactnativenavigation.R
+import com.reactnativenavigation.utils.ViewTags
+import com.reactnativenavigation.utils.borderRadius
+import com.reactnativenavigation.viewcontrollers.viewcontroller.overlay.OverlayLayout
+import java.lang.ClassCastException
+
+/**
+ * Gets the border radius for a react-native-fast-image view that
+ * is eventually embedded in other components. It might be the case
+ * that the FastImage doesn't have a border radius, in this case
+ * a inherited border radius will be looked up. Example of such a case
+ *
+ * @example
+ *
+ * <View style={{ borderRadius: 30, overflow: "hidden" }}>
+ *     <FastImage {...} />
+ * </View>
+ */
+fun FastImageBorderRadiusAnimator.getInheritedBorderRadius(v: View): Float {
+    val borderRadius = getBorderRadius(v)
+    if (borderRadius > 0f) {
+        return borderRadius
+    }
+
+    return when(val parentView = getParent(v)) {
+        null -> 0f
+        else -> getInheritedBorderRadius(parentView)
+    }
+}
+
+private fun getBorderRadius(v: View): Float {
+    // gets the border radius from a react view
+    if (v is ReactViewGroup && v.borderRadius > 0f) {
+        return v.borderRadius
+    }
+
+    // gets the border radius for a FastImage view
+    return getBorderRadiusFastImage(v);
+}
+
+private fun getBorderRadiusFastImage(v: View): Float {
+    val parentView = getParent(v)
+
+    if (parentView == null || v is OverlayLayout) {
+        return 0f
+    }
+
+    val parentIsUsedOnlyToDrawBorderRadiusOverImage = parentView.childCount <= 1 || parentView.children.contains(v)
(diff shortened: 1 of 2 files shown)


hannojg:
Hmm, I think we don't need any of this anymore:

https://github.com/wix/react-native-navigation/blob/06eaa5f0620b7f3b512473c888962e123c38a2e4/lib/android/app/src/main/java/com/reactnativenavigation/views/element/animators/FastImageBorderRadiusAnimator%2Butilities.kt#L45-L61

This check earlier is enough, it will also gets the View's border radius that wraps the RNFastImage view. Not sure if this code has been build for any specific reason like that @guyca ?

lib/android/app/src/main/java/com/reactnativenavigation/views/element/animators/FastImageBorderRadiusAnimator+utilities.kt
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * <View style={{ borderRadius: 30, overflow: "hidden" }}>
+ *     <FastImage {...} />
+ * </View>
+ */
+fun FastImageBorderRadiusAnimator.getInheritedBorderRadius(v: View): Float {
+    val borderRadius = getBorderRadius(v)
+    if (borderRadius > 0f) {
+        return borderRadius
+    }
```
guyca, line 26:
```suggestion
    if (borderRadius > 0f) return borderRadius
```

lib/android/app/src/main/java/com/reactnativenavigation/views/element/animators/FastImageBorderRadiusAnimator+utilities.kt
```diff
@@ -0,0 +1,71 @@
...
+
+    return when(val parentView = getParent(v)) {
+        null -> 0f
+        else -> getInheritedBorderRadius(parentView)
+    }
+}
+
+private fun getBorderRadius(v: View): Float {
+    // gets the border radius from a react view
```
guyca:
```suggestion
```

lib/android/app/src/main/java/com/reactnativenavigation/views/element/animators/FastImageBorderRadiusAnimator+utilities.kt
```diff
@@ -0,0 +1,71 @@
...
+    if (v is ReactViewGroup && v.borderRadius > 0f) {
+        return v.borderRadius
+    }
+
+    // gets the border radius for a FastImage view
```
guyca:
```suggestion
```

lib/android/app/src/main/java/com/reactnativenavigation/views/element/animators/FastImageBorderRadiusAnimator+utilities.kt
```diff
@@ -0,0 +1,71 @@
...
+    return getBorderRadiusFastImage(v);
+}
+
+private fun getBorderRadiusFastImage(v: View): Float {
+    val parentView = getParent(v)
+
+    if (parentView == null || v is OverlayLayout) {
+        return 0f
+    }
```
guyca:
```suggestion
    if (parentView == null || v is OverlayLayout) return 0f
```

lib/android/app/src/main/java/com/reactnativenavigation/views/element/animators/FastImageBorderRadiusAnimator+utilities.kt
```diff
@@ -0,0 +1,71 @@
...
+
+    val parentIsUsedOnlyToDrawBorderRadiusOverImage = parentView.childCount <= 1 || parentView.children.contains(v)
```
guyca: @hannojg Isn't `parentView.children.contains(v)` always true? 🤔

lib/android/app/src/main/java/com/reactnativenavigation/views/element/animators/FastImageBorderRadiusAnimator+utilities.kt
```diff
@@ -0,0 +1,52 @@
...
+
+    return 0f;
+}
+
+private fun getParent(view: View): ViewGroup? = try {
+    when(view.parent) {
+        null -> null
+        else -> ViewTags.get<ViewGroup>(view, R.id.original_parent, view.parent as ViewGroup)
+    }
+// TODO: can we handle this better?: java.lang.ClassCastException: android.view.ViewRootImpl cannot be cast to android.view.ViewGroup
```
guyca, line 49: @hannojg Yes. Basically, we should return from the recursion earlier and not traverse up the hierarchy all the way to system views. I'll add the relevant changes as suggestions.

lib/android/app/src/main/java/com/reactnativenavigation/views/element/animators/FastImageBorderRadiusAnimator+utilities.kt
```diff
@@ -0,0 +1,52 @@
...
+    if (v is ReactViewGroup && v.borderRadius > 0f && v.overflow == "hidden") {
+        return v.borderRadius
+    }
+
+    return 0f;
+}
+
+private fun getParent(view: View): ViewGroup? = try {
+    when(view.parent) {
+        null -> null
```
guyca, line 46:
```suggestion
        null, is ReactView, is OverlayLayout -> null
```

This should prevent us from getting a ClassCastException. You can safely remove the try-catch.

lib/android/app/src/main/java/com/reactnativenavigation/views/element/animators/FastImageBorderRadiusAnimator.kt
```diff
@@ -11,23 +11,31 @@ import androidx.core.animation.doOnEnd
...
-            Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP &&
-            fromChild !is ReactImageView
-                    && toChild !is ReactImageView
-                    && (getBorderRadius(from) != 0f || getBorderRadius(to) != 0f)
+    override fun shouldAnimateProperty(fromChild: ImageView, toChild: ImageView): Boolean {
+        return Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP &&
+                fromChild !is ReactImageView
+                && toChild !is ReactImageView
+                && (getInheritedBorderRadius(from) != 0f
+                || getInheritedBorderRadius(to) != 0f)
```
guyca, line 31:
```suggestion
                && (getInheritedBorderRadius(from) != 0f || getInheritedBorderRadius(to) != 0f)
```

guyca (changes requested):

lib/android/app/src/main/java/com/reactnativenavigation/views/element/animators/FastImageBorderRadiusAnimator+utilities.kt
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * <View style={{ borderRadius: 30, overflow: "hidden" }}>
+ *     <FastImage {...} />
+ * </View>
+ */
+fun FastImageBorderRadiusAnimator.getInheritedBorderRadius(v: View): Float {
+    val borderRadius = getBorderRadius(v)
+    if (borderRadius > 0f) return borderRadius
+
+    if (v is ReactView || v is OverlayLayout) {
```
guyca, line 28: This check should be done first thing in the method as both ReactView and OverlayLayout are RNN internal layouts. Even if they had a border-radius, we don't want to use it in our calculations.

guyca (changes requested):

guyca (approved):

guyca:
The FastImageBorderRadiusAnimator#shouldAnimateProperty returned false in cases where the FastImage component is a child of another component that has a border radius.

PR #7125, opened 2021-05-19
Compile jsx to plain js

alpha0010:
When configured as `react-native`, jsx is not converted to function calls https://www.typescriptlang.org/tsconfig#jsx .  This means other tools (e.g. `jest`) must themselves be configured to transpile, instead of being able to consume the js directly.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -5,7 +5,7 @@
     "allowJs": false,
     "target": "esnext",
     "module": "commonjs",
-    "jsx": "react-native",
+    "jsx": "react",
     "declaration": true,
     "esModuleInterop": true,
     "skipLibCheck": true,


alpha0010:
Actually, for the specific case of `jest`, this is probably not necessary.  (I think something is broken in my local configuration.)

danilobuerger:
Can't really confirm this. I am on 0.64 too. The `transformIgnorePatterns` changed in 0.64, maybe you are seeing issues with that?

mrousavy:
Why does the `react-native` option exist then?

alpha0010:
> Can't really confirm this. I am on 0.64 too. The `transformIgnorePatterns` changed in 0.64, maybe you are seeing issues with that?

Probably, or at least that is how I fixed it.  I used the pattern:
```
"node_modules/(?!((jest-)?react-native|@react-native(-community)?|react-native-(alert-async|in-app-review|navigation|url-polyfill))/)"
```
(Includes several other libraries that broke for me on upgrade.)

> Why does the `react-native` option exist then?

From what I can tell, it is to convert ts to js, and allow a later compile step to determine what technique to apply to compile the jsx (https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html).  Now does anyone actually make use of that?  I do not know...

yogevbd:
I encountered this bug while trying to run jest with our mocks on a different project, when using `"jsx": "react-native"`, this error appears: `Uncaught ReferenceError: React is not defined`.

PR #7138, opened 2021-05-28
fix: Only animate modals if they're not presented as `pageSheet` or `formSheet`

mrousavy:
Modals can have custom animations via `animations.showModal` or `animations.dismissModal`. This works fine for Modals presented in the following presentation style:

```
    overFullScreen
    overCurrentContext
    currentContext 
    popover 
    fullScreen
    none
```

but once you try animating modals with either `pageSheet` or `formSheet` presentation style, the animation goes crazy (goes beyond top, leaves blank see-thru area at the bottom)

This PR disables animation options for modals presented as page- or form-sheets to avoid this mistake. This means, for `pageSheet` and `formSheet` modals you cannot have custom content animations. Element transitions and Shared element transitions still work as expected.


lib/ios/RNNModalManager.m
```diff
@@ -33,6 +33,13 @@ - (instancetype)initWithBridge:(RCTBridge *)bridge
     return self;
 }
 
+- (BOOL)canAnimate:(UIViewController *)viewController {
+    // formSheet and pageSheet modals have a custom slide-from-bottom animation. Applying any custom
+    // animation here will break that default iOS animation.
+    return viewController.modalPresentationStyle != UIModalPresentationFormSheet &&
```
guyca, line 39: @yogevbd We should check `modalPresentationStyle` from resolvedOptions, right?

yogevbd: also, `modalPresentationStyle` is a Text object and `UIModalPresentationFormSheet` is an integer, best thing is to add a new option object type for `modalPresentationStyle`. I can take it

yogevbd: actually we apply `modalPresentationStyle` on the top parent and we also present/hide the top parent so it looks good to me

yogevbd (approved):

yogevbd (approved):

Issue #7152, opened 2021-06-09
popTo does not reject properly [fix provided]


lindesvard:
## 🐛 Bug Report

So `popTo` method does not resolve/reject when providing a componentId that does *not* exist in the stack

### Have you read the [Contributing Guidelines on issues](https://wix.github.io/react-native-navigation/docs/meta-contributing/)?

Yes

## To Reproduce

Run this code and notice that the promise never resolve
```
try {
     await Navigation.popTo('id_that_do_not_exist');
} catch (e) {
     console.log('Should print this and give an error', e)
}
console.log('Should log this as well')
```

## Expected behavior

Should print `Should print this and give an error` and then `Should log this as well`.

## Actual Behavior

Nothing gets printed

## Your Environment

- React Native Navigation version: 7.16.0
- React Native version: 16.13.1
- Platform(s) (iOS, Android, or both?): ios (**works for android**)
- Device info (Simulator/Device? OS version? Debug/Release?): iOS Simulator 14.3

## Reproducible Demo

https://github.com/lindesvard/rnn-bugs

Run it without the patch https://github.com/lindesvard/rnn-bugs/blob/master/patches/react-native-navigation%2B7.16.0.patch

yogevbd:
The fix looks good! could you please submit a PR?

lindesvard:
Sure, will try to get time today 👍

PR #7155, opened 2021-06-15
fix popTo so promise get resolved

lindesvard:
This should resolve the promise even if you add a none existing componentId to `popTo`.

More context: https://github.com/wix/react-native-navigation/issues/7152


yogevbd:
@lindesvard try running `yarn run clang-format -i lib/ios/**/*.(h|m|mm)`. it should reformat all iOS files.

lindesvard:
> yarn run clang-format -i lib/ios/**/*.(h|m|mm)

Done! @yogevbd

yogevbd (approved):

PR #7164, opened 2021-06-22
docs: Fix wrong API in example for custom button

atlatosBast:
This took me some time. For searchability, using a string instead of the object for component throws "Attempted to assign to readonly property.". Leaving out the id freezes the app (it's marked as mandatory in the button docs).


yogevbd:
@atlatosBast Awesome fix, could you please update all `stack-buttons.mdx` with this fix? thanks!

yogevbd (approved):

Issue #7148, opened 2021-06-07
Android: setStackRoot on a non-selected tab triggers componentDidAppear


tastyqbit:
## 🐛 Bug Report

When trying to setStackRoot on a tab that is not currently selected, the componentDidAppear method triggers anyway. Then the next time the unselected tab is selected, the componentDidAppear doesnt trigger

EDIT: The bug only happens when setStackRoot animations are disabled

## To Reproduce

Only on android

1. A tab based root navigation
2. setStackRoot on non selected tab screen with default setStackRoot animated disabled
3. componentDidAppear triggers on that tab
4. Clicking the unselected tab and showing it does not trigger componentDidAppear

## Expected behavior

componetnDidAppear should not trigger if the screen is not active

## Actual Behavior

componentDidAppear triggers

## Your Environment

- React Native Navigation version: 7.6.0
- React Native version: 0.63.4
- Platform(s) (iOS, Android, or both?): Android
- Device info (Simulator/Device? OS version? Debug/Release?): All

swabbass:
@tastyqbit Thx for the info :] we are looking into it

PR #7169, opened 2021-06-30
setStackRoot crash  & wrong ViewDidAppear event fix

swabbass:
# Issue:
- calling setStackRoot, on a stack that was not already created caused a crash while accessing its UI elements.
- DidAppear event sent when setStackRoot is called immediately while the view did not appear.

# Fix:
- Manipulate the stack during `setStackRoot`, `push`, `pop`, without changing UI, since there isn't any.
- Fix calls of did appear if it is already not appeared when setting stack root.

Closes: #7148


swabbass:
E2E fails over Android Emulator ANR

lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/stack/StackController.java
```diff
@@ -222,7 +227,8 @@ public void setRoot(@Size(min = 1) List<ViewController> children, CommandListene
         CommandListener listenerAdapter = new CommandListenerAdapter() {
             @Override
             public void onSuccess(String childId) {
-                child.onViewDidAppear();
+                if (child.isViewShown())
```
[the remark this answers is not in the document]
swabbass: it will be called once the view appeared, the tests for the lifecycle (unit and e2e) grants that

lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/stack/StackController.java
```diff
@@ -263,6 +262,18 @@ public void onSuccess(String childId) {
         }
     }
 
+    private void refillStack(List<ViewController> children) {
+        stack = new IdStack<>();
```
yogevbd, line 266:
```suggestion
```
look like this line isn't needed

lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/stack/StackController.java
```diff
@@ -263,6 +262,18 @@ public void onSuccess(String childId) {
         }
     }
 
+    private void refillStack(List<ViewController> children) {
```
yogevbd, line 265:
```suggestion
    private void setChildren(List<ViewController> children) {
```

lib/android/app/src/test/java/com/reactnativenavigation/viewcontrollers/navigator/NavigatorTest.java
```diff
@@ -359,18 +359,6 @@ public void onSuccess(String childId) {
         });
     }
 
-    @Test
```
yogevbd, line 362: Doesn't look like we need to delete this test

e2e/StackSetRoot.test.js
```diff
@@ -0,0 +1,19 @@
+import Utils from './Utils';
+import TestIDs from '../playground/src/testIDs';
+
+const { elementById } = Utils;
+
+describe.e2e('Stack SetRoot', () => {
```
yogevbd, line 6: should be moved to `Stack.test.js`

yogevbd (changes requested):

lib/android/app/src/test/java/com/reactnativenavigation/viewcontrollers/navigator/NavigatorTest.java
swabbass: this checks creation of stack to make sure that set-root works which i replaced in the stack test

lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/stack/StackController.java
swabbass: that needs to change in more places

yogevbd (approved):

PR #7174, opened 2021-07-04
Move remx from devDependencies

yogevbd:
Since we use `remx` as the store in Navigation mock, it is no longer a dev dependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,8 @@
     "lodash": "4.17.x",
     "prop-types": "15.x.x",
     "react-lifecycles-compat": "2.0.0",
-    "tslib": "1.9.3"
+    "tslib": "1.9.3",
+    "remx": "3.x.x"
   },
   "devDependencies": {
     "@babel/core": "7.10.3",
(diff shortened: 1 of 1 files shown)


michaelduminy:
@yogevbd What is the purpose of the mock?

Can this be a peer dependency instead? 

The issue is that this now leaks into all projects that use RNN :(

michaelduminy:
Also, if the mock is only used in tests then this definitely qualifies as a devDependency

yogevbd:
@michaelduminy It allows running render tests with RNN. https://github.com/wix/react-native-navigation/pull/7140

The mock isn't used only in RNN tests but in our consumers tests as well. Is there an issue with having remx in your dependencies other than [that issue](https://github.com/wix/react-native-navigation/issues/7187)?

lindesvard:
Me and @michaelduminy  work at the same company so I can answer this. 

We do not explicit have a problem but this adds a lot of extra work for us and bloating our package.json. 

Currently we have a big app with a lot of dependencies. We have some checks to see if we get duplicate version of any package, if we do we need to add that version to our `resolutions`. Just a glimpse how much we need to add.

image

Also we do not want to add extra dependencies to our app even tho this likely will not be added in the production build.

yogevbd:
I Will move it to peerDependencies and will add:
```
"peerDependenciesMeta": {
    "remx": {
      "optional": true
    }
  }
```
as it required only when using RNN mocks.

michaelduminy:
Thanks so much @yogevbd!

yogevbd:
Happy to help! https://github.com/wix/react-native-navigation/pull/7230

PR #7178, opened 2021-07-08
Theme color to support dark mode themes

swabbass:
## `ThemeColour`

A new way to support dark/light mode colouring.
Quick Demo:

https://user-images.githubusercontent.com/7227793/125622899-c09a6b7e-d455-443c-a33b-d86cfe9e7559.mov

- [X] POC.
- [X] Base cases (mergeOptions, defaultOptions, NoColor).
- [X] Optimize `OnConfigurationChange` on Android.
- [x] iOS, implementation.
- [X] Support Backwards Compatibility.
- [x] Docs


lib/src/interfaces/Options.ts
```diff
@@ -3,7 +3,7 @@ import { ImageRequireSource, ImageSourcePropType, Insets } from 'react-native';
 
 // TODO: Import ColorValue instead when upgrading @types/react-native to 0.63+
 // Only assign PlatformColor or DynamicColorIOS as a Color symbol!
-declare type Color = string | symbol;
+export declare type RNNColor = string | symbol | ModedColour;
```
yogevbd, line 6: I think `RNNColor` should be renamed to `Color` as it doesn't have much sense to add a prefix in js

lib/android/app/src/test/java/com/reactnativenavigation/views/TitleAndButtonsContainerTest.kt
```diff
@@ -12,7 +12,8 @@ import com.nhaarman.mockitokotlin2.verify
 import com.reactnativenavigation.BaseTest
 import com.reactnativenavigation.options.Alignment
 import com.reactnativenavigation.options.params.Colour
-import com.reactnativenavigation.options.params.NullColor
+import com.reactnativenavigation.options.params.ThemeColour
+import com.reactnativenavigation.options.params.RNNNullColor
```
yogevbd: `RNNNullColor` should be `NullThemeColour` right?

lib/android/app/src/test/java/com/reactnativenavigation/viewcontrollers/viewcontroller/ViewControllerTest.java
```diff
@@ -1,6 +1,8 @@
 package com.reactnativenavigation.viewcontrollers.viewcontroller;
```
yogevbd, line 1: this file contain only imports

lib/android/app/src/main/java/com/reactnativenavigation/options/params/ThemeColour.kt
```diff
@@ -0,0 +1,49 @@
...
+    fun canApplyValue() = selectedColor().canApplyValue()
+
+    companion object{
+        @JvmStatic
+        fun of(color:Int) = ThemeColour(Colour(color), Colour(color))
+        @JvmStatic
+        fun of(light:Int,dark:Int) = ThemeColour(Colour(light),Colour(dark))
+
+        @JvmStatic
+        fun parseThemeColour(context: Context, json: JSONObject?): ThemeColour {
```
yogevbd, line 40: makes more sense to rename it to `parse` to be consistent with our other parsers

yogevbd (changes requested):

lib/android/app/src/test/java/com/reactnativenavigation/views/TitleAndButtonsContainerTest.kt
swabbass: outdated

lib/src/commands/OptionsProcessor.ts
```diff
@@ -115,7 +116,20 @@ export class OptionsProcessor {
 
   private processColor(key: string, value: any, options: Record<string, any>) {
     if (isEqual(key, 'color') || endsWith(key, 'Color')) {
-      options[key] = value === null ? 'NoColor' : this.colorService.toNativeColor(value);
+      const newColorObj: Record<string, any> = { dark: 'NoColor', light: 'NoColor' };
+      if (!value) {
```
yogevbd, line 120:
```suggestion
      if (value === null) {
```

I think we should check here if `value === null` because if it's undefined it shouldn't have any effect in native. passing `NoColor` will reset the current color.

yogevbd (approved):

yogevbd (approved):

Andrew-Chen-Wang:
For those who don't understand how to use this but were following this PR and the release, take a look here: https://github.com/Hear-Ye/Hear-Ye-App/commit/00a7b6a0806855e82672022b4db3b61f95435419

PR #7186, opened 2021-07-15
Fix statusBar.blur option on iOS

### LINKED ISSUE #7185: StatusBar is not blurring on iOS
## 🐛 Bug Report

The status bar is covering and conflicting with the content of the screen. We need to blur the status bar on iOS so it appears above the content of the screen.

Yes

## To Reproduce

1. Added blur option to navigation default options
2. Tested with other options and on Android
3. Tried upgrading to the latest version

## Expected behavior

The StatusBar should be blurred or have a background color. Background color can only be used on Android, so we need the blur feature on iOS so we overlay the status bar with a blurred background to not conflict with screen content.

## Actual Behavior

Nothing is being changed on iOS when we enable the blur option

## Your Environment

- React Native Navigation version: 7.8.1
- React Native version: 0.63.3
- Device info (Simulator/Device? OS version? Debug/Release?): Simulator and Device - all versions

## Reproducible Demo

```js
Navigation.events().registerAppLaunchedListener(async () => {
  await Navigation.setDefaultOptions({
    statusBar: {
      visible: true,
      blur: true
    },
   ....
  })
}

yogevbd:
Closes #7185

### DIFF
--- a/lib/ios/UIViewController+RNNOptions.m
+++ b/lib/ios/UIViewController+RNNOptions.m
@@ -126,7 +126,7 @@ const NSInteger BLUR_STATUS_TAG = 78264801;
                 initWithEffect:[UIBlurEffect effectWithStyle:UIBlurEffectStyleLight]];
             blur.frame = [[UIApplication sharedApplication] statusBarFrame];
             blur.tag = BLUR_STATUS_TAG;
-            [self.view insertSubview:blur atIndex:0];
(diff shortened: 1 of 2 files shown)


fadinasr:
Hi @yogevbd and @guyca,  is that released or not yet? I've upgraded to 7.18.1 and it's still not working.

PR #7196, opened 2021-07-20
Fix persenting react-native modals from overlays

yogevbd:
Presenting react-native declarative modals from an overlay freeze the app as the overlay window didn't pass the touch event to the underlying window.


tomasfrancomediamonks:
@yogevbd this is still happening in version 7.20.0

PR #7200, opened 2021-07-22
[WIP] ReactNative modals android support

swabbass:
support overlays and match ios


swabbass:
Ditched this in favour of more elegant solution #7227.

PR #7235, opened 2021-08-31
Component Lifecycle events fix onPause/onResume

swabbass:
# Issue:
Unfinished work from #7177. (my bad)
Modals and overlays did not receive events `componentDidAppear` and `componentDisappear` when Activity was moving to Pause or Resume State on Android, which caused the underlying views to receive `componentDidAppear` while there are modals on top.

# Fix:
- Overlays must receive those events since they are the highest layer.
- Top modal (in Modal stack) child should receive events, and not pass to underlying views.
- When no modals are shown, the currently visible child will send the events. 

CC: @yogevbd


lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/navigator/Navigator.java
```diff
@@ -266,10 +266,20 @@ CoordinatorLayout getOverlaysLayout() {
     }
 
     public void onHostPause() {
-        super.onViewDisappear();
+        overlayManager.onHostPause();
+        if (!modalStack.isEmpty()) {
```
guyca, line 270: I think this is not 100% correct. If the currently presented modal is displayed with `modalPresentationStyle: 'overCurrentContext'`, then the root layout also needs to get a disappear event.

lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/navigator/Navigator.java
```diff
@@ -266,10 +266,20 @@ CoordinatorLayout getOverlaysLayout() {
...
+            modalStack.onHostPause();
+        } else {
+            onViewDisappear();
+        }
     }
 
     public void onHostResume() {
-        super.onViewDidAppear();
+        overlayManager.onHostResume();
+        if (!modalStack.isEmpty()) {
```
guyca, line 279: Same as the comment above. If the modal is displayed with  `overCurrentContext`, the root layout also needs to receive an appear event.

guyca (changes requested):

lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/navigator/Navigator.java
swabbass: hmm, agree, knew that I'm missing something, checking options will do the work.
Thx dude

guyca (approved):

swabbass:
@guyca  thx for the review :]

PR #7238, opened 2021-09-01
Change bottom tabs height constant value to the dimen-based one

svbutko:
`AHBottomNavigation` which is used as bottom tabs on Android, allows users the ability to override different values using 
`res/values/dimens.xml:` file, like so:

```
<dimen name="bottom_navigation_height">56dp</dimen>
<dimen name="bottom_navigation_text_size_active">14sp</dimen>
<dimen name="bottom_navigation_text_size_inactive">12sp</dimen>
```

And it also stores some of the values as default ones (_e.g. bottom_navigation_height - 56dp_).

My primary concern is `bottom_navigation_height`, on rare occasions there's a need to change the size and since it's hard to pass an option without changing `AHBottomNavigation` it remains the only way to change the size (_I might be wrong here but based on `AHBottomNavigation` docs and code it looks like it's impossible_)

So instead of returning a hardcoded constant I think it would be better to rely on `bottom_navigation_height` when possible.


swabbass:
@svbutko looks amazing :], we can add such capability to [AHBottomNavigatipn](https://github.com/wix-playground/ahbottomnavigation) since we have it forked we can customize it as we wish :] maybe add an option to the current `Options`, if you have the time then you are welcome, for now, this PR looks good for what we have today.

svbutko:
> @svbutko looks amazing :], we can add such capability to [AHBottomNavigatipn](https://github.com/wix-playground/ahbottomnavigation) since we have it forked we can customize it as we wish :] maybe add an option to the current `Options`, if you have the time then you are welcome, for now, this PR looks good for what we have today.

Thanks, yes I'm at the moment working to make this possible on iOS using `Options` and after that will try to change it in the `AHBottomNavigation` fork to make it using `Options` too. 

But I think it would be better to make it as a separate PR.

PR #7241, opened 2021-09-02
Fix remx optional dependency

yogevbd:
Currently, remx is an optional peer dependency that is required for our mock consumers only. But it is actually being imported from our main `index.ts` which break users who don't import it even though they are not using mocks.
The solution is to change the way users import our mocks and make it explicit. 

Mocks now should be imported from `react-native-navigation/Mock`.


yogevbd:
@lindesvard Could you please verify that this fix solves your current issue?

lindesvard:
Will test it now! 

For people who use the mock outside this repo I assume they would need to access it from `react-native-navigation/lib/distMock`. Might need to update documentation if you have any on this.

yogevbd:
@lindesvard with latest commit you can import `react-native-navigation/Mock`

PR #7250, opened 2021-09-05
Fix color processing on iOS

yogevbd:
On iOS, `DynamicColorIOS` should never receive undefined or null values, therefor there is no reason to always pass the native `DynamicColorIOS` even when the color value is undefined.


swabbass:
@yogevbd did you check that this did not break android?

PR #7253, opened 2021-09-06
Fix left and right top bar buttons with component crash on Android

### LINKED ISSUE #7229: [Android] navigation v.7.19.0 RN: 0.65.1  Error for Android the mergeOptions (rightButtons->component), crach app
## 🐛 Bug Report

about:
"react-native": "0.65.1",
"react-native-navigation": "7.19.0",

(Write your steps here:)

2. `Navigation.mergeOptions(SEARCH_MAIN, {
        topBar: {
          rightButtons: [
            {
              id: 'avatarSearchButton',
              component: {
                id: AVATAR_MAIN,
                name: AVATAR_MAIN,
                passProps: {
                  source: props.profile.avatar,
                  borderColor: colors.WHITE,
                  isStatus: false,
                  onPress: this.handleProfilePress,
                },
              },
            },
          ],
        },
      });`

error for Android
`com.facebook.react.uimanager.IllegalViewOperationException: Trying to attach a ReactRootView with an explicit id already set to [1]. React Native uses the id field to track react tags and will overwrite this field. If that is fine, explicitly overwrite the id field to View.NO_ID.
        at com.facebook.react.ReactRootView.attachToReactInstanceManager(ReactRootView.java:644)
        at com.facebook.react.ReactRootView.onMeasure(ReactRootView.java:169)
        at com.reactnativenavigation.views.stack.topbar.titlebar.TitleBarReactButtonView.onMeasure(TitleBarReactButtonView.java:27)
        at android.view.View.measure(View.java:22216)
        at android.view.ViewGroup.measureChildWithMargins(ViewGroup.java:6671)
        at androidx.appcompat.widget.LinearLayoutCompat.measureChildBeforeLayout(LinearLayoutCompat.java:1387)
        at androidx.appcompat.widget.LinearLayoutCompat.measureHorizontal(LinearLayoutCompat.java:1034)
        at androidx.appcompat.widget.LinearLayoutCompat.onMeasure(LinearLayoutCompat.java:563)
        at androidx.appcompat.widget.ActionMenuView.onMeasure(ActionMenuView.java:173)
        at android.view.View.measure(View.java:22216)
        at androidx.appcompat.widget.Toolbar.measureChildConstrained(Toolbar.java:1645)
        at androidx.appcompat.widget.Toolbar.onMeasure(Toolbar.java:1738)
        at android.view.View.measure(View.java:22216)
        at com.reactnativenavigation.views.stack.topbar.titlebar.TitleAndButtonsContainer.measureTitleComponentExact(TitleAndButtonsContainer.kt:146)
        at com.reactnativenavigation.views.stack.topbar.titlebar.TitleAndButtonsContainer.onMeasure(TitleAndButtonsContainer.kt:138)
        at android.view.View.measure(View.java:22216)
        at android.view.ViewGroup.measureChildWithMargins(ViewGroup.java:6671)
        at android.widget.LinearLayout.measureChildBeforeLayout(LinearLayout.java:1539)
        at android.widget.LinearLayout.measureVertical(LinearLayout.java:823)
        at android.widget.LinearLayout.onMeasure(LinearLayout.java:702)
        at android.view.View.measure(View.java:22216)
        at android.view.ViewGroup.measureChildWithMargins(ViewGroup.java:6671)
        at android.widget.FrameLayout.onMeasure(FrameLayout.java:185)
        at android.view.View.measure(View.java:22216)
        at android.view.ViewGroup.measureChildWithMargins(ViewGroup.java:6671)
        at android.widget.LinearLayout.measureChildBeforeLayout(LinearLayout.java:1539)
        at android.widget.LinearLayout.measureVertical(LinearLayout.java:823)
        at android.widget.LinearLayout.onMeasure(LinearLayout.java:702)
        at com.google.android.material.appbar.AppBarLayout.onMeasure(AppBarLayout.java:419)
        at android.view.View.measure(View.java:22216)
        at android.view.ViewGroup.measureChildWithMargins(ViewGroup.java:6671)
        at androidx.coordinatorlayout.widget.CoordinatorLayout.onMeasureChild(CoordinatorLayout.java:760)
        at androidx.coordinatorlayout.widget.CoordinatorLayout.onMeasure(CoordinatorLayout.java:833)
        at android.view.View.measure(View.java:22216)
        at android.view.ViewGroup.measureChildWithMargins(ViewGroup.java:6671)
        at androidx.coordinatorlayout.widget.CoordinatorLayout.onMeasureChild(CoordinatorLayout.java:760)
        at androidx.coordinatorlayout.widget.CoordinatorLayout.onMeasure(CoordinatorLayout.java:833)
        at android.view.View.measure(View.java:22216)
        at android.view.ViewGroup.measureChildWithMargins(ViewGroup.java:6671)
        at androidx.coordinatorlayout.widget.CoordinatorLayout.onMeasureChild(CoordinatorLayout.java:760)
        at androidx.coordinatorlayout.widget.CoordinatorLayout.onMeasure(CoordinatorLayout.java:833)
        at android.view.View.measure(View.java:22216)
        at android.view.ViewGroup.measureChildWithMargins(ViewGroup.java:6671)
        at android.widget.FrameLayout.onMeasure(FrameLayout.java:185)
        at androidx.appcompat.widget.ContentFrameLayout.onMeasure(ContentFrameLayout.java:146)
        at android.view.View.measure(View.java:22216)
        at android.view.ViewGroup.measureChildWithMargins(ViewGroup.java:6671)
        at android.widget.LinearLayout.measureChildBeforeLayout(LinearLayout.java:1539)`

svbutko:
Fixes #7229 

It also happens with left button.

Also noticed that width spec was used for the height measurement (probably typo).

### DIFF
--- a/lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/TitleBarReactButtonView.java
+++ b/lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/TitleBarReactButtonView.java
@@ -24,7 +24,7 @@ public class TitleBarReactButtonView extends ReactView {
 
     @Override
     protected void onMeasure(int widthMeasureSpec, int heightMeasureSpec) {
-        super.onMeasure(createSpec(widthMeasureSpec, component.width), createSpec(widthMeasureSpec, component.height));
+        super.onMeasure(createSpec(widthMeasureSpec, component.width), createSpec(heightMeasureSpec, component.height));
     }
 
     private int createSpec(int measureSpec, Number dimension) {
--- a/lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/TitleBarReactButtonView.java
+++ b/lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/TitleBarReactButtonView.java
@@ -24,6 +24,10 @@ public class TitleBarReactButtonView extends ReactView {
 
     @Override
     protected void onMeasure(int widthMeasureSpec, int heightMeasureSpec) {
+        if (!this.isAttachedToWindow() || this.getReactInstanceManager() == null) {
+            this.setId(View.NO_ID);
+        }
+
         super.onMeasure(createSpec(widthMeasureSpec, component.width), createSpec(heightMeasureSpec, component.height));
     }
 
--- a/lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/TitleBarReactButtonView.java
+++ b/lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/TitleBarReactButtonView.java
@@ -25,6 +25,9 @@ public class TitleBarReactButtonView extends ReactView {
 
     @Override
     protected void onMeasure(int widthMeasureSpec, int heightMeasureSpec) {
+        
+        //This is a workaround, ReactNative throws exception when views have ids, On android MenuItems 
+        // With ActionViews like this got an id, see #7253
         if (!this.isAttachedToWindow() || this.getReactInstanceManager() == null) {
             this.setId(View.NO_ID);
         }
--- a/lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/TitleBarReactButtonView.java
+++ b/lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/TitleBarReactButtonView.java
@@ -2,6 +2,7 @@ package com.reactnativenavigation.views.stack.topbar.titlebar;
 
 import android.annotation.SuppressLint;
 import android.content.Context;
+import android.view.View;
 
 import com.facebook.react.ReactInstanceManager;
 import com.reactnativenavigation.options.ComponentOptions;


swabbass:
@svbutko really appreciated the contribution, can you please explain the solution to why it fixes the issue?

svbutko:
@swabbass when `onMeasure` is happening under the hood in `ReactRootView` in the very first run (_width = 0, height = 0_) the initialisation of `ReactInstanceManager` is happening and the view attaches to that instance (`IsAttachedToInstance`).

When this attachment is happening, the very first thing it does is checking whether the view already been attached or not, it's been done by checking the current ID `this.getId() != -1` and if it's -1 it throws error like what we have now.

Somehow when the actual width and height are coming the view is not yet attached to instance or react instance manager doesn't exist yet BUT this view has an ID already which leads to the error.

My solution isn't the best but I couldn't find a better one.

svbutko:
This ID check btw was added in 0.65:

https://github.com/facebook/react-native/commit/4f3b17412018a10f9293247c802598d2b94a844b#diff-c6c68fb9580db5698d241fdd97b31a25ca1fdc061023a1bf1959f64fc4a00743

swabbass:
Great, now it sounds clear, let's check the `title.component` also for such thing, if that the case, then we can do `setId` as early as possible, hence, `setId` in C'tor. 

We did not do RN65 upgrade till now, but it is in the works, and such bugs, will help us find out what really they are trying to do.

from the dev note:
```
The RootView being managed by Fabric should have an id of View.NO_ID when it is "handed over" to RN. This is true for Fabric and non-Fabric
and setting a custom id on the ReactRootView has never been supported. I'm temporarily (?) adding an additional check earlier and into ReactRootView to hopefully
catch any of these issues early.
```

looks like it is worth reporting such a case to them in order to see what they are trying to achieve.

what concerns me is, we are not setting a custom id, anywhere for such views, which might trigger this flow, we can fix this by not setting the id, if we found out where it is being done.

svbutko:
> Great, now it sounds clear, let's check the title.component also for such thing, if that the case, then we can do setId as early as possible, hence, setId in C'tor.

I'll check it but I think the result will be the same.

> what concerns me is, we are not setting a custom id, anywhere for such views, which might trigger this flow, we can fix this by not setting the id, if we found out where it is being done.

Same thing, that's why I didn't add it at initialisation because I thought it is being set somewhere but didn't find the place (though I didn't spend a lot of time searching for it manually).

swabbass:
I think it is being done on their side, I looked for setId in our code and did not found something, but, we can report such a thing to RN and tag that dev that added the check so he can play and find, and maybe remove this exception.

swabbass:
Hey @JoshuaGross, I think we found out a case where your exception is triggered, you can play with our playground to debug what is happening, what is important to note is that we don't set id for natively hosted ReactComponents, so suspect that this happens internally in RN, hope that helps.

svbutko:
> Great, now it sounds clear, let's check the `title.component` also for such thing

@swabbass checked it and it works fine and at the first `onMeasure` it has ID -1

svbutko:
Also regarding the playground, shouldn't it be first updated from #7034 ?

swabbass:
> Also regarding the playground, shouldn't it be first updated from #7034 ?

yes but since there are a lot of breaking changes, it is healthy to take master and bump rn only to rn65 just to check this case.

swabbass:
> > Great, now it sounds clear, let's check the `title.component` also for such thing
> 
> @swabbass checked it and it works fine and at the first `onMeasure` it has ID -1

Thats weird, both `title.component` and `button.component` has the same logic, if our analysis is right, both title and button should have the same issue, but to be honest, I did a hack their in `onMeasure` , [see](https://github.com/wix/react-native-navigation/blob/master/lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/TitleBarReactView.kt#L17), maybe if we apply the same hack to [buttons](https://github.com/wix/react-native-navigation/blob/master/lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/TitleBarReactButtonView.java#L27) would solve the problem without setting the id.

swabbass:
I think that MenuItem.ActionView might set an id to a view, that could be the case, like let's log the id of the view once it is created [here](https://github.com/wix/react-native-navigation/blob/master/lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/stack/topbar/button/ButtonController.kt#L60), and in `onMeasure`, if that is the case, then it might affect, removing buttons, and modifying them. 

we need to be sure that the current functionality won't get changed, because of RN65, hence, the id might come from the menu, which we are using to modify it or remove it.

svbutko:
I'll take a look soon, thanks for direction

svbutko:
> I think that MenuItem.ActionView might set an id to a view, that could be the case, like let's log the id of the view once it is created [here](https://github.com/wix/react-native-navigation/blob/master/lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/stack/topbar/button/ButtonController.kt#L60), and in `onMeasure`, if that is the case, then it might affect, removing buttons, and modifying them.
> 
> we need to be sure that the current functionality won't get changed, because of RN65, hence, the id might come from the menu, which we are using to modify it or remove it.

In `createView` it's again -1 but in `onMeasure` the ID is 1 already

swabbass:
that means something in between sets the id, now if we add 3 buttons (different components), I think that their ids will be 1,2,3 ? i might be wrong but it could be that toolbars on android  set ids to the views or RN somewhere sets an id, either way:
- we need to check that your change does not affect buttons addition/modifying/removing ability.
- looks like this exception was thrown for a good reason, lets see what the dev who add it say.

yogevbd:
@swabbass we need this for upgrading the Wix app to RN 0.65. Is there anything else to be done here? 
otherwise we can merge and open a PR for the title component if needed.

Thanks @svbutko 💪💪

swabbass:
> @swabbass we need this for upgrading the Wix app to RN 0.65. Is there anything else to be done here?
> otherwise we can merge and open a PR for the title component if needed.
> 
> Thanks @svbutko 💪💪

I wrote what need to be checked :], mainly we need to check that your change does not affect buttons addition/modifying/removing ability.

Issue #7255, opened 2021-09-09
PlatformColor crashes app when adding it as dark/light color on Android


svbutko:
## 🐛 Bug Report

Application crashes on Android when using the new way of adding dark and light theme colors with `PlatformColor`

### Have you read the [Contributing Guidelines on issues](https://wix.github.io/react-native-navigation/docs/meta-contributing/)?

Yes

## To Reproduce

1. Set default options with, for instance `topBar`
2. Add color to `background` like so:
```
        color: {
          light: PlatformColor("?android:attr/colorBackground"),
          dark: PlatformColor("?android:attr/textColorPrimary"),
        },
```

## Expected behavior

Colors are set correctly and app works fine

## Actual Behavior

App crashes:

```
    com.facebook.react.bridge.JSApplicationCausedNativeException: ColorValue: the value must be a number or Object.
        at com.facebook.react.bridge.ColorPropConverter.getColor(ColorPropConverter.java:77)
        at com.reactnativenavigation.options.parsers.ColorParser.parse(ColorParser.kt:33)
        at com.reactnativenavigation.options.params.ThemeColour$Companion.parse(ThemeColour.kt:44)
        at com.reactnativenavigation.options.params.ThemeColour.parse(Unknown Source:2)
        at com.reactnativenavigation.options.TopBarBackgroundOptions.parse(TopBarBackgroundOptions.java:18)
        at com.reactnativenavigation.options.TopBarOptions.parse(TopBarOptions.java:34)
        at com.reactnativenavigation.options.Options.parse(Options.java:22)
        at com.reactnativenavigation.react.NavigationModule.parse(NavigationModule.java:215)
        at com.reactnativenavigation.react.NavigationModule.lambda$setDefaultOptions$1$NavigationModule(NavigationModule.java:120)
        at com.reactnativenavigation.react.-$$Lambda$NavigationModule$vc1HzyGbL5Tqc0F69ZZFUMbcIB0.run(Unknown Source:4)
        at com.reactnativenavigation.react.NavigationModule.lambda$handle$14$NavigationModule(NavigationModule.java:221)
        at com.reactnativenavigation.react.-$$Lambda$NavigationModule$sXUG5B2EQ5skHUFYfoWeiHKR6OE.run(Unknown Source:4)
        at android.os.Handler.handleCallback(Handler.java:938)
        at android.os.Handler.dispatchMessage(Handler.java:99)
        at android.os.Looper.loop(Looper.java:223)
[...]

Issue #6618, opened 2020-09-28
PlatformColor crashes on Android


rosskhanas:
### Issue Description

The app crashes when PlatformColor is defined via `?attr/XXX`.

### Steps to Reproduce / Code Snippets / Screenshots

```
Navigation.setDefaultOptions({
  backButton: {
    color: PlatformColor("?attr/colorOnBackground"),
  },
}
```

---
### Environment
* React Native Navigation version: 7.0.0
* React Native version: 0.63.2
* Platform(s) (iOS, Android, or both?): Android
* Device info (Simulator/Device? OS version? Debug/Release?): Debug

guyca:
@rosskhanas Have your tried with one of the latest snapshot versions? `7.0.0-snapshot.1225` perhaps. I think this issue is already resolved by @danilobuerger

rosskhanas:
@guyca hm, it seems not fixed. Everything works well when I use `PlatformColor("@color/XXX")`, but not when `?attr/XXX`.

Here are more details about the crash:

```
com.facebook.react.bridge.JSApplicationCausedNativeException: ColorValue: None of the paths in the `resource_paths` array resolved to a color resource.
    at com.facebook.react.bridge.ColorPropConverter.getColor(ColorPropConverter.java:71)
    at com.reactnativenavigation.options.parsers.ColorParser.parse(ColorParser.java:25)
    at com.reactnativenavigation.options.TitleOptions.parse(TitleOptions.java:31)
    at com.reactnativenavigation.options.TopBarOptions.parse(TopBarOptions.java:33)
    at com.reactnativenavigation.options.Options.parse(Options.java:22)
    at com.reactnativenavigation.react.NavigationModule.parse(NavigationModule.java:194)
    at com.reactnativenavigation.react.NavigationModule.lambda$setDefaultOptions$1$NavigationModule(NavigationModule.java:104)
    at com.reactnativenavigation.react.-$$Lambda$NavigationModule$N2JlaPHmovMYN5T38Ah4GsF6njg.run
    at com.reactnativenavigation.react.NavigationModule.lambda$handle$13$NavigationModule(NavigationModule.java:200)
    at com.reactnativenavigation.react.-$$Lambda$NavigationModule$_KMLiCFir6_k6qF7P2pzVizFOwk.run
    at android.os.Handler.handleCallback(Handler.java:873)
    at android.os.Handler.dispatchMessage(Handler.java:99)
    at android.os.Looper.loop(Looper.java:193)
[...]

PR #7256, opened 2021-09-09
Fix PlatformColor parsing on Android when using dark and light theme

svbutko:
Fixes: #7255


svbutko:
I checked #6618 and it partially exists until this time, like I can't use a primary color like `?attr/colorPrimary` if I use material theme (e.g. `Theme.MaterialComponents.DayNight`) but it might have something to do with this:

https://github.com/facebook/react-native/issues/29763

svbutko:
Maybe it would be better to completely change this behaviour and always put `JSONParser.convert(color)` because under the hood the value expects to be of `ReadableMap` type?

yogevbd:
@swabbass wdyt?

swabbass (approved):

swabbass:
you should not assign  `PlatformColor`  value to dark/light, it should replace the whole object since PlatformColor is a native way to achieve dark/light, hence, you define the same value in for example style.xml and style.night.xml and then assign the value to it, not assigning same value for both light and dark which is that what was being done natively

PR #7216, opened 2021-08-13
PlatformColor Support

swabbass:
Following the discussion here: #6618.

## Android
- Colours are parsed natively.
- Creating custom colour when a PlatformColor is passed to native.
- Both light/dark values have the same res attribute resolved when `get` is called.

## IOS
- Added a custom colour.
- Changed storyboard splash screen.
- Using `labelColor` as a test.

CC: @yogevbd 
closes: #6618

PR #7259, opened 2021-09-14
Fix empty dynamic colors object when using PlatformColors on iOS

svbutko:
Fixes the original error mentioned here: https://github.com/wix/react-native-navigation/pull/7216#issuecomment-918975922


svbutko:
Since `ThemeColor` type are allowing us to have one or both values of `light` and `dark` to be `undefined` the current code will be producing the same error

yogevbd (dismissed):

swabbass:
@yogevbd is this done?

yogevbd (approved):

PR #7260, opened 2021-09-16
Declarative modals on Android

swabbass:
## Motivation
We have our own declarative style modals, which can be used like **ReactNative** modals.
When using **ReactNative** modals it causes a problematic behaviour since those components are displayed out of our rules (order of modals and overlays), on Android they use Dialog which creates a DecorView on top of all views.

## Concept
We are defining a "Host" Component which will be the host of the modal content, and we will build a new component that "hunts" the host children and add them to a native view exported via a **ViewManager**, calling the internal show, dismiss APIs for modals when changing `visible` prop.

## Demo

https://user-images.githubusercontent.com/7227793/133554728-abd642a0-87e9-454a-afdb-e1e6118d514d.mov


lib/android/app/src/main/java/com/reactnativenavigation/react/NavigationModule.java
```diff
@@ -7,11 +7,14 @@
 import com.facebook.react.bridge.Arguments;
 import com.facebook.react.bridge.Promise;
 import com.facebook.react.bridge.ReactApplicationContext;
+import com.facebook.react.bridge.ReactContext;
```
swabbass, line 10: should remove

lib/android/app/src/main/java/com/reactnativenavigation/react/modal/ModalLayoutController.kt
```diff
@@ -0,0 +1,51 @@
...
+) : ViewController<ModalFrameLayout>(activity, id, yellowBoxDelegate, initialOptions, overlay) {
+    override fun isViewShown(): Boolean {
+        return !isDestroyed && view != null && view!!.isShown
+    }
+
+    override fun isRendered(): Boolean {
+        return isViewCreated
+    }
+
+    override fun getCurrentComponentName(): String = "GenericDeclaredLayoutController"
```
swabbass, line 29: rename this to modal

lib/src/components/RNNModal.tsx
```diff
@@ -0,0 +1,46 @@
...
+  presentationStyle: 'unspecified' | 'none' | 'overCurrentContext';
+  blurOnUnmount: boolean;
+  onShow?: () => any;
+  onRequestDismiss: () => any;
+  visible: boolean;
+}
+
+const RNNModalViewManager = requireNativeComponent('RNNModalViewManager');
+
+export class RNNModal extends React.Component<RNNModalProps> {
```
swabbass, line 14: renmae to modal

yogevbd (approved):

PR #7270, opened 2021-09-23
Fix iOS 15 bars appearance

### LINKED ISSUE #7266: bottomTabs: { backgroundColor } seems to have no effect (iOS 15/Xcode 13)
## 🐛 Bug Report

Building with Xcode 13 seems to have broken the bottom tab background color option. The color specified via `Navigation.setDefaultOptions` has no effect and the bottom tabs are always mounted with the default color.

My bottom tabs each have stack navigation within them. I've noticed that the correct BG color takes effect on some screens after being pushed, and that the bottom tabs remain the correct color as long as that screen is mounted (even if I navigate to other tabs). But as soon as that screen is popped the background color reverts to the default. I haven't found a reason for this yet... I do not alter the bottom tab options ever other than in the initial `setDefaultOptions`

### LINKED ISSUE #7268: TopBar noBorder property not working on iOS 15
## 🐛 Bug Report
Regardless of the value of `topBar.noBorder`, the bottom border below the topbar is still drawn on iOS 15.

Yes

## To Reproduce
1. Pulled an existing repo running correctly on iOS 14 that has `noBorder:true`
2. Recompiled in Xcode 13 in Release for an iOS 15 device
3. The topbar bottom border is drawn

## Expected behavior
The bottom border should not be there (it wasn't when the device is running iOS 14)

## Actual Behavior
The bottom border is there

## Your Environment

- React Native Navigation version: 7.21.0
- React Native version: 0.64.2
- Device info (Simulator/Device? OS version? Debug/Release?): Device, iOS 15, Release

### Are you willing to resolve this issue by submitting a Pull Request?

 - ✖️  &nbsp; Yes, I have the time, and I know how to start.
 - ✖️  &nbsp; Yes, I have the time, but I don't know how to start. I would need guidance.
 - ✅  &nbsp; No, I don’t have the time and I’m okay to wait for the community / maintainers to resolve this issue.

habovh:
Following #7266 and especially [my comment there](https://github.com/wix/react-native-navigation/issues/7266#issuecomment-925706532), I'm opening this PR to provide a quick fix to people experiencing issues when building RNN using Xcode 13.

This should be considered a fix PR rather than a feature PR, as what it's doing is basically duplicating the already set `standardAppearance` to a `scrollEdgeAppearance`. What this means is that the bars will have the same visual aspect whether they're displayed alongside the edge of a scroll view.

In a future, more elaborate PR, we could imagine a way to be able to provide `scrollEdgeAppearance` values from JS for more control over the native bars.

Fixes #7266
Fixes #7268


wilkuintheair:
Thanks @habovh 
I've applied your patch and it worked for me 💪

yogevbd (approved):

yogevbd:
Fixed failing build and refactored a bit. we currently running an old xcode version on our CI, that's why the build failed.
Awesome PR 💪

lib/ios/TabBarItemIOS15Creator.m
```diff
@@ -0,0 +1,37 @@
...
+@implementation TabBarItemIOS15Creator
+
+#if __IPHONE_OS_VERSION_MAX_ALLOWED >= 150000
+
+- (UITabBarItem *)createTabBarItem:(UITabBarItem *)mergeItem {
+    UITabBarItem *tabBarItem = [super createTabBarItem:mergeItem];
+    tabBarItem.scrollEdgeAppearance =
+        mergeItem.scrollEdgeAppearance ?: [[UITabBarAppearance alloc] init];
+    return tabBarItem;
+}
```
habovh, line 12: @yogevbd thanks for looking into this PR! Out of curiosity, don't we need to set the regular appearance as well here? I'm asking because in `lib/ios/BottomTabPresenterCreator.m` you have a strict if-else not calling the pre-iOS 15 tab item creator.

habovh:
@wilkuintheair you might want to check the changes that @yogevbd just made: he added macros that should allow you to compile on earlier versions of Xcode as well (among other things).

lib/ios/TabBarItemIOS15Creator.m
yogevbd: It will be created in the super call [here](https://github.com/wix/react-native-navigation/pull/7270/files#diff-729e5adb5dc3698d4499a5552e0e48a708328abad784540717995920364208efR7). I know it's not straight forward, will improve this soon.

habovh: Oh ok, I missed that sorry! Your refactoring makes much more sense now!

yogevbd (approved):

ptfly:
This PR added a visible border to the bottom tab bar which cannot be controlled through the options - `borderColor` & `borderWidth`. Now it looks like I have to borders, one controllable and one - not!

svbutko:
> This PR added a visible border to the bottom tab bar which cannot be controlled through the options - `borderColor` & `borderWidth`. Now it looks like I have to borders, one controllable and one - not!

Can you create an issue about that according to issue template with reproduction demo then we can take a look at it

PR #7271, opened 2021-09-23
Add SF Symbols support to bottom tab and top bar's back button

svbutko:
Since SF Symbols are built in starting from iOS 13 it makes sense to use them as icons rather than creating PNGs which will increase app size.

image

image


yogevbd (approved):

svbutko:
Test are failing correctly since SF Symbols are supported starting from iOS 13 and I did `available` checks but still this tests are running

yogevbd (changes requested):

svbutko:
@yogevbd, fixed the issue with tests

yogevbd (approved):

PR #7276, opened 2021-09-29
Remove default splash layout color on Android

svbutko:
User can set their own `windowBackground` in his xml style to match the theme


swabbass:
Legit!

PR #7279, opened 2021-10-01
use HMSegmentedControl from pod instead of standalone file

mohsinulhaq:
https://github.com/HeshamMegid/HMSegmentedControl
is available as a Pod and many libraries and frameworks use it from there. Using them along with ReactNativeNavigation causes duplicate symbols error when building the app. I have updated the library to use the class from pod instead of the standalone files embedded within the library.


mohsinulhaq:
@swabbass can you please approve the workflow?

swabbass:
> @swabbass can you please approve the workflow?

Hey, thx for mentioning such an issue.

@yogevbd if there is no custom code added to that library main functionality we can just add it, wdyt?

yogevbd (approved):

Issue #7264, opened 2021-09-17
Parameter 'translationStart' is never used for RN version >= 0.65


harleenarora:
## 🐛 Bug Report

Although this is not a bug per se, react-native-navigation is not optimized for RN version >= 0.65

When I build Android I get a bunch of not used parameters warnings

> `w: /node_modules/react-native-navigation/lib/android/app/src/main/java/com/reactnativenavigation/options/ValueAnimationOptions.kt: (63, 25): The corresponding parameter in the supertype 'Any' is named 'other'. This may cause problems when calling this function with named arguments.
> w: /node_modules/react-native-navigation/lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/stack/topbar/TopBarAnimator.kt: (7, 22): Parameter 'translationStart' is never used
> w: /node_modules/react-native-navigation/lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/stack/topbar/TopBarAnimator.kt: (7, 47): Parameter 'translationEndDy' is never used
> w: /node_modules/react-native-navigation/lib/android/app/src/main/java/com/reactnativenavigation/views/element/animators/PropertyAnimatorCreator.kt: (17, 44): Unchecked cast: View to T
> w: /node_modules/react-native-navigation/lib/android/app/src/main/java/com/reactnativenavigation/views/element/animators/PropertyAnimatorCreator.kt: (17, 53): Unchecked cast: View to T
> w: /node_modules/react-native-navigation/lib/android/app/src/main/java/com/reactnativenavigation/views/element/animators/PropertyAnimatorCreator.kt: (25, 91): Unchecked cast: Type! to Class<T>`

PR #7288, opened 2021-10-05
Fix Android project warnings

svbutko:
Also fixes: #7264

### DIFF
--- a/lib/android/app/src/main/java/com/reactnativenavigation/options/ElementTransitionOptions.kt
+++ b/lib/android/app/src/main/java/com/reactnativenavigation/options/ElementTransitionOptions.kt
@@ -1,7 +1,6 @@
 package com.reactnativenavigation.options
 
 import android.animation.Animator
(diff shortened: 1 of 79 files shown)


swabbass:
That is Great!.. Thanx dude.
How come we did not face such stuff in our #7034?

swabbass (approved):
You are awesome!

svbutko:
I just noticed an issue that there are some logs from LogCat and then checked warnings in Android Studio and I was shocked a bit 😄

svbutko:
About to do the same with iOS, I've seen several warnings in Xcode before

swabbass:
> I just noticed an issue that there are some logs from LogCat and then checked warnings in Android Studio and I was shocked a bit 😄

Yup since we are 2, one android dev and one iOS, we did not have time to deal with such stuff since there is more prioritized stuff, we are blessed that you are here dude!

PR #7291, opened 2021-10-05
Add react activity nullable check

### LINKED ISSUE #7290: null cannot be cast to non-null type com.reactnativenavigation.NavigationActivity
## 🐛 Bug Report

Upgrading from 7.21.0 -> 7.22.1 causes above error on Android only. 
No other change was made other than upgrading RNN package.

Yes

## Actual Behavior

IMG_1018

## Your Environment

- React Native Navigation version: 7.22.1
- React Native version: 0.64.1
- Platform(s) (iOS, Android, or both?): Android
- Device info (Simulator/Device? OS version? Debug/Release?): Android Galaxy S20

**EDIT:**
Happens on launch, right after splash screen.
Doesn't seem to be related to modals.

Reproducible demo repo: https://github.com/nextriot/rnn-example-repo

svbutko:
@swabbass  I'll try to reproduce it tomorrow if user won't give more info but there's definitely a bug, I think it's most likely happen at the start of application and we should return an empty activity or null

Should fix #7290

### DIFF
--- a/lib/android/app/src/main/java/com/reactnativenavigation/react/NavigationPackage.kt
+++ b/lib/android/app/src/main/java/com/reactnativenavigation/react/NavigationPackage.kt
@@ -5,7 +5,6 @@ import com.facebook.react.ReactPackage
 import com.facebook.react.bridge.NativeModule
 import com.facebook.react.bridge.ReactApplicationContext
 import com.facebook.react.uimanager.ViewManager
-import com.reactnativenavigation.NavigationActivity
 import com.reactnativenavigation.options.LayoutFactory
 import com.reactnativenavigation.react.modal.ModalViewManager
 
@@ -22,6 +21,6 @@ class NavigationPackage(private val reactNativeHost: ReactNativeHost) : ReactPac
 
     override fun createViewManagers(reactContext: ReactApplicationContext): List<ViewManager<*, *>> {
 
-        return listOf(ModalViewManager((reactContext.currentActivity as NavigationActivity).navigator))
+        return listOf(ModalViewManager(reactContext))
     }
 }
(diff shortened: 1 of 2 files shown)


svbutko:
https://github.com/wix/react-native-navigation/issues/7290#issuecomment-935382063

As I thought

svbutko:
> > @swabbass I'll try to reproduce it tomorrow if user won't give more info but there's definitely a bug, I think it's most likely happen at the start of application and we should return an empty activity or null
> > Should fix #7290
> 
> It can happen since the JS thread is bound to the Application lifecycle, packages are being registered in `NavigationApplication`, not the activity, hence it can have non-null `reactContext` while no Activity at that point, what exactly is that flow? we need to check how ReactContext is created, for sure we know that `createViewManagers` might get called before the activity is created in some flow, and for that, we might return an empty list since its logical, no activity => no view managers, but we can also check alternative way to pass navigator to the `ModalViewManager`.
> 
> how about this, see the commit messages for details, https://github.com/wix/react-native-navigation/compare/null_activity_modal_viewmanager

Looks great, should fix that issue

svbutko:
@swabbass will you open a separate PR and we close this one?

svbutko:
Also can make a minor release with it and hack removal

swabbass:
> @swabbass will you open a separate PR and we close this one?

You can just merge the branch here and update the PR and merge it to master

svbutko:
> > @swabbass will you open a separate PR and we close this one?
> 
> I can open a new one :], you can also merge the branch here and remove the old changes :]

Ok, will do

swabbass:
> > > @swabbass will you open a separate PR and we close this one?
> > 
> > 
> > I can open a new one :], you can also merge the branch here and remove the old changes :]
> 
> Ok, will do

Here you go https://github.com/wix/react-native-navigation/pull/7292.

svbutko:
Closed in favour of #7292

PR #7292, opened 2021-10-06
Fix Null activity when having initial modal at startup

swabbass:
# Issue:

Null activity reference when opening an app with initial modal, while react contexts is initialized and no activity alive to pass navigator to `ModalViewManager`.

# Fix:

Determine `Navigator` reference on demand while checking valid activity state

Closes: #7290


svbutko (approved):

svbutko:
@swabbass can you merge it and ship as a patch release? (I don't have an access)

PR #7353, opened 2021-11-16
Add accessibilityRole to OptionsTopBarButton type interface

klammm:
This PR adds the [accessibilityRole](https://reactnative.dev/docs/accessibility#accessibilityrole) prop to the `OptionsTopBarButton` type interface. 

This prop is used to communicate the purpose of a component to screen-reader users.

### DIFF
--- a/lib/src/interfaces/Options.ts
+++ b/lib/src/interfaces/Options.ts
@@ -535,6 +535,10 @@ export interface OptionsTopBarButton {
    */
   accessibilityLabel?: string;
   /**
+   * Communicates the purpose of a component to the user of an assistive technology
+   */
+  accessibilityRole?: string;
+  /**
    * Set the font family for the button's text
    */
   fontFamily?: FontFamily;


swabbass:
Hey, You added the prop to the options interface, but it seems that it does not have any effect, are you planing to continue work on it? do you want help?

klammm:
Hi @swabbass , 

I added this `accessibilityRole` prop to the Options interface so that I may pass the prop down to the Top Buttons in my use case. I didn't want to hard-code in an `accessibilityRole="button"` myself in case others wanted to set another type of role, which explains the extent of this PR. I could add it in, but I wasn't sure where to in the code.

What are your thoughts?

swabbass:
Gotcha, you can start by trying to find how ReactNative implemented such feature, see their implementation for Android for [example](https://github.com/facebook/react-native/blob/main/ReactAndroid/src/main/java/com/facebook/react/uimanager/BaseViewManager.java#L160) 

and On iOS you might have a look at [this](https://github.com/facebook/react-native/blob/main/React/Views/RCTView.m#L323).

You will have all search results related to accessibilityRole [here](https://github.com/facebook/react-native/search?q=accessibilityRole).

klammm:
Ahh ok. I'll see what I can do to add in that native code. I've never worked with native code before, only with React Native's javascript. 

So I have to pass in the `accessibilityRole` prop into the TopBar button itself. Is that correct? 

I was under the assumption that all these props in [OptionsTopBarButton](https://github.com/wix/react-native-navigation/blob/931ed18decb68937ace61f7fe9a73f50f918e13e/lib/src/interfaces/Options.ts#L478) were already passed down to the button itself.

PR #7376, opened 2021-11-25
Fix lagging keyboard when using autoFocus

### LINKED ISSUE #7019: Keyboard is lagging when we navigate to a pushed stack
## 🐛 Bug Report

When we navigate to a new screen with autoFocus (_or using useEffect() => textInputRef.current.focus();_) the Keyboard animation is lagging on iOS. Looks okay in Android.

Yes.

## To Reproduce

1. Clone the project
2. Go to `bug/focus-keyboard-ios` branch
3. yarn ios
4. "Press me" to navigate to the lagging screen

## Expected behavior
The keyboard should appear without lag.
Usually the Keyboard animation comes from the bottom in iOS. Not sure why it's coming from right-to-left, may be related with the bug.

## Actual Behavior

https://user-images.githubusercontent.com/8364345/110526959-53c6e100-810e-11eb-98c8-3421586dda72.mov

## Your Environment

- React Native Navigation version: 7.11.3
- Device info (Simulator/Device? OS version? Debug/Release?): iPhone X / Emulator / Debug / Release

## Reproducible Demo

Clone and use this branch: https://github.com/nelsonprsousa/rnn-testing/tree/bug/focus-keyboard-ios

yogevbd:
When using TextInput's `autoFocus`, The keyboard view is trying to animate its slide-in transition within the context of a the screen push animation.
We solved this by swizzling RN code and delay the becomeFirstResponder to after the component is appeared.

Closes #7019


swabbass:
Isn't there a more elegant solution than swizzling, seems to me once the inner implementation changes or the name of the class changes it might break, for my Lil understanding, why not natively remove the focus of the screen from the beginning of the animation until the screen is fully visible and then change the focus to be true for that text input?

like save the autoFocus value, until the view is fully visible, the solutions [here](https://github.com/wix/react-native-navigation/issues/7019#issuecomment-804133678), is quite straightforward if it can be done natively :]

PR #7378, opened 2021-11-28
Fix More tab appearance on iOS

### LINKED ISSUE #7160: Issue rendering text on "More" tab page
## 🐛 Bug Report

When we have more than 5 tabs the `More` tab appears. When we click this the rest of the tabs appears. But they do not have any text. 

Screenshot 2021-06-21 at 15 14 49
Screen Recording 2021-06-21 at 15 11 19

Yes

## To Reproduce

1. Checkout: https://github.com/lindesvard/rnn-bugs/tree/more-tab
2. `yarn && yarn ios`
3. Click on the more tab

## Expected behavior

The rest of the tabs should be displayed with text

## Actual Behavior

The rest of the tabs are displayed but *without* text

## Your Environment

- React Native Navigation version: 7.16.0
- Platform(s): iOS
- Device info: iOS Sim 14.3 Debug

## Reproducible Demo

https://github.com/lindesvard/rnn-bugs/tree/more-tab

yogevbd:
There is no reason to wait for the child view controller to be added (`willMoveToParentViewController`) for applying its `bottomTab` appearance. It caused an issue in More tab because those view controllers are added to the parent controller in a different timing.

Closes #7160


lib/ios/RNNBottomTabsController.m
```diff
@@ -54,16 +56,18 @@ - (instancetype)initWithLayoutInfo:(RNNLayoutInfo *)layoutInfo
     return self;
 }
 
+- (void)createTabBarItems:(NSArray<UIViewController *> *)childViewControllers {
+    for (UIViewController *child in childViewControllers) {
```
swabbass, line 60: you might need to limit tabs for 5, be aligned with Android.

In android we cant have more than 5 tabs, RN apps will use the same amount of tabs for ios and for android, so apps with more than 5 tabs will crash the Android clients.

yogevbd: That's what the `More` tab is for. We have this natively supported on iOS.

PR #7380, opened 2021-11-29
Bottom tabs selection stack (HW button Android)

swabbass:
Implementing bottom tab history selection and handling hardware back buttons (Android only) with different behaviours.
Following this discussion: #7365.

https://user-images.githubusercontent.com/7227793/143840706-8800426a-d351-4a4f-9c21-97abe6a49c65.mov


lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/bottomtabs/BottomTabsController.java
```diff
@@ -156,7 +161,27 @@ public void mergeChildOptions(Options options, ViewController<?> child) {
...
+            } else if (options.hardwareBack.getBottomTabOnPress() instanceof HwBackBottomTabsBehaviour.JumpToFirst) {
+                if(getSelectedIndex() != 0){
+                    selectTab(0, false);
+                    return true;
+                }else{
+                    return false;
+                }
+            }else{
+                return false;
+            }
```
aliraza-noon, line 182:
```suggestion
                if (getSelectedIndex() != 0) {
                    selectTab(0, false);
                    return true;
                } else {
                    return false;
                }
            } else {
                return false;
            }
```

swabbass: spaces?

aliraza-noon: yes

yogevbd (approved):

PR #7411, opened 2021-12-19
Keyboard autofocus android FIX

swabbass:
#Issue:

When requesting autofocus on a screen that is being pushed with **animation**, insets will be applied once the view added to the view hierarchy, but when having animation the focus will be delayed until the view is fully visible in order for the edit text to take action and has a focus which will miss the first insets measurements.

#Fix:

Apply insets once the view is finished with the animation in order to resize/pan content when the view receives focus, once autofocus is passed to an EditText.

### DIFF
--- a/lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/component/ComponentViewController.java
+++ b/lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/component/ComponentViewController.java
@@ -2,6 +2,7 @@ package com.reactnativenavigation.viewcontrollers.component;
 
 import android.app.Activity;
 import android.content.res.Configuration;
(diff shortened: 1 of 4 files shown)


lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/component/ComponentViewController.java
```diff
@@ -2,6 +2,7 @@
 
 import android.app.Activity;
 import android.content.res.Configuration;
+import android.util.Log;
```
yogevbd, line 5:
```suggestion
```

yogevbd (approved):

yogevbd (approved):

PR #7416, opened 2021-12-28
Add layout.insets option

yogevbd:
This feature allows adding insets to the topmost view controller in the hierarchy, it is useful for presenting a banner across the app with overlay without having to re-render it each time a new screen is being pushed.


N3TC4T:
Thanks @yogevbd for the fix, just one more issue, setting `adjustResize` to `false` when setting the default options with `setDefaultOptions` won't apply the value. so adjustResize is always true.

yogevbd:
That's right, @swabbass we should resolve `adjustResize` option from default

PR #7446, opened 2022-02-09
[WIP] Fix topbar icon flickering Android

Andarius:
From what I see, the flickering is due to `applyButtonsOptions` being called too many times.

flicker

The tests are failing here with this fix so I'm not sure it's the right one:

image

Could I have your opinion on this @guyca @yogevbd  ?

Thanks!

TODO: clean playground once working


swabbass:
Hey, Good catch!!.

a little note: Can you please delete all the reformatting and keep the code as-is while showing the changes only.

Andarius:
Hey @swabbass, sure will do :)

I think the issue is that `applyButtonOptions`(https://github.com/wix/react-native-navigation/blob/7347ed6fa71781c7646b6aec45122a50f9b5aaac/lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/stack/topbar/TopBarController.kt) is called too many times (each time we click a bottom tab button, it's called).

I'm not sure which function is called too often between `mergeLeftButtonsOptions` / `mergeButtons` and `applyButtonOptions` though...

aashishshrestha5532 (approved):

Issue #7433, opened 2022-01-31
Keyboard issue


vimalswaroopj:
I am using react native navigation library in react native, the keyboard in android is pushing the UI up. If I remove some of the options such as:`navigationBar: {
        visible: false,
      },`
it works fine, but not all the time it works fine.

Can anyone give some hints or tips to solve this issue. 

We are using "KeyboardAwareScrollView" to make the keyboard scroll the view to bring the input filed on top of keyboard.

ItsNoHax:
If you check your AndroidManifest, what `android:windowSoftInputMode` are you using?

vimalswaroopj:
> If you check your AndroidManifest, what `android:windowSoftInputMode` are you using?

android:windowSoftInputMode="adjustPan"
tried "adjustResize" as well but the issue is still there

keeprock:
I got the same issue. It starts with 7.24.0. 7.24.1 doesn’t' fix it either. The latest release (7.25.1) still contains it.

Bug looks like this:

WIthout keyboard:
![image](https://user-images.githubusercontent.com/5209820/152118518-2a49bf60-6475-41fe-84d8-3ccfa45a51db.png)

With keyboard:
![image](https://user-images.githubusercontent.com/5209820/152118530-639eb2b1-e8c2-42ba-be40-e1c9feb01ccc.png)

Notice a light brown block under the white background, while the keyboard is up. It shouldn't be there.

For me, the solution is to downgrade to the previous version (7.23.1). 

![image](https://user-images.githubusercontent.com/5209820/152121097-31b72580-81d2-4c1a-8569-07742df66031.png)

swabbass:
Hey, [vimalswaroopj](https://github.com/vimalswaroopj), can you please share a reproducible demo, a minimal one, so we can debug what is really happens in your code? can you, kindly, follow the issue guidelines so we have all the information about how to reproduce your issue? :]

swabbass:
Should be fixed [here](https://github.com/wix/react-native-navigation/pull/7441).

Mitesh-2009:
I am also having this type of issue, in android 8

vimalswaroopj:
Hi @swabbass, I am still getting this issue even after the changes. I have shared a reproducible demo.
https://user-images.githubusercontent.com/42696640/153193258-968072e5-1fd5-46f2-944e-7f5e5cb27e29.mov

vimalswaroopj:
[...]

Issue #7428, opened 2022-01-25
Android windowSoftInputMode is ignored


kabus202:
## 🐛 Bug Report

after the 7.24.0 release the windowSoftInputMode is ignored  for android. For example android:windowSoftInputMode="adjustPan" has no effect at all. 

Last working version: 7.23.1

### Have you read the [Contributing Guidelines on issues](https://wix.github.io/react-native-navigation/docs/meta-contributing/)?

Yes

## To Reproduce

1. In AndroidManifest.xml set android:windowSoftInputMode  to android:windowSoftInputMode="adjustPan" 
1. Add a <TextInput> component
1. run

## Expected behavior

Is that the Keyboard is not pushed up. 

## Actual Behavior

Keyboard getting pushed up

alan-nakagiri:
@yogevbd I'd like to reopen this issue. It looks like the recent release may have reintroduced this issue for Android 30.

Below are some screenshots of what I am seeing. The same code except swapping emulators and RNN versions.

**Android 30, v7.25.4 with AdjustPan**
![adjustPan30-25-4](https://user-images.githubusercontent.com/11300359/157148221-d097e5cd-8e08-4266-95e5-735c83af8624.png)

**Android 30, v7.26.0 with AdjustPan**
![adjustPan30-26](https://user-images.githubusercontent.com/11300359/157148623-4e42f560-f4c2-4fb1-bc17-1bf5ef2ce7c9.png)

**Android 29, v7.26 and with AdjustPan**
![adjustPan29-26](https://user-images.githubusercontent.com/11300359/157148729-8a6bdfdc-685b-4728-a0c7-eea81951561c.png)

**Android 29, v7.25.4 with AdjustPan**
![adjustPan29](https://user-images.githubusercontent.com/11300359/157149879-371197bb-dc3a-4b18-abbc-846bca8386fb.png)

I double checked I had the setups right, but it looks like Android 30 w/ 7.26 does not work but Android 29 with 7.26 does, while Android 30 w/ 7.25.4 does work but Android 29 w/ 7.25.4 does not.

stachu2k:
I have the same problem. My app has android:windowSoftInputMode="adjustPan" and it's ignored. Version 7.26.0.

marf:
Same issue here

lethienhung:
Same issue here

huynqjmango360:
yeb. Android is suck!

kabus202:
[...]

Issue #7427, opened 2022-01-22
[Android 9][7.24.1+] Popping with open keyboard does not uses screenspace of the keyboard


Ainias:
[[## 🐛 Bug Report

I don't know how to describe it very well. If your keyboard is open and you pop to a previous screen, the keyboard hides, but the space the keyboard used is not filled from the screen. View the screencasts

### Have you read the [Contributing Guidelines on issues](https://wix.github.io/react-native-navigation/docs/meta-contributing/)? 

yes

## To Reproduce

1. Create a screen A
2. Open a screen B from A with a text-input
3. Focus text-input such that the keyboard pops up
4. Use the back button inside the topbar to return to screen A
5. The keyboard hides, but the space the keyboard used is not used by the app again
  - If screen A has an input-component which is autofocused, the keyboard stays open and nothing is wrong

## Expected behavior

The screen which is popped to uses the full display space

## Actual Behavior

See step 5 and the screen casts. I removed the `selectable={true}` inside the profile-screen-cast (See https://github.com/wix/react-native-navigation/issues/7426 for referecne) (the person infos inside the casts are dummy data and relates to no real person)

![keyboardHalfScreenLogin](https://user-images.githubusercontent.com/17934715/150654754-2ec2302c-571f-41cf-807a-087cb856d38c.gif)
![keyboardSpaceProfile](https://user-images.githubusercontent.com/17934715/150654758-b47a1af1-de36-47f6-92df-4746cb3ada58.gif)

## Your Environment

- React Native Navigation version: 7.24.1,7.24.2, 7.24.3-hotfix.1, 7.25.1 
- React Native version: 0.66.3
- Platform(s) (iOS, Android, or both?): Android 9 
- Device info (Simulator/Device? OS version? Debug/Release?): Simulator and real device, Android 9 (10+ in simulator was fine, I have no real device with Android 10+), Debug & Release

### Are you willing to resolve this issue by submitting a Pull Request?

 - ✖️  &nbsp; Yes, I have the time, and I know how to start.
 - ✖️  &nbsp; Yes, I have the time, but I don't know how to start. I would need guidance.
 - ✅  &nbsp; No, I don’t have the time and I’m okay to wait for the community / maintainers to resolve this issue.

swabbass:
Hey @Ainias, I tried the described behaviour in our playground and could not reproduce it on Android 9 Emulator, can you please try to reproduce it in our playground, or maybe share a minimal reproducible demo project so we can debug the issue?

[...]

Issue #7402, opened 2021-12-14
[Android] Content shifting up when keyboard appear


N3TC4T:
## 🐛 Bug Report

Content always shifting up when keyboard appears, even with setting `windowSoftInputMode` to `adjustNothing`.

removing `keyboardInsets` is a temporary solution but I'm not sure if it will effect other situations or not.

https://github.com/wix/react-native-navigation/blob/fb6d02897156e4dbbc4ba1e29d76bfce0f320f4b/lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/component/ComponentViewController.java#L146

## To Reproduce

(Write your steps here:)

1. Set `windowSoftInputMode` to `adjustNothing`
2. Focus on TextInput
3. Keyboard pushing content to top

## Expected behavior
Keyboard visibility should NOT effect content visibility when `windowSoftInputMode` set to `adjustNothing`.

## Your Environment

- React Native Navigation version: `7.24.1`
- React Native version: `0.66.1`
- Platform(s): `Android`
- Device info: `Android Device running SDK 30`

swabbass:
We should take  `windowSoftInputMode` options like `adjustNothing` into consideration, you are right, soon it will be fixed.

thuongtv-vn:
> We should take `windowSoftInputMode` options like `adjustNothing` into consideration, you are right, soon it will be fixed.

Hi @swabbass Do you know how to android consider "windowSoftInputMode" normally? You are right, we should consider this option as well.

swabbass:
we might use `getActivity().getWindow().getAttributes().softInputMode;` and take it into consideration, soon we will have a PR for this

alan-nakagiri:
Hello! I just wanted to share that I too have seen an issue with `softInputMode` and RNN `7.25.0`. 

Instead my app is set to `adjustPan` and it does not seem to recognize it, and it is behaving like `adjustResize` is set instead.

PR #7453, opened 2022-02-10
Android < 30 Keyboard insets adjustNothing & adjustPan Fix

swabbass:
# Issue:

Window insets does not apply when keyboard got closed in case of adjustPan, and in case of adjustNothing it will just ignore it and treat it like adjustResize.

The difference between Pan and Resize as pointed [here](https://www.techrepublic.com/article/pro-tip-make-sense-of-pan-vs-resize-in-the-android-manifest/):
adjustResize: The activity’s main window is always resized to make room for the soft keyboard on screen.
adjustPan: The activity’s main window is not resized to make room for the soft keyboard. Rather, the contents of the window are automatically panned so that the current focus is never obscured by the keyboard and users can always see what they are typing. **This is generally less desirable than resizing, because the user may need to close the soft keyboard to get at and interact with obscured parts of the window.**

# Fix:
- Due to the nature of adjustPan, we need to wait for the layout to be updated after the keyboard closed, and then apply insets to determine the final state.
- Take adjustNothing into consideration and don't adjust the insets.

Adjust Nothing:

https://user-images.githubusercontent.com/7227793/153453913-7af3fcb7-87f6-4cfa-aaf8-e2f32b023c61.mov

Adjust Pan:

https://user-images.githubusercontent.com/7227793/153453981-73ccf0c9-e42e-41e3-bede-8f561f9ba53a.mov

AdjustPan old Androids (the white space fixed in master, which is not included in the recorded demo)

https://user-images.githubusercontent.com/7227793/153454164-300c115b-9e7a-407c-b612-c0ca8ac158ea.mov

**NOTE:
As checked, if flags are set to adjustResize, then there is no issue at all across all versions.**

Closes: #7433.
Closes: #7427.
Closes: #7428.
Closes: #7402.

### DIFF
--- a/lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/component/ComponentViewController.java
+++ b/lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/component/ComponentViewController.java
@@ -4,6 +4,7 @@ import android.app.Activity;
 import android.content.res.Configuration;
 import android.view.View;
 import android.view.ViewGroup;
+import android.view.WindowManager;
 
 import com.reactnativenavigation.options.OverlayAttachOptions;
 import com.reactnativenavigation.viewcontrollers.viewcontroller.ScrollEventListener;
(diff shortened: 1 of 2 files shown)


kevinwaelkens:
I tried out your PR and it fixes our problem. However I'm still facing [the issue](https://github.com/wix/react-native-navigation/issues/7455) for which I created an issue on Github.

yogevbd (approved):

kabus202:
still have the same problem on 7.29.0 @yogevbd. Is the fix on this release?

Issue #7463, opened 2022-02-17
Build Error with Android Gradle Plugin 7.1.1: It is too late to set sourceCompatibility


kurtiscc:
## 🐛 Bug Report

The `app/build.gradle` file is setting `sourceCompatibilty` in an `afterEvaluate` block unnecessarily because it is already being set during evaluation and now it is causing build errors with the latest Android Gradle Plugin (7.1.1):

```groovy
allprojects { p ->
    p.afterEvaluate {
        p.android.compileOptions.sourceCompatibility JavaVersion.VERSION_1_8
        p.android.compileOptions.targetCompatibility JavaVersion.VERSION_1_8
    }
    p.repositories {
        maven { url "https://jitpack.io" }
    }
}
```

### Have you read the [Contributing Guidelines on issues](https://wix.github.io/react-native-navigation/docs/meta-contributing/)? 

Yes

## To Reproduce

1. In your own React Native Android app, bump the `build.gradle` plugin version to `7.1.1`: 
```groovy
classpath('com.android.tools.build:gradle:7.1.1')
```
2. Android Studio should then automatically prompt you to update your gradle-wrapper version to `7.2`:
```properties
distributionUrl=https\://services.gradle.org/distributions/gradle-7.2-bin.zip
```
3. Sync Gradle

## Expected behavior

Gradle syncs and the apk builds with no errors.

## Actual Behavior

Syncing Gradle results in the following error:
```
A problem occurred configuring project ':react-native-navigation'.
> It is too late to set sourceCompatibility
  It has already been read to configure this project.
  Consider either moving this call to be during evaluation,
  or using the variant API.
```

## Your Environment

- React Native Navigation version: `7.25.1`
- React Native version: `0.66.3`
- Platform(s) (iOS, Android, or both?): Android
- Device info (Simulator/Device? OS version? Debug/Release?): N/A (all of the above)

## Reproducible Demo

Since this is a build error after a simple Android Gradle Plugin version bump, I don't see a need for a demo. See the error message above to see the behavior.

### Are you willing to resolve this issue by submitting a Pull Request?

 - ✅  &nbsp; Yes, I have the time, and I know how to start.
 - ✖️  &nbsp; Yes, I have the time, but I don't know how to start. I would need guidance.
 - ✖️  &nbsp; No, I don’t have the time and I’m okay to wait for the community / maintainers to resolve this issue.

juneyoungmorgan:
[...]

PR #7464, opened 2022-02-17
Fix Build Error After Upgrading To Android Gradle Plugin 7.1.1

kurtiscc:
See [Issue 7463](https://github.com/wix/react-native-navigation/issues/7463)


SYoder1:
We have running into the same issue. Can we please get this merged? 🙏

PR #7552, opened 2022-07-06
Fix dark mode buttons support on iOS

yogevbd:
Buttons with icon dark mode support didn't work on iOS.


mattmcdonald-uk:
@yogevbd This change introduced a new bug documented here https://github.com/wix/react-native-navigation/issues/7566

Issue #7560, opened 2022-07-13
ModalDismissedEvent type should not have componentName field since 2020


NullPainter2:
## 🐛 Bug Report

**ModalDismissedEvent** type should not have **componentName** field. It is apparently not filled since 2020 https://github.com/wix/react-native-navigation/issues/6605#issuecomment-701172387 .

## Have you read the [Contributing Guidelines on issues](https://wix.github.io/react-native-navigation/docs/meta-contributing/)?

No. :)

## Are you willing to resolve this issue by submitting a Pull Request?

Yes. :)

yedidyak:
Closing this as part of a maintainer cleanup of open issues older than one year.

Decision: Closed: stale/non-critical; no evidence it is a critical current-version RNN issue.

If this still reproduces on the latest React Native Navigation with a minimal reproduction, please open a new issue with current RN/RNN versions and repro steps.

Issue #6605, opened 2020-09-23
event.componentName no longer provided from registerModalDismissedListener


hugo-chq:
### Issue Description

Previously registerModalDismissedListener would provide the componentName , but it's now unavailable / undefined.  

registerModalDismissedListener was being used to recognise when returning to the previous screen underneath.  

Is this change intentional? And if so is there a recommended way to get the componentName from the componentId?  
 Thanks

### Steps to Reproduce / Code Snippets / Screenshots

```
// logs 'undefined'
Navigation.events().registerModalDismissedListener((event) => {
  console.log(event.componentName);
})
```

---
### Environment
* React Native Navigation version: 7.0.0
* React Native version: 0.63.2
* Platform: tested on iOS
* Device info: iOS Simulator iPhone 11 iOS 13.7 Debug

danilobuerger:
Whats the content of `event` for you? (`console.log(event)`)

jinshin1013:
It seems like `V7` removed the emitting `componentName` on `ModalDismissed` event [PR here](https://github.com/wix/react-native-navigation/pull/6422/files#diff-e42324e91a1b48834d165e91fc3c9a25R126).

@yogevbd Was this an intentional change? Android is still emitting `componentName`.

hugo-chq:
Hi @danilobuerger sorry for the delay, the event logs `{"componentId": "Stack10", "modalsDismissed": 1}`

guyca:
Hey guys, indeed we've removed `componentName` from the `modalDismissed` event. It should have been removed from Android as well, this is my mistake.

The reason for removing the parameter from the event is because it's not reliable. Modals can contain any type of layout, including a `SideMenu` or a `SplitView`. What's the current component of a SideMenu? Is the the center layout? or the currently opened drawer?

@hugo-chq Can you please describe the use case for which you need the current component?

hugo-chq:
@guyca Hi , yes the current use case is keeping track of the latest active / visible screen.
we use `registerComponentDidAppearListener` to detect the current visible screen, and then 
 'registerModalDismissedListener' to detect returning to a previous screen, as `registerComponentDidAppearListener` doesn't fire when dismissing a modal.

For example in an app with a settings modal.

1. tap `Home Screen` , 
registerComponentDidAppearListener fires, 
state.visibleScreen = componentName ("Home")

2. tap `Settings Modal` icon, 
registerComponentDidAppearListener fires, 
[...]

PR #7561, opened 2022-07-13
#7560 removed componentName from ModalDismissedEvent

NullPainter2:
Solves #7560.

**componentName** is not filled since 2020 https://github.com/wix/react-native-navigation/issues/6605#issuecomment-701172387 ...


lib/src/interfaces/ComponentEvents.ts
```diff
@@ -26,7 +26,6 @@ export interface NavigationButtonPressedEvent extends ComponentEvent {
 }
 
 export interface ModalDismissedEvent extends ComponentEvent {
-  componentName: string;
```
NullPainter2, line 29: Check if .componentName fields in other events are filled or not ...

PR #7576, opened 2022-08-04
Support RN new architecture

yogevbd:


ReactNativeNavigation.podspec
```diff
@@ -17,18 +17,61 @@ Pod::Spec.new do |s|
...
     s.exclude_files       = "lib/ios/ReactNativeNavigationTests/**/*.*", "lib/ios/OCMock/**/*.*"
   end
   
-  s.subspec 'Fabric' do |ss|
-    ss.xcconfig = { "HEADER_SEARCH_PATHS" => "\"$(PODS_ROOT)/boost-for-react-native\" \"$(PODS_ROOT)/RCT-Folly\"",
-                    "OTHER_CFLAGS" => "$(inherited) -DRN_FABRIC_ENABLED -DFOLLY_NO_CONFIG -DFOLLY_MOBILE=1 -DFOLLY_USE_LIBCPP=1" }
-    ss.dependency 'React-RCTFabric'
-    ss.dependency 'React-Fabric'
-    ss.dependency 'RCT-Folly/Fabric'
+  reactJson = JSON.parse(File.read(File.join(__dir__, "node_modules", "react-native", "package.json")))
```
gaguirre, line 24: I think this could fail in a monorepo, where `react-native` dependency could be located in other folder than _local `node_modules`_.
Anyway, it would be very unlikely since hoisting `react-native` bring too many problems.

Just wondering if there is an easy way to provide a fallback in case that happens (e.g. specifying the RN version manually).

What do you think @yogevbd?

gaguirre (commented):
I focused on the lib files, not the playground changes since I guess are the ones required for the new arch.

I'm wondering how did you come with all the changes for the `.podspec` files. 
Are those changes suggested by React Native for native libs maintainers or something? Because I'd like to know more about new architecture.

ReactNativeNavigation.podspec
enahum: out of curiosity is this comment holding up the PR? perhaps the fallback can be about setting up an env var?

Just wondering was pending to push this forward ;)

aliraza-noon:
out of curiosity are we going to support an sync api with this change ?

xgenem:
Will this ever be supported? The new Arch is out for a while now.

enahum:
Any progress on this?

SudoPlz:
@yogevbd I think we should rebase this PR to https://github.com/wix/react-native-navigation/pull/7717

And then merge my Android fabric support PR https://github.com/wix/react-native-navigation/pull/7720 into [yours](https://github.com/wix/react-native-navigation/pull/7576)

Any thoughts?

SudoPlz:
@yogevbd I see tests still failing here. What's remaining to get this one merged? I remember you made progress w/ tests and they were passing, is it just a matter of pushing your new code?

yogevbd:
Closing in favor of #7744

Issue #7584, opened 2022-08-18
[IOS] BottomTabItem title and icon are GONE after mergeOptions


thuongtv-vn:
## 🐛 Bug Report

I have a simple UI with 3 bottomTabs. When i pushed to other views and using mergeOptions api to adjust the bottomTab at this view the issue happen. BottomTabItem title and icon are GONE after mergeOptions. It only happen on iOS.

### Have you read the [Contributing Guidelines on issues](https://wix.github.io/react-native-navigation/docs/meta-contributing/)?

YES

## To Reproduce

1. First tab has a list item 
2. Pressed on any item then the detail item view opened.
3. Change bottomTab items color by:

`mergeOptions(componentId, {
      topBar: {
        title: {
          text: title
        }
      },
      bottomTab: {
        ...theme.bottomTab,
        textColor: 'red'
      }
    })`

## Expected behavior

The bottom tab items change color only when I only want to change its color.

## Actual Behavior

The icon and title at first tab are GONE. other tabs the title is changed the color properly.

## Your Environment

- React Native Navigation version: 7.18.1
- React Native version: 0.63.4
- Platform(s) (iOS, Android, or both?): iOS
- Device info (Simulator/Device? OS version? Debug/Release?): All

### Are you willing to resolve this issue by submitting a Pull Request?

 - ✅  &nbsp; Yes, I have the time, and I know how to start.
 - ✖️  &nbsp; Yes, I have the time, but I don't know how to start. I would need guidance.
 - ✖️  &nbsp; No, I don’t have the time and I’m okay to wait for the community / maintainers to resolve this issue.

Ainias:
We eperience the same bug right now with `react-native-navigation: 7.37.0-hotfix.1` and `react-natice: 0.71.7`

yedidyak:
Closing this as part of a maintainer cleanup of open issues older than one year.

Decision: Closed: stale/non-critical; no evidence it is a critical current-version RNN issue.

If this still reproduces on the latest React Native Navigation with a minimal reproduction, please open a new issue with current RN/RNN versions and repro steps.

PR #7585, opened 2022-08-18
fix(bottomtabs): fix bottom tabbar items missing title and icon after mergeOptions

thuongtv-vn:
resolved the issue: https://github.com/wix/react-native-navigation/issues/7584


7chenys:
Hello, has this problem been solved? I have the same problem

PR #7586, opened 2022-08-21
Expose when setNavigationRoot is called on RootPresenter

zzorba:
This change exposes when `setNavigationRoot` is called by the javascript application on the RootPresenter class. 

This is needed to implement the new Android 12 style splash screen (discussion here: https://github.com/wix/react-native-navigation/discussions/7363), so that the screen can persist until the application is ready to show real content.

### DIFF
--- a/lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/viewcontroller/RootPresenter.java
+++ b/lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/viewcontroller/RootPresenter.java
@@ -19,6 +19,7 @@ public class RootPresenter {
     private final RootAnimator animator;
     private CoordinatorLayout rootLayout;
     private final LayoutDirectionApplier layoutDirectionApplier;
+    private boolean setRootCalled = false;
 
     public void setRootContainer(CoordinatorLayout rootLayout) {
         this.rootLayout = rootLayout;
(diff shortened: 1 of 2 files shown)


zzorba:
Not stale, still very relevant for anyone wanting to implement the new splash screen behavior from Android 12

zzorba:
Just wondering if this might be able to get in soon.  It's needed to support the new Android 11 style splash screen, which has been out for some time now.

marcshilling:
Throwing my vote in here as well. I've applied these changes via patch-package and they are working to allow to me to use the new SplashScreen API on Android.

lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/viewcontroller/RootPresenter.java
```diff
@@ -28,13 +29,18 @@ public RootPresenter() {
...
+    }
+
     @VisibleForTesting
     public RootPresenter(RootAnimator animator, LayoutDirectionApplier layoutDirectionApplier) {
         this.animator = animator;
         this.layoutDirectionApplier = layoutDirectionApplier;
     }
 
     public void setRoot(ViewController appearingRoot, ViewController<?> disappearingRoot, Options defaultOptions, CommandListener listener, ReactInstanceManager reactInstanceManager) {
+        setRootCalled = true;
```
marcshilling, line 43: It would be better to move this into `animateSetRootAndReportSuccess`, right before the `listener.onSuccess` calls. And you'll need 2 lines of it, one in each block.

manoj-mehra-spraxa (approved):

PR #7661, opened 2023-01-05
Improve logging when failing to render components

guyca:
In cases where a component was indeed registered, but inline required incorrectly - RNN failed without throwing an informative error.


doronpr:
we'll ditch this

PR #7665, opened 2023-01-07
Fix issue with bottom-tabs being obscured on cold launch on android 13

zzorba:
Starting with Android 13, it seems like there has been a change in the order of events being fired for layout of the system navigationBars and systemBars.  This results in what appears to be a race condition of sorts, which leads to the bottom tabs not accounting for the OS navigationBar height when the app is initially launched, causing them to render 'behind' the OS controls.
Screen Shot 2023-01-07 at 10 52 59 AM

Upon backgrounding and relaunching the app, or interacting with the TINY bar there, this would be fixed:
Screen Shot 2023-01-07 at 10 53 32 AM

I don't know if this is the optimal way to fix it, but adding this override appears to at least trigger an additional redraw and fixes the issue so that the bar is visible at the start.

Note: in order to see this behavior on an emulator, you need to set the `hw.mainKeys=no` in the config.ini field as well as turning off the gesture navigation system in System settings.

### DIFF
--- a/lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/bottomtabs/BottomTabsController.java
+++ b/lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/bottomtabs/BottomTabsController.java
@@ -5,6 +5,7 @@ import android.app.Activity;
 import android.content.res.Configuration;
 import androidx.core.view.ViewCompat;
 import androidx.core.view.WindowInsetsCompat;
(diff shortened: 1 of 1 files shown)


guyca:
@yogevbd You might want to consider integrating [Insetter](https://github.com/chrisbanes/insetter). It's the only sane way I know of to handle insets on Android.

ninjz:
I see this wasn't merged. Will this be fixed in the near future?

PR #7675, opened 2023-01-25
Fix: Wrong options are set when onConfigurationChanged occurs

BlackCat1397:
## Repro steps
1. Set system theme to light.
2. Create stack root with multiple tabs
3. Set nav bar color to black on first tab.
```
Navigation.mergeOptions({
  navigationBar: {
    backgroundColor: '#000000',
  }
});
```
4. Change app orientation.

**Expected Result:** Nav bar stays black.
**Achieved Result:** Nav bar becomes white.

## Issue cause
As I've figured out during debugging, `mergeOptions` changes only option of controller with specified `componentId`. However, when `Navigator.onConfigurationChanged` is invoked the options of all existing controllers are being applied to presenter, therefore the last created controller's options are finally set.

## Solution
Check for `controller.isViewShown` before triggering it's `onConfigurationChanged`.


BlackCat1397:
@guyca Can you, please, take a look at this PR or assign some of your teammembers to look at it?)

guyca:
@yogevbd Think we should look into this one.

yogevbd (approved):

PR #7691, opened 2023-03-04
feat: support base64 images on bottomsheet

asherLZR:
Currently images for bottomsheet must be from a local resource. One way to support dynamic images such as for a profile image in the bottom tabs is to allow base64 images. This seems to work already for iOS but we can add a little extra logic to support this on Android.


xgenem:
+1 

Do you think you can modify to allow component as well?

PR #7714, opened 2023-05-17
RN Mock: push didAppear to next event loop

doronpr:
This is needed to simulate real app life cycle where componentDidAppear is called after componentDidMount as per https://wix.github.io/react-native-navigation/docs/screen-lifecycle/#mounting


doronpr:
Note for future reference: this is indeed an issue but we decided not to merge due to potential unknown implications (there could be none but needs to investigate).

Niryo:
@yogevbd what do you think about it?

yogevbd (approved):

PR #7717, opened 2023-05-22
Upgrade react-native to 0.71.8

yogevbd:


lib/Mock/mocks/NativeCommandsSender.tsx
```diff
@@ -47,6 +47,7 @@ export class NativeCommandsSender {
 
   pop(commandId: string, componentId: string, _options?: object) {
     return new Promise((resolve) => {
+      if (!LayoutStore.getLayoutById(componentId)) return;
```
SudoPlz, line 50: curious, what's the use-case here? Why does this return silently? (should we consider throwing an error/warning instead?

SudoPlz (commented):
this is shaping up, but we're actually duplicating work here, I also made those changes on another branch. 

That's totally fine though since you picked it up.
Btw, I left 1 comment up there (a question) and reviewed the rest of the PR.
This looks good.

lib/Mock/mocks/NativeCommandsSender.tsx
yogevbd: Fixed 👍

playground/android/app/build.gradle
```diff
@@ -1,19 +1,12 @@
...
-        bundleInBeta: true,
-        enableHermes: true,
-        hermesFlagsDebug:['-Xes6-proxy','-output-source-map'],
-        hermesFlagsRelease:['-output-source-map'],
-        hermesCommand: "../../../node_modules/react-native/sdks/hermesc/%OS-BIN%/hermesc",
-]
-
-apply from: "../../../node_modules/react-native/react.gradle"
+react {
+     root = file("../../../")
```
SudoPlz, line 5: is that all we need?

What about?

```
react = {
    root = file("../../../")
    entryFile = file("index.js")
    cliPath = file("node_modules/react-native/cli.js")
    bundleAssetName = "index.android.bundle"
    bundleInAlpha = true
    bundleInBeta = true
    enableHermes = true
    hermesFlagsDebug = ['-Xes6-proxy','-output-source-map']
    hermesFlagsRelease = ['-output-source-map']
    hermesCommand = file("../../../node_modules/react-native/sdks/hermesc/%OS-BIN%/hermesc")
}
```
Do we not care about the other values that used to be there?

playground/android/app/build.gradle
```diff
@@ -1,19 +1,12 @@
...
+}
+
+
+def jscFlavor = 'org.webkit:android-jsc:+'
 
 android {
     compileSdkVersion rootProject.ext.get("compileSdkVersion")
```
SudoPlz, line 12: consider adding a namespace here, like so:

`namespace "com.reactnativenavigation.playgroun` and remove `package="com.reactnativenavigation.playground"` from `AndroidManifest.xml` for the package name.

playground/android/app/build.gradle
```diff
@@ -57,20 +50,20 @@ dependencies {
     implementation 'com.google.android.material:material:1.4.0'
```
SudoPlz, line 50: I think we can safely delete the line above the one I highlighted:

`implementation fileTree(dir: 'libs', include: ['*.jar'])`

playground/android/build.gradle
```diff
@@ -6,9 +6,9 @@ buildscript {
         RNNKotlinVersion = kotlinVersion
         detoxKotlinVersion = kotlinVersion
         compileSdkVersion = 33
-        buildToolsVersion = "31.0.0"
+        buildToolsVersion = "33.0.0"
         minSdkVersion = 21
-        targetSdkVersion = 31
+        targetSdkVersion = 33
     }
```
SudoPlz, line 12: We may want to specify an ndk version here:

```suggestion
        // We use NDK 23 which has both M1 support and is the side-by-side NDK version from AGP.
        ndkVersion = "23.1.7779620"
    }
```

playground/android/build.gradle
```diff
@@ -18,26 +18,18 @@ buildscript {
 
     dependencies {
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlinVersion"
-        classpath 'com.android.tools.build:gradle:7.1.1'
+        classpath("com.facebook.react:react-native-gradle-plugin")
+        classpath 'com.android.tools.build:gradle:7.3.1'
         classpath("de.undercouch:gradle-download-task:5.0.1")
```
SudoPlz, line 23: we probably no longer need this `classpath("de.undercouch:gradle-download-task:5.0.1")`

playground/android/app/build.gradle
yogevbd: I think those are RN's defaults

SudoPlz (approved):
LGTM

PR #7730, opened 2023-06-04
Fix bottomTabs visibility on setRoot with second tab selected

yogevbd:
Calling `setRoot` with two `bottomTabs` children and the second one is selected and `bottomTabs.visible: false` caused in wrong visibility.


PrudhviBanki:
when tap to first tab data not refreshing

PR #7732, opened 2023-06-05
pass componentName prop to screens

yogevbd:
Currently we inject only the `componentId` to screens, this PR add `componentName` as well.

### DIFF
--- a/lib/src/interfaces/NavigationComponentProps.ts
+++ b/lib/src/interfaces/NavigationComponentProps.ts
@@ -1,3 +1,4 @@
 export interface NavigationComponentProps {
   componentId: string;
+  componentName: string;
 }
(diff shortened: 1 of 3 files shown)


ethanshar (approved):

lib/src/interfaces/NavigationComponentProps.ts
```diff
@@ -1,3 +1,4 @@
 export interface NavigationComponentProps {
   componentId: string;
+  componentName: string;
```
ejain, line 3: Shouldn't this be `componentName?: string`, to avoid breakage of existing code?

yogevbd: I will revert this for now, It doesn't make a lot of sense to add it as optional as it always exist

PR #7765, opened 2023-08-09
Update RNNSplashScreen.m

JimTeva:
In 2020, Apple has announced the end of LaunchImage asset for all apps in AppStore:

“Note: Don’t use a static image for your launch screen. Static images have been deprecated and all App Store apps must use an Xcode storyboard to provide an app’s launch screen by June 30, 2020. Learn more here.”

So we must use storyboard. But if we create a LaunchScreen.storyboard, the app will crash because it is looking for a storyboard named "LaunchScreen.storyboard" while it should look for "LaunchScreen".

So to fix it we remove the extension to the name file and it works.


pmogollons:
This fixes an issue with the latest rn version 0.72.4 crashing after showing the splashscreen. Thanks

yogevbd (approved):

PR #7774, opened 2023-08-22
Run detox on iPhone 14 Pro Max

yogevbd:


asafkorem:
@yogevbd detox version was updated with a fix. Note that the current failures seem to be due to screen size changes

PR #7781, opened 2023-09-05
Fix buttons layout

yogevbd:


markhomoki:
I had the same issue after upgrading to the latest RNN. Implementing this solution fixed my problem, when a wider button was off screen even though it had plenty of space next to it. I also have a center logo. See the screenshot with and without this fix.

image

PR #7704, opened 2023-04-27
fix: measure overlay touch based on child dimensions

asherLZR:
Attempts to resolve https://github.com/wix/react-native-navigation/issues/7674 in a backward compatible and "cast-safe" way based on the fix from @likesebiii. In RN71 there are additional view groups which prevent `interceptTouchOutside` from getting the correct hitbox.

The build config branch doesn't feel ideal so any suggestions on how to improve are welcome.

**RN 0.71**
Screenshot 2023-04-27 at 20 16 00

**RN 0.70**
Screenshot 2023-04-27 at 20 02 57

### Test Plan
Use this as a patch in both 0.70 and 0.71 apps, checking `interceptTouchOutside` true and false on overlay.

Issue #7674, opened 2023-01-24
Overlay Screen with `interceptTouchOutside` doesn't work correctly on Android


joetannenbaum:
### What happened?

On a fresh React Native installation (0.71 and 0.71.1 thus far), an overlay screen on Android with `interceptTouchOutside: false` isn't behaving correctly, you can't interact with the screen underneath it. It still works correctly on iOS.

### What was the expected behaviour?

With `interceptTouchOutside: false` specified, you should be able to interact with the content beneath the overlay.

### Was it tested on latest react-native-navigation?

- [X] I have tested this issue on the latest react-native-navigation release and it still reproduces.

### Help us reproduce this issue!

1. Clone https://github.com/joetannenbaum/sample-react-native-navigation-overlay-bug
2. Yarn
3. npx react-native run-android

This screen is where the `interceptTouchOutside` option is specified:

https://github.com/joetannenbaum/sample-react-native-navigation-overlay-bug/blob/main/OverlayScreen.tsx

### In what environment did this happen?

React Native Navigation version: 7.31.1
React Native version: 0.71.1
Has Fabric (React Native's new rendering system) enabled: (yes/no) no
Node version: 19.4.0
Device model: Pixel 3XL and Pixel 6 Pro
Android version: 12 and 13

Omelyan:
Any updates on this..?

Omelyan:
ok, I'm not familiar with Kotlin and Java at all, but here's how I solved it:

/lib/android/app/src/main/java/com/reactnativenavigation/utils/MotionEvent.kt

```javascript
+import android.view.ViewGroup

fun MotionEvent.coordinatesInsideView(view: View?): Boolean {
     view ?: return false
-    view.getHitRect(hitRect)
+    val content = ((view as ViewGroup).getChildAt(0) as ViewGroup).getChildAt(0) as ViewGroup // it looks like sht but I don't know Kotling at all
+    content.getHitRect(hitRect)
     return hitRect.contains(x.toInt(), y.toInt())
}

```
so without even knowing RN in nitty gritty details, it seems that the latest version somehow wraps the root view in containers and we should get overflow's child dimensions instead of the parent's one.

anglingiq:
Having the same issue. @Omelyan solution worked for me!

LYevhen:
Having same issue, for some reason after patch apply created by @likesebiii or direct module modification with @Omelyan proposal did not fix an issue, instead of click was not intercepted, now app just crash.

[...]

PR #7786, opened 2023-09-14
[Android] Wrong user's tap area in overlay

v-korotkov:
Hi!

**I use:
    "react-native": "0.72.4",
    "react-native-navigation": "^7.37.0",**

I faced an annoying problem with `overlays` on `Android`. I show an overlay with a not full-screen view with `interceptTouchOutside: false`. For example, this view is `{x: 0, y: 500, width: 100, height: 100}`

When the user tries to tap the overlay the app receives a tap through the overlay (under it). Also, when the user tries to tap an area `{x: 0, y: 0, width: 100, height: 100}` through overlay (using `interceptTouchOutside: false`) it is blocked for taps.

It works well on iOS. Also, it worked well on older versions of react-native and react-native-navigation.

I found a problem in `react-native-navigation/lib/android/app/src/main/java/com/reactnativenavigation/utils/MotionEvent.kt`

When we use the old version of the code our not full-screen overlay view performs `getHitRect` in the full-screen parent, so it works correctly.

```
view ?: return false
view.getHitRect(hitRect)
return hitRect.contains(x.toInt(), y.toInt())
```

However, when we started to use the new version of the code, we had an issue. It happens cause we use a child of this view from `getChildAt` instead of the original view for the `getHitRect` method.

```
val viewGroup = (view as? ViewGroup)?.getChildAt(0) as? ViewGroup ?: view
viewGroup?.getHitRect(hitRect)
return hitRect.contains(x.toInt(), y.toInt())
```
[Android View getHitRect](https://developer.android.com/reference/android/view/View#getHitRect(android.graphics.Rect))
The documentation says that `getHitRect - Hit rectangle in parent's coordinates`.

Therefore, when we use the original view for `getHitRect`, it calculates the hit for his full-screen parent, but when we use a view's child, it calculates the hit for the original view as a parent.

So, when the user taps on `{x: 50, y: 550}` coordinates on the screen, it will be just `{x:50, y:50}` in the not full-screen parent, and the blocking tap area for overlay will be translated to the top of the screen in the wrong place.

Additional changes: typo in `REACT_NATVE_VERSION_MINOR`

I prepared changes to fix this issue. Let's review and apply ones.

### DIFF
--- a/lib/android/app/build.gradle
+++ b/lib/android/app/build.gradle
@@ -73,43 +73,43 @@ android {
     productFlavors {
         reactNative51 {
             dimension "RNN.reactNativeVersion"
-            buildConfigField("int", "REACT_NATVE_VERSION_MINOR", "51")
+            buildConfigField("int", "REACT_NATIVE_VERSION_MINOR", "51")
         }
         reactNative55 {
             dimension "RNN.reactNativeVersion"
-            buildConfigField("int", "REACT_NATVE_VERSION_MINOR", "55")
+            buildConfigField("int", "REACT_NATIVE_VERSION_MINOR", "55")
         }
         reactNative56 {
             dimension "RNN.reactNativeVersion"
-            buildConfigField("int", "REACT_NATVE_VERSION_MINOR", "56")
+            buildConfigField("int", "REACT_NATIVE_VERSION_MINOR", "56")
         }
         reactNative57 {
             dimension "RNN.reactNativeVersion"
-            buildConfigField("int", "REACT_NATVE_VERSION_MINOR", "57")
+            buildConfigField("int", "REACT_NATIVE_VERSION_MINOR", "57")
         }
         reactNative57_5 {
             dimension "RNN.reactNativeVersion"
-            buildConfigField("int", "REACT_NATVE_VERSION_MINOR", "57")
+            buildConfigField("int", "REACT_NATIVE_VERSION_MINOR", "57")
         }
         reactNative60 {
             dimension "RNN.reactNativeVersion"
-            buildConfigField("int", "REACT_NATVE_VERSION_MINOR", "60")
+            buildConfigField("int", "REACT_NATIVE_VERSION_MINOR", "60")
         }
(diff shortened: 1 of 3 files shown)


lfaz:
Had similiar issues on android, this fixed!

v-korotkov:
> Had similiar issues on android, this fixed! EDIT: its not consistent, can't recommand as an actual fix!

Could you please add the specific versions of "react-native" and "react-native-navigation" where it was fixed and works fine now? I have a stable reproduction of this issue since updating to the latest versions of these modules ("react-native": "0.72.4" and "react-native-navigation": "^7.37.0"). Therefore I have to use a workaround in my code to avoid this issue in the "react-native-navigation" module.

I know that the current code has already had changes that trying to fix a similar problem on Android ([problem](https://github.com/wix/react-native-navigation/issues/7674), [PR](https://github.com/wix/react-native-navigation/pull/7704), [commit](https://github.com/wix/react-native-navigation/commit/93453f664f14672865e20e32b6bfd2e0e5a25a04)), but it is obviously the wrong way to fix it because it adds a new issue to the latest versions.

If we want to resolve the old problem too we should consider updating to the latest version or suggest a more specific way to detect the case when we should use child view instead of the original view instead of just always using it by default.

jyoti-63:
Is third commit [46349889ef510a2de2d84f956b366e63aac20533]([REACT_NATIVE_VERSION_MINOR typo](https://github.com/wix/react-native-navigation/pull/7786/commits/46349889ef510a2de2d84f956b366e63aac20533)) required to in this fix ?

v-korotkov:
> Is third commit [[4634988](https://github.com/wix/react-native-navigation/commit/46349889ef510a2de2d84f956b366e63aac20533)]([REACT_NATIVE_VERSION_MINOR typo](https://github.com/wix/react-native-navigation/pull/7786/commits/46349889ef510a2de2d84f956b366e63aac20533)) required in this fix ?

No, this is not needed for the problem fix, it was made just as an additional typo fix.

PR #7788, opened 2023-09-18
Make bottom tabs more visually correct

Niryo:
I have made some changes it order to make the bottom tabs more correct when rendering visually (by using react-native-web).
before: 
image

after:
image


Niryo:
I added also icon support now:
image

yogevbd (approved):

PR #7809, opened 2023-11-06
feat(constants): add navigationBarHeight to getConstants()

oblador:
This value is missing in the current set of constants, but only relevant for android. It's useful to know if you have a transparent/translucent navbar and what to know what inset to use etc.


oblador:
Let’s get a review instead of closing thx

PR #7792, opened 2023-10-05
Fix runOnMainThread to run at front of queue

yogevbd:
This fixes a crash happening in the Wix app, look like there's a race condition where the activity is destroyed and navigator.onHostPause() gets called too late and should always be called at front of queue. Most of times this shouldn't be delayed as it is called on the main thread.

This is the crash:
```
Location
NavigationModule.java line 213 in com.reactnativenavigation.react.NavigationModule.navigator
Exception
java.lang.NullPointerException
Message
Attempt to invoke virtual method 'com.reactnativenavigation.viewcontrollers.navigator.Navigator com.reactnativenavigation.NavigationActivity.getNavigator()' on a null object reference
```

Closes #7769

PR #7810, opened 2023-11-09
[Android] fix: Check null in order to avoid crash randomly and unexpectedly

moonjava2005:
From our app crashlytic, we saw a lot of crashes like this
`Attempt to invoke virtual method 'com.reactnativenavigation.viewcontrollers.navigator.Navigator com.reactnativenavigation.NavigationActivity.getNavigator()' on a null object reference`
I cannot figure out why navigator is NULL. However, we should check null before calling method in order to avoid unexpected crash. I think the React host is null in moment, and then when users click again, the host will be available

### DIFF
--- a/lib/android/app/src/main/java/com/reactnativenavigation/react/NavigationModule.java
+++ b/lib/android/app/src/main/java/com/reactnativenavigation/react/NavigationModule.java
@@ -67,14 +67,18 @@ public class NavigationModule extends ReactContextBaseJavaModule {
             @Override
             public void onHostResume() {
                 eventEmitter = new EventEmitter(reactContext);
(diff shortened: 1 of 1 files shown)


yogevbd:
Does this crash reproduce in the latest RNN version? because it should be fixed in [this PR](#7792)

moonjava2005:
> Does this crash reproduce with the latest RNN version? because it should be fixed in [this PR](#7792)

I got these crashes with RNN `7.37.1` And the PR you have just attached to your reply is this PR :D

MatthewPattell:
I can confirm the same, on the latest version of the package.

moonjava2005:
@yogevbd Could your review this PR?

Issue #7593, opened 2022-09-02
java.lang.NullPointerException: Attempt to invoke virtual method 'h.g.k.g.f h.g.b.d()' on a null object reference


DenisPostnikov:
Hey there,

here's the stacktrace of the issue:

## StackTrace

Fatal Exception: java.lang.NullPointerException: Attempt to invoke virtual method 'h.g.k.g.f h.g.b.d()' on a null object reference
       at com.reactnativenavigation.react.NavigationModule.navigator(NavigationModule.java:4)
       at com.reactnativenavigation.react.NavigationModule.access$100(NavigationModule.java)
       at com.reactnativenavigation.react.NavigationModule$1.lambda$onHostPause$0(NavigationModule.java:2)
       at com.reactnativenavigation.react.-$$Lambda$NavigationModule$1$YAogg-YWvGlqmTQTIc1m0ewJv3s.run(-.java:2)
       at android.os.Handler.handleCallback(Handler.java:883)
       at android.os.Handler.dispatchMessage(Handler.java:100)
       at android.os.Looper.loop(Looper.java:237)
       at android.app.ActivityThread.main(ActivityThread.java:8167)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:496)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1100)

## Environment

- React Native Navigation version: "^7.29.0"
- React Native version:"0.66.3"
- Platform(s) (iOS, Android, or both?): Android
- Device info (Simulator/Device? OS version? Debug/Release?): samsung Galaxy S9, Release

mcapocaccia:
Many crashes here too, same RNN and RN versions.

Crash is in NavigationModule.java:213

    private Navigator navigator() {
        return activity().getNavigator();
    }

This is the context:
- app goes to background
- activity gets destroyed
- getCurrentActivity returns null, so "activity()" above is null
- crash

marf:
Same problem here.

mcapocaccia:
We patched with the diff below in postinstall and the crash disappeared completely.
@guyca @yogevbd you could consider patching 7.30.x

[react-native-navigation+7.29.0.patch.zip](https://github.com/wix/react-native-navigation/files/9557913/react-native-navigation%2B7.29.0.patch.zip)

yogevbd:
Will add this fix in the next release, thanks!

[...]

PR #7812, opened 2023-11-13
Fix onHostPause crash on Android

yogevbd:
This fixes a crash happening in the Wix app, look like there's a race condition where the activity is destroyed and navigator.onHostPause() gets called too late and should always be called at front of queue. Most of times this shouldn't be delayed as it is called on the main thread.

This is the crash:
```
Location
NavigationModule.java line 213 in com.reactnativenavigation.react.NavigationModule.navigator
Exception
java.lang.NullPointerException
Message
Attempt to invoke virtual method 'com.reactnativenavigation.viewcontrollers.navigator.Navigator com.reactnativenavigation.NavigationActivity.getNavigator()' on a null object reference
```


manuhook:
Hi @yogevbd 

Can we do the same for **onHostResume** please ?

Please see this patch : https://github.com/wix/react-native-navigation/issues/7593#issuecomment-1406987778

Thank you,

PR #7815, opened 2023-11-20
Fix NullPointerException in NavigationModule

Wiingaard:
We're seeing a crash that affects ~1% of our user base per month, that's caused by:
```
NullPointerException
Attempt to invoke virtual method 'boolean android.app.Activity.isFinishing()' on a null object reference
```

After looking into the code, we realised that that the null check and `.isFinishing()`  are not referencing the same activity.

Here's a screenshot with stack trace from Sentry:
Screenshot 2023-11-20 at 14 38 24

### DIFF
--- a/lib/android/app/src/main/java/com/reactnativenavigation/react/NavigationModule.java
+++ b/lib/android/app/src/main/java/com/reactnativenavigation/react/NavigationModule.java
@@ -223,7 +223,8 @@ public class NavigationModule extends ReactContextBaseJavaModule {
 
     protected void handle(Runnable task) {
         UiThread.post(() -> {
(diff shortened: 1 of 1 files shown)


lib/android/app/src/main/java/com/reactnativenavigation/react/NavigationModule.java
```diff
@@ -223,7 +223,8 @@ private Options parse(@Nullable ReadableMap mergeOptions) {
 
     protected void handle(Runnable task) {
         UiThread.post(() -> {
-            if (getCurrentActivity() != null && !activity().isFinishing()) {
```
yogevbd, line 226: I don't understand how it's a different activity, `activity()` returns `getCurrentActivity()`. wdyt?

PR #7829, opened 2024-01-08
build(playground): patch boost podspec on post-install.

asafkorem:
add post-install patch to use boost.io's mirror due to installation errors from the original source
See https://github.com/boostorg/boost/issues/843

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
   "main": "lib/dist/src/index.js",
   "typings": "lib/dist/src/index.d.ts",
   "scripts": {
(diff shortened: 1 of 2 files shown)


artald (approved):

oblador:
This has been fixed in [0.73.2](https://github.com/facebook/react-native/releases/tag/v0.73.2). Let's revert his one as it causes the package to fail installation.

asafkorem:
Thanks @oblador, reverting: https://github.com/wix/react-native-navigation/pull/7850

PR #7844, opened 2024-02-11
Support RN 73

gosha212:


.buildkite/pipeline.yml
d4vidi, line 1: @igorgn this is not cool... we need to bring back the config to our end, as with Detox

d4vidi (approved):
LGTM
Let's break in mind that the next release will require everyone to switch to Java17 - we could link up a setup guide etc.

PR #7127, opened 2021-05-20
Add topBar button animations on iOS

yogevbd:
Adds the ability to use the system's default button animations.

```js
options: {
  topBar: {
    animateLeftButtons: true, // false by default for backward consistency
    animatedRightButtons: true // false by default for backward consistency
  }
}
```

PR #7848, opened 2024-02-15
Android: Fix ordering of top-bar left-hand button layout

### LINKED ISSUE #7847: Android - Left buttons are not in order
### What happened?

Android - Left buttons are in opposite order then IOS, or the order we send them in.
In our code we send navigation options the next way:
```
export const getTopBarLeftButtons = () => {
  const backButton = {
    id: NAV_BUTTON_IDS.BACK,
    icon: Assets.icons.navigation.back,
  };

  const infoButton = {
    id: NAV_BUTTON_IDS.INFO,
    icon: Assets.icons.general.questionmarkOutline,
  };

  return shouldShowViewCTAInsideCustomizer() ? [backButton, infoButton] : [infoButton];
};
```
But the result is different on IOS and Android, adding screenshots as an example:
   

Buttons must in same order in IOS and android.

- [X] I have tested this issue on the latest react-native-navigation release and it still reproduces.

https://github.com/wix-private/wix-one-app-member-view/blob/a382ca73e2d5deca75af70ea41c8d054212bd894/packages/wix-one-app-member-view/src/screens/Editor/index.tsx#L76-L77

React Native Navigation version: Latest
React Native version: 0.71
Has Fabric (React Native's new rendering system) enabled: IDK
Node version: 16.20
Device model: All
Android version: 10

d4vidi:
Fixes #7847 (Introduced in: #7127 - https://github.com/wix/react-native-navigation/commit/a475b2dc74e887d9cf1cd197e4b7451f8d4db61e#diff-675155ea574532072e1490bebacc2f3d6bd26523ae2fcd60981d7e52cb27c160R74).

Fixes (custom) reversed ordering logic of left-hand buttons layout.


lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/TitleAndButtonsContainer.kt
```diff
@@ -71,7 +71,7 @@ class TitleAndButtonsContainer(context: Context) : ViewGroup(context) {
         component?.layoutDirection = layoutDirection
         titleSubTitleBar.layoutDirection = layoutDirection
         rightButtonBar.layoutDirection = layoutDirection
-        leftButtonBar.layoutDirection = if(isRTL()) View.LAYOUT_DIRECTION_LTR else View.LAYOUT_DIRECTION_RTL
+        leftButtonBar.layoutDirection = if (isRTL()) View.LAYOUT_DIRECTION_RTL else layoutDirection
```
gosha212, line 74: Why not just setting the direction?

gosha212 (commented):
Great test coverage, but I can't understand the fix =(

lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/TitleAndButtonsContainer.kt
d4vidi: Android has RTL resolution magic that cannot be accounted-for in a simply way. I actually wonder whether the question should be whether we should apply this over the right-hand button bar, as well 🤷🏻

PR #7851, opened 2024-02-20
fix: revert #7844 (React Native 73 support)

gosha212:
Reverts wix/react-native-navigation#7844


asafkorem:
Please merge it only in case of regression (if we had fabric support before this change)

asafkorem:
We found some regressions with this version. Reverting.

asafkorem (approved):

PR #7860, opened 2024-03-18
Disable action-menu clipChildren=false

### LINKED ISSUE #7832: RightButtons jumps on Android
### What happened?

When the icon is used for rightButtons inside topBar, it jumps when the new screen is pushed/popped in the stack.

Jan-22-2024 16-18-56

Note: If a custom component is used for the button, the button doesn't jump.

The RightButton with the icon should not jump during navigation.

Is there any solution for this other than using a custom button? A custom button adds unnecessary complexity.

- [X] I have tested this issue on the latest react-native-navigation release and it still reproduces.

It can be reproduced with Playground app. Just add the icon and use Push navigation.

``` javascript
export default class PushedScreen extends NavigationComponent<Props> {
  static options(): Options {
    return {
      topBar: {
        testID: PUSHED_SCREEN_HEADER,
        title: {
          text: 'Pushed Screen',
        },
        rightButtons: [
          {
            id: 'singleBtn',
            text: 'single',
            testID: TOP_BAR_BTN,
            icon: require('../../img/navicon_edit.png'), // Add icon
          },
        ],
        backButton: {
          testID: BACK_BUTTON,
          enableMenu: false,
        },
      },
    };
  }
...
```

React Native Navigation version: 
React Native version: 
Has Fabric (React Native's new rendering system) enabled: (yes/no) 
Node version: 
Device model: 
Android version:

d4vidi:
## Description

fixes #7832

After spending a fair amount of efforts trying to get a deep understanding the layout and view arrangement flow, i couldn't get to the very bottom of why the buttons menu transitively renders in a broken way, namely like this:

image

instead of:

image

Nevertheless, `clipChildren=false` over the action menu's root (i.e. the `ActionMenuView` class) - introduced [here](https://github.com/wix/react-native-navigation/pull/6918/files) seems to be causing the flicker, while removing it doesn't seem to be introducing any regressions.

#### Further steps

While with `clipChildren=false` the layout is broken transitively, removing it doesn't make all-things-perfect. There is still a very noticeable flicker happening in each stack push/pop, which could probably be avoided. Read below for hints as to why that is.

_Update: Reported in #7862_

---
Some details for future reference:
- Pausing the app using a breakpoint over `ButtonBar#onLayout()` seems to be catching the app in with the visibility bug in-effect, meaning the menu layout phase is two-fold and that the button bar layout happening in between.
- Whether during a push _or_ a pop, the brokenly-rendered menu is laid out with the buttons of the **to-screen**. This hints that the problem comes from a process that contains a buttons-list clear and a repopulation that should be happening atomically but could be invalidating the menu view prematurely.
- The toolbar is a single instance per the generated stack (i.e. of the test app: tapping `Layouts` tab → `Stack` button)

### DIFF
--- a/lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/ButtonBar.kt
+++ b/lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/ButtonBar.kt
@@ -25,13 +25,6 @@ open class ButtonBar internal constructor(context: Context) : Toolbar(context) {
 
     override fun onViewAdded(child: View) {
         super.onViewAdded(child)
-        enableOverflowForReactButtonViews(child)
-    }
-
-    private fun enableOverflowForReactButtonViews(child: View) {
-        if (child is ActionMenuView) {
-            (child as ViewGroup).clipChildren = false
-        }
     }
 
     override fun setLayoutDirection(layoutDirection: Int) {


asafkorem (approved):

asafkorem:
Released with [7.38.4](https://github.com/wix/react-native-navigation/releases/tag/7.38.4)

Issue #7855, opened 2024-02-25
Bottom tabs text is cut off on iPad


d4vidi:
### What happened?

Complete info: `WOARNN-568`

The text in some of the bottom tabs is cut off -

![image](https://github.com/wix/react-native-navigation/assets/9818880/90abe344-3859-490c-a941-752a20e12acb)

### Was it tested on latest react-native-navigation?

- [X] I have tested this issue on the latest react-native-navigation release and it still reproduces.

### In what environment did this happen?

React Native Navigation version: 
React Native version: 
Has Fabric (React Native's new rendering system) enabled: (yes/no) 
Node version: 
Device model: 
iOS version:

phuongwd:
Related to

1. https://github.com/wix/react-native-navigation/issues/7506
2. https://github.com/wix/react-native-navigation/issues/7827

Patch

```
diff --git a/node_modules/react-native-navigation/lib/ios/RNNTabBarItemCreator.m b/node_modules/react-native-navigation/lib/ios/RNNTabBarItemCreator.m
index d313bfc..c22f984 100644
--- a/node_modules/react-native-navigation/lib/ios/RNNTabBarItemCreator.m
+++ b/node_modules/react-native-navigation/lib/ios/RNNTabBarItemCreator.m
@@ -91,13 +91,19 @@
     NSNumber *fontSize = [bottomTabOptions.fontSize withDefault:@(10)];
     NSString *fontWeight = [bottomTabOptions.fontWeight withDefault:nil];
 
+    /**
+     * The title of tabbar item has been cut off in iPad 12.9 with iOS 15.4.
+     * It was mentioned at https://github.com/wix/react-native-navigation/issues/7506
+     * Hotfix: centered the text will resolve problem.
+     */
+
     NSDictionary *normalAttributes = [RNNFontAttributesCreator
         createFromDictionary:[tabItem titleTextAttributesForState:UIControlStateNormal]
                   fontFamily:fontFamily
                     fontSize:fontSize
                   fontWeight:fontWeight
                        color:textColor
-                    centered:NO];
+                    centered:TRUE];
     [self setTitleAttributes:tabItem titleAttributes:normalAttributes];
 
     NSDictionary *selectedAttributes = [RNNFontAttributesCreator
@@ -106,7 +112,7 @@
                     fontSize:fontSize
                   fontWeight:fontWeight
                        color:selectedTextColor
-                    centered:NO];
+                    centered:TRUE];
[...]

PR #7863, opened 2024-03-20
fix(iOS): avoid cutting-off text on bottom bar buttons on iPad.

asafkorem:
For some reason, this change resolves the issue in the attached snapshots.

When `centered=YES`, setting the text alignment to `NSTextAlignmentCenter` ([see here](https://github.com/wix/react-native-navigation/blob/4b48f75ae5e3b319d6243362bbf77f769d50990e/lib/ios/RNNFontAttributesCreator.m#L44)) may incidentally cause the text layout engine to more effectively manage the available space. Doesn't seem like it affects anything in iPhones.

Overall, it seems like a bug in UIKit..

**Before:**
Screenshot 2024-03-20 at 17 16 15

**After:**
Screenshot 2024-03-20 at 18 10 10

### DIFF
--- a/lib/ios/RNNTabBarItemCreator.m
+++ b/lib/ios/RNNTabBarItemCreator.m
@@ -97,7 +97,7 @@
                     fontSize:fontSize
                   fontWeight:fontWeight
                        color:textColor
-                    centered:NO];
+                    centered:YES];
     [self setTitleAttributes:tabItem titleAttributes:normalAttributes];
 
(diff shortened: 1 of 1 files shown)


asafkorem:
Thanks @phuongwd for finding this patch https://github.com/wix/react-native-navigation/issues/7855#issuecomment-2009196262

d4vidi (approved):

PR #7899, opened 2024-08-06
fix: make sure we don't swallow errors inside the mock

Niryo:
adding {cause} should make it easier to understand where the error comes from.


d4vidi:
@Niryo the image link is broken

Niryo:
@d4vidi 
```
try {
  throw new Error('this is the internal error!');
} catch(e) {
  throw new Error('Oops, something went wrong', { cause: e });
}
```
image

asafkorem (approved):

PR #7900, opened 2024-08-06
test: add test for title undefined in merge options

Niryo:


Niryo:
looks like e2e tests fail as well, so maybe it actually works? I need to check it against production and see what happens. closing for now

PR #7904, opened 2024-08-14
Android support topbar title font scaling

mika-lindell:
- Android only
- To enable font scaling that can be adjusted via Android Settings-app:
   - `Accessibility -> Display size and text -> Font size`
- Adds options.topBar.title.allowFontScaling
- Adds and updated unit tests
- When allowFontScaling is enabled, sets top bar title to use SP units. When font scaling is 1.0, it should be the same size as when using DIP 
- Here it's set to false by default, because it can conflict with other top bar options like subtitle, height, fontSize etc. that existing users may have
- I'm thinking about making similar changes to iOS, but I wanted feedback on this one first. iOS has some issues especially when "Larger accessibility sizes"-setting is enabled

```
{
   topBar: {
      title: {
         text: "My title",
         fontSize: 18,
         allowFontScaling: true,
      },
   },
}
```

| allowFontScaling=false | allowFontScaling=true |
| --- | --- |
| Screenshot_20240813-183132 | Screenshot_20240813-175043 |

### Allowing font scaling in top bar title seems to be convention in "standard" Android apps:
| | | |
| --- | --- | --- |
| Screenshot_20240812-155831| Screenshot_20240812-155816| 
|Screenshot_20240812-155559| Screenshot_20240812-155736|


lib/android/app/src/test/java/com/reactnativenavigation/views/TitleAndButtonsContainerTest.kt
```diff
@@ -23,12 +24,14 @@ import org.assertj.core.api.AssertionsForInterfaceTypes.assertThat
...
 import org.mockito.Mockito
 import org.mockito.Mockito.times
+import org.robolectric.annotation.Config
 import kotlin.math.roundToInt
 import kotlin.test.assertFalse
 
 private const val UUT_WIDTH = 1000
 private const val UUT_HEIGHT = 100
 
+@Config(sdk = [30])
```
mika-lindell, line 34: textSizeUnit is not available before api level 30

lib/android/app/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/TitleSubTitleLayout.kt
```diff
@@ -46,7 +48,16 @@ class TitleSubTitleLayout(context: Context) : LinearLayout(context) {
         }
     }
 
-    fun setTitleFontSize(size: Float) = titleTextView.setTextSize(TypedValue.COMPLEX_UNIT_DIP, size)
+    fun setTitleFontSize(size: Float) {
```
mika-lindell, line 51: For it to make sense from the perspective of options, I didn't find a way to set units without setting font size. Using titleTextView.textSize returns scaled font size, so need to have the absolute font size somewhere

Issue #7852, opened 2024-02-21
Android Regression RN 0.73 - `onDropViewInstance` not called anymore, View not cleaned up


bvonheid:
### What happened?

With the update to RN 0.73 the `onDropViewInstance` of ViewManager is not called anymore when a screen is popped from the stack.
As stated [here](https://javadoc.io/static/com.facebook.react/react-native/0.20.1/com/facebook/react/uimanager/ViewManager.html#onDropViewInstance(T)) for the `onDropViewInstance`:
> Called when view is detached from view hierarchy and allows for some additional cleanup by the [ViewManager](https://javadoc.io/static/com.facebook.react/react-native/0.20.1/com/facebook/react/uimanager/ViewManager.html) subclass.

### What was the expected behaviour?

`onDropViewInstance` should be called when a view is popped from the stack.

### Was it tested on latest react-native-navigation?

- [X] I have tested this issue on the latest react-native-navigation release and it still reproduces.

### Help us reproduce this issue!

You can reproduce the error in the playground app:
1. Run [7.38.1](https://github.com/wix/react-native-navigation/releases/tag/7.38.1) on Android 
2. Attach a Debugger to the App
3. Set a Breakpoint in the `com/facebook/react/uimanager/ViewManager.class` file in the `onDropViewInstance` Method
4. Click in Layout tab on Stack -> Push -> Pop
5. => After the Pop `onDropViewInstance` should be called, but it isn't

With the same steps with [7.37.2](https://github.com/wix/react-native-navigation/releases/tag/7.37.2) `onDropViewInstance` is called.

### In what environment did this happen?

React Native Navigation version: -
React Native version: 0.73
Has Fabric (React Native's new rendering system) enabled: no
Node version: -
Device model: -
Android version: -

szaboa:
any workaround then?

tranhoangduong1994:
I am not using this navigation library but having a similar issue when working on a native app using embedded React native components. It seems to me that this issue is caused by the changes in React Native 0.73.

Currently, this issue leads to a lot of performance problems in our app due to memory used by native components not getting released when leaving screens.

The way that I'm working around this issue is to downgrade RN to 0.72.15.

szaboa:
Yes we can also confirm that the issue is in React Native 0.73.

PR #7919, opened 2024-09-30
fix: android onDropViewInstance not invoked when page popped

NormanWangEndeavor:
### What happened?
The onDropViewInstance of ViewManager is not called anymore when a screen is popped from the stack, since RN version `0.71.x`.

### What was the expected behavior?
The onDropViewInstance of ViewManager should be called.

### Was it tested on the latest react-native-navigation?
I have tested this issue on the latest react-native-navigation release and it still reproduces.

### Root Cause
The `unmountReactApplication` removes all children from RN version `0.71.x` which causes the `onDropViewInstance` not to be called. Adding these removed children back resolves the problem.

React Native Navigation version: -
React Native version: since 0.71.x
Has Fabric (React Native's new rendering system) enabled: no
Node version: -
Device model: -
Android version: -

related issue: https://github.com/wix/react-native-navigation/issues/7852

### DIFF
--- a/lib/android/app/src/main/java/com/reactnativenavigation/react/ReactView.java
+++ b/lib/android/app/src/main/java/com/reactnativenavigation/react/ReactView.java
@@ -4,6 +4,8 @@ import android.annotation.SuppressLint;
 import android.content.Context;
 import android.os.Bundle;
 import android.view.MotionEvent;
+import android.view.View;
+import android.view.ViewGroup;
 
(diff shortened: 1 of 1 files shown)


maheshwaripurvesh:
@NormanWangEndeavor Thank you for the quick turnaround. @mobileoss @Niryo @asafkorem I confirm, we can merge this which will resolve #7852

asafkorem:
@d4vidi @gosha212 can you please review this?

PR #7922, opened 2024-10-11
fix(ios): ensure correct initial tab is rendered when onSwitchToTab is set

ismarbesic:
Fixes #7921 by setting _currentTabIndex in the RNNBottomTabsController before initialization happens.

>  _currentTabIndex needs to be set before the super() is called, since - [RNNBottomTabsController selectedViewController] will be called during initialization and which depends on _currentTabIndex. If not set, an initial render of the tab content with index 0 will occur when tabsAttachMode is set to "onSwitchToTab".

Issue #7921, opened 2024-10-11
First tab is always rendered when onSwitchToTab and a custom currentTabIndex is set


ismarbesic:
### What happened?

We have noticed that when using an application with bottom tabs with multiple tabs and the "tabsAttachMode" option is set to "onSwitchToTab" and setting "currentTabIndex" to any value higher than 0 (e.g. 1), the first tab's content will always be rendered together with the content of the second tab. This behavior only occurs when the "currentTabIndex" is not set to its default value. In a test application running 7.39.2 (see code in the repro section), we can see the following in the console:

```
BUNDLE  ./index.js 

LOG  Running "Home" with {"rootTag":1,"initialProps":{"componentId":"Component1"}}
LOG  render home
LOG  Running "Settings" with {"rootTag":11,"initialProps":{"componentId":"Component2"}}
LOG  render settings
```

Upon debugging I inspected the `RNNBottomTabsController` class, placing a breakpoint at `-[RNNBottomTabsController selectedViewController]`, and noticed that this method is called before `_currentTabIndex` is properly initialized, which defaults to 0. This means the renderer will always render the content of the first tab, before the rest of the execution happens which sets the selected tab to index 1.

One solution to fix this issue is to properly initialize the `_currentTabIndex` value before the super() method is called.

<img width="1467" alt="Screenshot 2024-10-11 at 10 21 46" src="https://github.com/user-attachments/assets/9be84335-e19d-4dd2-a64b-1bff4db8c304">

### What was the expected behaviour?

We expect only one screen being rendered:
```
BUNDLE  ./index.js 

LOG  Running "Settings" with {"rootTag":21,"initialProps":{"componentId":"Component2"}}
LOG  render settings
```

### Was it tested on latest react-native-navigation?

- [X] I have tested this issue on the latest react-native-navigation release and it still reproduces.

### Help us reproduce this issue!

Repro (App.tsx):
```
import {Navigation} from 'react-native-navigation';

const HomeScreen = () => {
  console.log('render home');
  return null;
};

HomeScreen.options = {
  topBar: {
    title: {
      text: 'Home',
    },
  },
  bottomTab: {
    text: 'Home',
  },
};

const SettingsScreen = () => {
  console.log('render settings');
  return null;
};

SettingsScreen.options = {
  topBar: {
    title: {
      text: 'Settings',
    },
  },
  bottomTab: {
    text: 'Settings',
  },
};

Navigation.registerComponent('Home', () => HomeScreen);
[...]

PR #7925, opened 2024-10-16
fix(ios): avoid setting selected-index before setup has finished.

asafkorem:
This fixes the issue mentioned on #7921, #7922 and #7924.

This prevents the bug of loading of the tab from index `0` immediately after calling the `super` init regardless of the custom initial rendered tab. More specifically, the original issue was the method `setSelectedIndex` was called immediately when calling to the `super` initializer with a `0` value regardless of the initial `currentTabIndex` value.

After this fix, changing the `selectedIndex` with the non-initial value from `options` will not have any effect before the setup has finished.


asafkorem:
Released with https://github.com/wix/react-native-navigation/releases/tag/7.40.3

Issue #7926, opened 2024-10-17
Fix: Disable iOS 18 TabBar


anduong96:
### Motivation

Ios 18 introduced a new tab bar UI that overwrites the bottom tab UI
![](https://i.sstatic.net/fQOy0b6t.png)

This PR disable that

### Are you willing to resolve this issue by submitting a Pull Request?

- [ ] Yes, I have the time, and I know how to start.
- [X] Yes, I have the time, but I don't know how to start. I would need guidance.
- [ ] No, I don’t have the time and I’m okay to wait for the community / maintainers to resolve this issue.

manuhook:
Thank you very much

bumpingChris:
Hi @anduong96, Thanks for doing this. Any progress so far?

aschmitz-wbd:
I wouldn't want this to be disabled outright. This is the native way iPad deals with tabs. I think there should be an option or it should stay as it is.

Btw. the iPad also dynamically updates to the bottom bar when the app is scaled down, e.g. when in a split window.

Kioshia:
for me the biggest problem with that damn update is that IOS18 sets the TabBar on Top competly messing up my NavigationItems on top... what you just wrote that with the compact style kinda doesnt fix it for me cuase i have 8 NavigationTabs when i have more then 4 it just shows "More" which completly messes up my styling in the app i cant change that cause the app is super duper old so is there a way to just turn it back the os17 one ? I want my 8 Tabs displayed on the bottom...

Meet-iottive:
Using your solution tab bar goes to bottom but tab bar items title and icons are disappearing any solution for that ? @anduong96

Aikenwu:
Maybe use a stupid way to solve the problem, I get the bottom tab height, then if iPad and iOS 18, set all root page padding bottom, to avoid a hidden problem.

anduong96:
> Using your solution tab bar goes to bottom but tab bar items title and icons are disappearing any solution for that ? [@anduong96](https://github.com/anduong96)

No, sorry. This code works for my use case. Im not looking to be a regular contributor.

anduong96:
> I wouldn't want this to be disabled outright. This is the native way iPad deals with tabs. I think there should be an option or it should stay as it is.
> 
> Btw. the iPad also dynamically updates to the bottom bar when the app is scaled down, e.g. when in a split window.

Feel free to lift this code and make it optional

PR #7947, opened 2024-12-15
fix: [iPadOS 18] Disable the new UITabBarController view style in iPadOS 18

maheshwaripurvesh:
## Description

Since iPadOS 18, the Bottom Tab Bar on iPads is now positioned at the top of the screen, floating over the content, instead of at the bottom. This may be the feature by Apple in new version however creating poor user experience

### Fix
1. Check the ios18 and iPad variant 
2. If above condition statisfy, apply following 
    >.traitOverrides.horizontalSizeClass = UIUserInterfaceSizeClassUnspecified;

This will disable the new UITabBarController view style in iPadOS 18
 
### References
**Announcement:** https://developer.apple.com/design/human-interface-guidelines/tab-bars#iPadOS

### DIFF
--- a/lib/ios/RNNBottomTabsController.m
+++ b/lib/ios/RNNBottomTabsController.m
@@ -81,6 +81,14 @@
 
 - (void)viewDidLoad {
     [super viewDidLoad];
+
+    if (@available(iOS 18, *))
+    {
+        if ([UIDevice currentDevice].userInterfaceIdiom == UIUserInterfaceIdiomPad)
+        {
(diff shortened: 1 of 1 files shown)


aschmitz-wbd:
There is an issue https://github.com/wix/react-native-navigation/issues/7926 for that and I want to mention here, that I think it should just be disabled outright. Since it's the native behavior of the iPad, there should at least be an option to turn it off and not have it disabled in all cases.

PR #8005, opened 2025-04-20
Feat/rn 76 to 77 upgrade

gosha212:


lib/src/adapters/NativeRNNTurboEventEmitter.ts
```diff
@@ -1,18 +1,12 @@
...
+import { TurboModule, TurboModuleRegistry, NativeModule } from 'react-native';
 
 export interface Spec extends TurboModule {
   addListener: (eventType: string) => void;
   removeListeners: (count: number) => void;
 }
 
-let eventEmitter: Spec | null = null;
+let eventEmitter: Spec | null;
+eventEmitter = TurboModuleRegistry.get<Spec>('RNNTurboEventEmitter');
```
d4vidi, line 9: 🚀

playground/index.js
```diff
@@ -1 +1,2 @@
+import "setimmediate";
 require('./src/index');
```
d4vidi, line 2: Either:
```js
require('setImmediate');
require('./src/index');
```

OR:
```js
import ...
import
```

Please don't mix both styles 😄

And anyways, shouldn't setImmediate be required inside `src/index.js`?

PR #8015, opened 2025-04-30
iOS: RN 77 no lifecycle events for Top Bar Title component

markdevocht:
The problem was that when setting a titleView of a navigationItem the normal life cycle flow of a view will not be called so the RNNReactView componentWillAppear and DidAppear need to be implicitly called. 

This was somehow removed from the TopBarTitlePresenter.mm unit but is present on the current Master branch's file.

This PR restores those two lines and also actives the Unit test for that module.


gosha212 (approved):
Amazing!

PR #8019, opened 2025-05-08
Better alignment of willAppear and didAppear events throughout the lifecycle

markdevocht:
* New test added to the ApplicationLifecycleTest.test.js for:
componentWillAppear
componentDidAppear

* Added to StaticLifecycleEvents.test.js:
componentWillAppear | EventsOverlay | Component

*waitForRender set to TRUE for the overlay

* willMountComponentsWithRootTag delegate call implemented in RNNReactButtonView.mm
* willMountComponentsWithRootTag delegate callback function implemented in RNNReactView.mm
* Added new 'componentWillAppear' state + handling in LayoutsScreen.tsx

### DIFF
--- a/lib/ios/RNNCommandsHandler.mm
+++ b/lib/ios/RNNCommandsHandler.mm
@@ -467,6 +467,7 @@ static NSString *const setDefaultOptions = @"setDefaultOptions";
     RNNAssertMainQueue();
 
     UIViewController *overlayVC = [_controllerFactory createLayout:layout];
+    overlayVC.waitForRender = [RNNUtils getDefaultWaitForRender];
     [_layoutManager addPendingViewController:overlayVC];
 
     __weak UIViewController *weakOverlayVC = overlayVC;
--- a/lib/ios/RNNReactButtonView.mm
+++ b/lib/ios/RNNReactButtonView.mm
@@ -5,7 +5,7 @@
 - (instancetype)initWithHost:(RCTHost *)host
                   moduleName:(NSString *)moduleName
            initialProperties:(NSDictionary *)initialProperties
-                eventEmitter:(RNNEventEmitter *)eventEmitter
+                eventEmitter:(RNNTurboEventEmitter *)eventEmitter
              sizeMeasureMode:(RCTSurfaceSizeMeasureMode)sizeMeasureMode
          reactViewReadyBlock:(RNNReactViewReadyCompletionBlock)reactViewReadyBlock {
     self = [super initWithHost:host moduleName:moduleName initialProperties:initialProperties eventEmitter:eventEmitter sizeMeasureMode:convertToSurfaceSizeMeasureMode(RCTRootViewSizeFlexibilityWidthAndHeight) reactViewReadyBlock:reactViewReadyBlock];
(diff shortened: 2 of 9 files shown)


lib/ios/RNNReactButtonView.mm
```diff
@@ -15,6 +15,12 @@ - (instancetype)initWithHost:(RCTHost *)host
     return self;
 }
 
+- (void)willMountComponentsWithRootTag:(NSInteger)rootTag {
```
gosha212, line 18: Why do we need it at the first place here? Isn't polymorphism enough?

lib/ios/RNNReactView.mm
```diff
@@ -131,18 +141,43 @@ - (void)componentDidDisappear {
     _willAppear = NO;
     _didAppear = NO;
 }
+    
+- (NSString *)componentId {
```
gosha212, line 145: What is the reason for moving this lines? Code organization for pragma mark?

playground/ios/playground.xcodeproj/xcshareddata/xcschemes/SnapshotRecordTests.xcscheme
gosha212, line 1: I think this is not related to this PR

playground/src/screens/LayoutsScreen.tsx
```diff
@@ -25,19 +25,23 @@ const {
...
 
 interface State {
+  componentWillAppear: boolean;
   componentDidAppear: boolean;
 }
 
 export default class LayoutsScreen extends NavigationComponent<NavigationProps, State> {
   constructor(props: NavigationProps) {
     super(props);
+    console.log("constructor - LayoutsScreen", props);
```
gosha212, line 35: Remove the log

playground/src/screens/StaticLifecycleOverlay.tsx
```diff
@@ -77,6 +77,7 @@ export default class StaticLifecycleOverlay extends React.Component<OverlayProps
   }
 
   constructor(props: OverlayProps) {
+    console.log("constructor - StaticLifecycleOverlay");
```
gosha212, line 80: Remove it

lib/ios/RNNCommandsHandler.mm
```diff
@@ -467,6 +467,7 @@ - (void)showOverlay:(NSDictionary *)layout
     RNNAssertMainQueue();
 
     UIViewController *overlayVC = [_controllerFactory createLayout:layout];
+    overlayVC.waitForRender = [RNNUtils getDefaultWaitForRender];
```
gosha212, line 470: Can we move it inside the factory?

gosha212 (commented):
Amazing work. I have a few comments and a few question

lib/ios/RNNReactView.mm
markdevocht: It's part of the RNNComponentProtocol delegate, it's clearer to read like this. (my opinion)

playground/src/screens/LayoutsScreen.tsx
markdevocht: You don't want the logs for the constructor, np. I'll remove them.

playground/src/screens/StaticLifecycleOverlay.tsx
markdevocht: You don't want the logs for the constructor, np. I'll remove them.

lib/ios/RNNCommandsHandler.mm
markdevocht: Not really, the createLayout function is to general and is called for the same specific layout type (isComponent) too often from all over the place. I don't want to override the waitForRender in every controller that is created here.

gosha212 (approved):
Amazing and very clean work!
Lets wait to the build to be green and merge

Issue #7943, opened 2024-12-02
npx rnn-link is outdates aswell as documentation


rendomnet:
### What happened?

npx rnn-link
is not working with android and kotlin files.

Documentation is also outdated
https://wix.github.io/react-native-navigation/next/docs/installing

### Was it tested on latest react-native-navigation?

- [X] I have tested this issue on the latest react-native-navigation release and it still reproduces.

### In what environment did this happen?

React Native Navigation version: "^7.40.3"
React Native version:  "0.76.3",   
Has Fabric (React Native's new rendering system) enabled: (yes/no) 
Node version: 18
Device model: 
Android version: 35

retyui:
clone: https://github.com/wix/react-native-navigation/issues/7821

retyui:
see https://github.com/wix/react-native-navigation/issues/7821#issuecomment-2514579985

rendomnet:
Can somebody confirm that react-native-navigation is working with react-native 0.76.3?

retyui:
@rendomnet Yes it's working. after linking you need to apply patches for android

do this setup https://www.npmjs.com/package/patch-package#set-up
and put this file in `./patches` [react-native-navigation+7.40.3.patch](https://github.com/user-attachments/files/18006034/react-native-navigation%2B7.40.3.patch) and run once `npx patch-package`

rendomnet:
> @rendomnet Yes it's working. after [linking](https://github.com/wix/react-native-navigation/issues/7821#issuecomment-2514579985) you need to apply patches for android
> 
> do this setup https://www.npmjs.com/package/patch-package#set-up and put this file [react-native-navigation+7.40.3.patch](https://github.com/user-attachments/files/18006034/react-native-navigation%2B7.40.3.patch) into a `./patches` folder and run once `npx patch-package`

I tried it with blank new react-native project and it is not working for iOS and Android.
https://github.com/rendomnet/repro-bug-rnn

rendomnet:
@retyui
This is what 'yarn ios' returns after completing your steps.

[...]

PR #8043, opened 2025-06-09
feat(iOS): RN-77 Swift support (#8037)

markdevocht:
Added Swift support for RN77.
When running pod install, a modulemap for the ReactNativeNavigation is now created with the ReactNativeNavigation-umbrella.h file. This will allow easy linking in Swift by just adding the following to the AppDelegate.swift file:
`import ReactNativeNavigation`

Also the 'rnn-link' script will now detect if we are dealing with an Objc or Swift project and automatically update the Swift file with the following changes:
`import ReactNativeNavigation` is added
`class AppDelegate: RCTAppDelegate` is converted to `class AppDelegate: RNNAppDelegate`.


gosha212:
Fixes #8037.
Implement iOS part of rnn-link #7943

gosha212 (approved):

Issue #8069, opened 2025-07-23
Date picker is not working with RNN https://wix.atlassian.net/browse/MADS-4777


markdevocht:

markdevocht:
**The problem:** The RNDatePicker isn't being displayed correctly.

**The deeper problem:** The RNDatePicker is wrapped in a RCTLegacyViewManagerInteropComponentView.

### What we are missing in our rn-navigation project is the registration of 3rd party components.

In normal RN projects the AppDelegate is inherited from the RCTAppDelegate which is a `RCTComponentViewFactoryComponentProvider` delegate.
The `RCTComponentViewFactoryComponentProvider` delegate provides an interface called 

`- (NSDictionary<NSString *, Class<RCTComponentViewProtocol>> *)thirdPartyFabricComponents;`
which is a class (_RCTThirdPartyComponentsProvider.mm_) that is filled during compile time to include all the 3rd party components that are being used on the React side of the project.


In our RNNavigation project our RNNAppDelegate also inherited from the RCTAppDelegate but when the 
`- (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions`
is called inside the RNNAppDelegate, we don't call the RCTAppDelegate's `application didFinishLaunchingWithOptions`
which contains the line:
`[RCTComponentViewFactory currentComponentViewFactory].thirdPartyFabricComponentsProvider = self;`

So in the RNNavigation project the [RCTComponentViewFactory currentComponentViewFactory].thirdPartyFabricComponentsProvider is NULL and thus doesn't call the necessary functions to register the included 3rd party components. 

So the fix is to add `[RCTComponentViewFactory currentComponentViewFactory].thirdPartyFabricComponentsProvider = self;` to our RNNAppDelegate's `- (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions`

PR #8070, opened 2025-07-23
RNDatePicker not registering - including any 3rd party component - #8069

markdevocht:
We were missing this one liner in our RNNAppDelegate
`[RCTComponentViewFactory currentComponentViewFactory].thirdPartyFabricComponentsProvider = self;`
which would assign a delegate that would take care of calling the auto-generated RCTThirdPartyComponentsProvider.mm to register 3rd party components.


markdevocht:
All functionality in the RCTAppDelegate's createRCTRootViewFactory function, which we don't have, are related to the RCTBridge and I cannot find anything of these that could be translated to RCTHost.

Issue #7945, opened 2024-12-09
React Native 0.77.x


ertugruldogan:
The for Android 27.01.2025 **topic updated**.

![Image](https://github.com/user-attachments/assets/59ffb4c9-803c-47c9-a31c-45fd181be332)

react-native-navigation\lib\android\app\src\main\java\com\reactnativenavigation\react\modal\ModalHostLayout.kt
```java
package com.reactnativenavigation.react.modal

import android.annotation.SuppressLint
import android.annotation.TargetApi
import android.view.View
import android.view.ViewGroup
import android.view.ViewStructure
import android.view.accessibility.AccessibilityEvent
import com.facebook.react.bridge.LifecycleEventListener
import com.facebook.react.bridge.ReactContext
import com.facebook.react.bridge.UiThreadUtil
import com.facebook.react.uimanager.ThemedReactContext
import com.reactnativenavigation.options.Options
import com.reactnativenavigation.options.params.Bool
import com.reactnativenavigation.utils.CompatUtils
import com.reactnativenavigation.viewcontrollers.viewcontroller.YellowBoxDelegate
import com.reactnativenavigation.viewcontrollers.viewcontroller.overlay.ViewControllerOverlay
import java.util.*

@SuppressLint("ViewConstructor")
open class ModalHostLayout(reactContext: ThemedReactContext) : ViewGroup(reactContext), LifecycleEventListener {
    val viewController = ModalLayoutController(
        reactContext,
        reactContext.currentActivity, CompatUtils.generateViewId().toString(),
        YellowBoxDelegate(reactContext), Options().apply {
            hardwareBack.dismissModalOnPress = Bool(false)
        }, ViewControllerOverlay(reactContext),
        getHostId = { this.id }
    )
    private val mHostView = viewController.view.modalContentLayout

    init {
        (context as ReactContext).addLifecycleEventListener(this)
    }

@TargetApi(23)
override fun dispatchProvideStructure(structure: ViewStructure?) {
    structure?.let { 
        mHostView.dispatchProvideStructure(it)
    }
}


    override fun onLayout(changed: Boolean, l: Int, t: Int, r: Int, b: Int) {}

    override fun addView(child: View?, index: Int) {
        UiThreadUtil.assertOnUiThread()
        mHostView.addView(child, index)
    }

    override fun getChildCount(): Int {
        return mHostView.childCount
    }

    override fun getChildAt(index: Int): View? {
        return mHostView.getChildAt(index)
    }

[...]

PR #8084, opened 2025-08-08
[Android] Fix safe area miscalculation on HONOR phones

Reeywhaar:
Somehow bottomButtons layout breaks keyboard safe area on Honor (MagicOs) phone models.

### Description:
We have an app with bottomTabs layout.
When we push some modal that has TextInput and focus it, `adjustResize` doesn't work and important views that are placed on bottom became hidden behind the keyboard. But only on Honor phones 🤷‍♂️

**Initial screen:**
Screenshot_1754628010

**Before fix:**
Screenshot_1754628261

**After fix:**
Screenshot_1754628018

## Additional info:
`react-native`: `0.77.3` (with some patches from [here](https://github.com/wix/react-native-navigation/issues/7945))
`react-native-navigation`: `7.50.0`
`softInputMode` is `adjustResize`

## Disclaimer
This is more of something to think about, rather that thorough merge request, as I'm not an advanced android developer and this works only for my case, but may break ton of others.
Screenshots are emulator representation of how it looks, not a real device screenshots.

### DIFF
--- a/lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/bottomtabs/BottomTabsController.java
+++ b/lib/android/app/src/main/java/com/reactnativenavigation/viewcontrollers/bottomtabs/BottomTabsController.java
@@ -317,7 +317,7 @@ public class BottomTabsController extends ParentController<BottomTabsLayout> imp
     protected WindowInsetsCompat onApplyWindowInsets(View view, WindowInsetsCompat insets) {
         Insets sysInsets = insets.getInsets(WindowInsetsCompat.Type.systemBars());
         view.setPaddingRelative(0, 0, 0, sysInsets.bottom);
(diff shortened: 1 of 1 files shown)


gosha212:
Hi, what android version you're running. It looks like a problem with edge to edge and not specific brand

reeywhaar:
Problem with edge2edge and HONOR specifically, because everything works fine on e.g Samsung. OS is MagicOS 9 (Android 15)

reeywhaar:
I see there are new additions to BottomTabsController.java which may cover my PR already, though I need to test it. Maybe PR is already obsolete. I'll let you know as soon as I can.

gosha212:
we still have to implement the edge to edge support

reeywhaar:
Fixed in newer releases, thank you!

PR #8087, opened 2025-08-13
Fix keyboard-awareness bug

d4vidi:
Fix issue with new (partial) support for edge2edge with bottom-tabs, blocking the UI from being aware of the on-screen keyboard (i.e. in inset computation).

Originally introduced by adding the fundamental support in #7991 -- specifically by returning `CONSUMED` here: https://github.com/wix/react-native-navigation/pull/7991/files#diff-9fb728cd9594f9968d72c811aa36c24769eccb338fbba2085acd3b9e023549a1R310


playground/src/screens/KeyboardScreen.tsx
```diff
@@ -104,20 +127,21 @@ export default class KeyboardScreen extends NavigationComponent<Props> {
             />
           </View>
         </ScrollView>
-        <View style={styles.footer}>
-          <Text style={styles.input}> {KEYBOARD_LABEL}</Text>
```
d4vidi, line 108: this is unbelievable

Issue #8082, opened 2025-08-07
[iOS] Property 'setImmediate' doesn't exist


topi-identio:
### What happened?

1. Created a new RN 0.77.3 project using `npx react-native@0.77.3 init ProjectName --template react-native-template-typescript`
2. Following installation docs in `https://wix.github.io/react-native-navigation/docs/installing`: installed `react-native-navigation`, ran `npx rnn-link` and updated `index.js`
3. Ran `react-native start` and `react-native run-ios` for iOS Simulator (iPhone 16, iOS 18.5)
5. iOS app builds but starts stuck with errors `Property 'clearImmediate' doesn't exists` and `Property 'setImmediate' doesn't exist`  <img width="1197" height="707" alt="Image" src="https://github.com/user-attachments/assets/aac3e661-7191-4e96-8864-5babd8f73ff8" />

### What was the expected behaviour?

No errors, app works

### Was it tested on latest react-native-navigation?

- [x] I have tested this issue on the latest react-native-navigation release and it still reproduces.

### Help us reproduce this issue!

Simple repo with the mentioned steps done: https://github.com/topi-identio/rn-0.77-test

### In what environment did this happen?

React Native Navigation version: 8.2.1
React Native version: 0.77.3
Has Fabric (React Native's new rendering system) enabled: yes
Node version: 20.17.0
Device model: iPhone 16 (Simulator)
iOS version: 18.5

topi-identio:
After some investigation this seems to be related to the RN `StatusBar` component as it doesn't happen if I remove it.

topi-identio:
With the example repo the issue seems to be related to the RN `StatusBar` component. The errors go away once that's removed. I also tried installing the dependency `setimmediate`, which gets rid of the error but causes another one.

In our actual repo we get the same issue on startup, but caused by some other component, not sure what it is yet.

huynhuchuy2001:
You can try to install the setimmediate library from npm or yarn, and import it into the index.js

topi-identio:
I can get our app to run with that change but it doesn't sound like a proper solution if something's not being properly initialised.

retyui:
Have the same issue but in different place **(RNN@8.1.2, RN:0.77.3 + RCT_NEW_ARCH_ENABLED=1)**

```tsx
// node_modules/react-native/Libraries/LogBox/Data/LogBoxData.js
  function handleUpdate() {
    if (updateTimeout == null) {
[...]

PR #8089, opened 2025-08-14
Fix Property setImmediate doesn't exist issue

markdevocht:
The RNNAppDelegate was implementing the RCTTurboModuleManagerDelegate with having the getTurboModule:jsInvoker returning nullptr.
Because the RNNAppDelegate's RCTTurboModuleManagerDelegate implementation was almost idenitcal to RCTAppDelegate's (except for the nullptr return) the RNNAppDelegate implementation of it is not needed so was removed.


gosha212:
Fixes #8082

PR #8104, opened 2025-08-31
Fixed crash in Android 

gosha212:
Crash in NavigationTurboModule.kt.
Fixed null safety

### DIFF
--- a/lib/android/app/src/main/java/com/reactnativenavigation/react/NavigationTurboModule.kt
+++ b/lib/android/app/src/main/java/com/reactnativenavigation/react/NavigationTurboModule.kt
@@ -275,7 +275,7 @@ class NavigationTurboModule(
         if (navigator == null) {
             Log.e("NavigationTurboModule", "navigator is null!")
         }
-        return activity()?.navigator
+        return navigator
     }
 
     private fun handle(task: Runnable) {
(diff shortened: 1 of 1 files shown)


lib/android/app/src/main/java/com/reactnativenavigation/react/NavigationTurboModule.kt
```diff
@@ -263,25 +265,36 @@ class NavigationTurboModule(
...
     }
 
-    private fun navigator(): Navigator {
-        return activity().navigator
+    private fun navigator(): Navigator? {
+        val navigator = activity()?.navigator
+        if (navigator == null) {
+            Log.e("NavigationTurboModule", "navigator is null!")
+        }
+        return activity()?.navigator
```
d4vidi, line 278:
```suggestion
        return navigator
```

gosha212:
> I know it's hard to repro, but still - is there a way we can test this e2e? Perhaps using a special activity with "tricks" inside it's onResume() etc.

I've a reproduction for this bug. Just start the app before the metro in debug mode.

d4vidi:
> > I know it's hard to repro, but still - is there a way we can test this e2e? Perhaps using a special activity with "tricks" inside it's onResume() etc.
> 
> I've a reproduction for this bug. Just start the app before the metro in debug mode.

So can we run detox in debug mode to test this?

d4vidi (approved):

angelos3lex:
> > I know it's hard to repro, but still - is there a way we can test this e2e? Perhaps using a special activity with "tricks" inside it's onResume() etc.
> 
> I've a reproduction for this bug. Just start the app before the metro in debug mode.

I verify. On android, if we open first the app then the metro and reload, then we get activity is null error.
If we first open metro, then the app, it's ok. (RN 0.83 and 8.7.4 navigation)

Issue #8061, opened 2025-07-09
Overlay with a view with style={zIndex} blocks touches


M-i-k-e-l:
### What happened?

Overlay that has a view which has a style that includes `zIndex` behaves as if `interceptTouchOutside` is `true` when it is `false`.

To reproduce:
1. Checkout `overlay-with-zindex` (current `master` slightly modified, see [diff](https://github.com/wix/react-native-navigation/compare/overlay-with-zindex))
2. yarn -> yarn start -> install the app
3. Go to the Navigation tab
4. Back Button
5. Show Overlay
6. (Try to press on) Modal

You cannot press on the buttons

### What was the expected behaviour?

Pressing Modal can be done and the modal is opened

### Was it tested on latest react-native-navigation?

- [x] I have tested this issue on the latest react-native-navigation release and it still reproduces.

### Help us reproduce this issue!

See steps above

### In what environment did this happen?

React Native Navigation version: 8.1.0
React Native version: 0.77.2
Has Fabric (React Native's new rendering system) enabled: (yes/no) yes
Node version: v22.13.1
Device model: emulator
Android version: API 34

merteroglu:
I'm having the same problem. Even without setting any zIndex, as soon as I register an overlay, the UI stops responding to touches. It's only Android. iOS is working well.

React Native Navigation version: 8.2.1
React Native version: 0.77.3
Has Fabric (React Native's new rendering system) enabled: (yes/no) yes
Device: I tried it on several real devices.

`       buildToolsVersion = "35.0.0"
        minSdkVersion = 26
        compileSdkVersion = 35
        targetSdkVersion = 34
        ndkVersion = "27.1.12297006"
`

ee-dd-cc:
same issue

gosha212:
@merteroglu can you create a sample project that reproduces the issue?

merteroglu:
> [@merteroglu](https://github.com/merteroglu) can you create a sample project that reproduces the issue?

I created a new project and added the other dependencies I was suspicious about. I used the showOverlay function to mimic the structure I use in my actual app. When I remove it, touch interactions start working again. But as soon as I add it, the UI becomes unclickable. I'm testing this on a real Android device in debug mode.

https://github.com/merteroglu/rnn_overlay_problem_example

topi-identio:
[...]

PR #8112, opened 2025-09-08
fix(android): #8061 new arch overlay problem

merteroglu:
Hello,
I’ve been investigating the overlay issue for a while. While debugging, I noticed that the touch events were falling into child 0 instead of 1 in this view. With this change, touch interactions started working correctly.

I also tested the overlay with different configurations, and it worked as expected.

This fixes the issue described here: https://github.com/wix/react-native-navigation/issues/8061

### DIFF
--- a/lib/android/app/src/main/java/com/reactnativenavigation/views/touch/OverlayTouchDelegate.kt
+++ b/lib/android/app/src/main/java/com/reactnativenavigation/views/touch/OverlayTouchDelegate.kt
@@ -19,7 +19,7 @@ open class OverlayTouchDelegate(private val component: ComponentLayout, private 
     }
 
     @VisibleForTesting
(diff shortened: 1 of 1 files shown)


gosha212:
hi @merteroglu. Thanks for the contribution. This fix "solves" the issue by mistake. The second child is always null. I will create a proper fix based on your idea. It gave me a good idea

merteroglu:
> hi @merteroglu. Thanks for the contribution. This fix "solves" the issue by mistake. The second child is always null. I will create a proper fix based on your idea. It gave me a good idea

Hi,
Yes, I later realized that this was only a fix for my specific use case. I’m really glad it helped contribute to finding a more permanent solution.
Wishing you all the best with your work!

Issue #7990, opened 2025-04-02
RNNav iOS18 Compat


d4vidi:
### Describe your idea

Validate iOS 18 compatibility after Detox is ready: https://github.com/wix/Detox/issues/4749

Tasks:
- [ ] Check and make fixes and adjustments
- [ ] Run regularly on iOS on CI

### Are you willing to resolve this issue by submitting a Pull Request?

- [ ] Yes, I have the time, and I know how to start.
- [ ] Yes, I have the time, but I don't know how to start. I would need guidance.
- [ ] No, I don’t have the time and I’m okay to wait for the community / maintainers to resolve this issue.

PR #8132, opened 2025-10-26
RNNav iOS 18 Compatibility #7990

markdevocht:
iOS 18 support and implementation in tests.

### DIFF
--- a/e2e/Orientation.test.js
+++ b/e2e/Orientation.test.js
@@ -12,6 +12,10 @@ describe.e2e(':ios: orientation', () => {
     await elementById(TestIDs.SHOW_ORIENTATION_SCREEN).tap();
   });
 
+  afterEach(() => {
+    device.setOrientation('portrait');
+  });
+
   it('landscape and portrait array', async () => {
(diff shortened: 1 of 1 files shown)


e2e/Orientation.test.js
```diff
@@ -40,6 +40,9 @@ describe.e2e(':ios: orientation', () => {
     await expect(element(by.id(TestIDs.LANDSCAPE_ELEMENT))).toBeVisible();
     await device.setOrientation('portrait');
     await expect(element(by.id(TestIDs.LANDSCAPE_ELEMENT))).toBeVisible();
+    await device.setOrientation('landscape');
+    await expect(element(by.id(TestIDs.LANDSCAPE_ELEMENT))).toBeVisible();
     await elementById(TestIDs.DISMISS_BTN).tap();
+    await device.setOrientation('portrait');
```
gosha212: Why do we have an action without expect?

gosha212:
Please note that the tests are not working

e2e/Orientation.test.js
markdevocht: Understood! Added AfterEach in next push

PR #8171, opened 2025-12-08
Documentation added for how to deal with Firebase with static frameworks

markdevocht:
There is no need to change the RNN podspecs file on RN0.79 and the only change that needs to be done to support Firebase with static frameworks is done on the user's podfile.
This change is now added in our documentation.
Ticket: https://github.com/wix/react-native-navigation/issues/8026


gosha212 (dismissed):

angelos3lex:
@markdevocht Is this change needed for RN 0.83 too together with 8.7.3 RNN + firebase? Or only for RN 0.79?

PR #8192, opened 2025-12-31
fix for fading modal dismissal

markdevocht:
**The Problem**
When dismissing a modal with custom animations, the screen behind (the presenting view controller) was turning black after the animation completed.

**Root Cause**
The original code didn't differentiate between present and dismiss transitions. It treated both the same way, which violates UIKit's expected behavior.

https://github.com/user-attachments/assets/0863bf2f-cc43-417b-b033-9bcb8ecd3286

### DIFF
--- a/ios/ScreenAnimationController.mm
+++ b/ios/ScreenAnimationController.mm
@@ -75,14 +75,29 @@
 }
 
 - (void)prepareTransitionContext:(id<UIViewControllerContextTransitioning>)transitionContext {
-	UINavigationController *toViewController =
-		[transitionContext viewControllerForKey:UITransitionContextToViewControllerKey];
-	toViewController.view.alpha = 0;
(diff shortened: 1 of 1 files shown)


gosha212 (approved):

manuhook:
Hi @markdevocht, I'm style having an issue even with this fix on iOS.

I have no animation when I'm dismissing the modal. 
The code below was working on old arch. No problem on showModal.
I'm using RNN 8.7.3

```
options: {
      modalPresentationStyle: 'overCurrentContext',
      animations: {
        showModal: {
          enter: {
            alpha: {
              from: 0,
              to: 1,
              duration: 400,
            },
          },
        },
        dismissModal: {
          exit: {
            alpha: {
              from: 1,
              to: 0,
              duration: 200,
            },
          },
        },
      },
  }
```
  
  Have you an idea ?
  
  Thanks!

PR #8210, opened 2026-01-21
fix: Update react-native-webview resolution in yarn.lock

bvonheid:
During `yarn install` the error: `YN0001: │ RequestError: getaddrinfo ENOTFOUND npm.dev.wixpress.com` occurs.
With the adaption in the `yarn.lock` the `yarn install` command is fixed.


markdevocht:
I added it to a different PR now.

PR #8215, opened 2026-01-26
Fix Mock exports for external consumers

bvonheid:
## Context / Problem
External consumers could not use the `Mock` anymore because react-native-builder-bob wasn’t compiling the Mock files in their previous location. In practice that meant the published package didn’t contain a usable, compiled Mock entry for downstream projects, even though the repository had the implementation.

## Summary

This PR makes the Mock implementation consumable from the published package by ensuring it’s included in the build output, cleaning up a legacy entry file, and aligning package export mappings for different toolchains (including React Native).

---

## Changes included

- **Move `Mock` directory into `src`**
  - Relocated `Mock` implementation under `src/Mock`
  - Updated imports across the module to match the new location

- **Remove legacy `Mock` entry file + adjust Jest**
  - Deleted the unused `Mock.ts`
  - Updated `jest.config.js` (and related config) to reflect the new structure

- **Update `package.json` export mappings**
  - Added/updated `react-native` and `source` mappings for `./Mock` to ensure correct resolution for different bundlers/tooling

---

## Testing / Verification

- Run unit tests (Jest) to confirm config changes and paths are correct.

### DIFF
--- a/ios/RNNOverlayManager.mm
+++ b/ios/RNNOverlayManager.mm
@@ -17,6 +17,8 @@
     overlayWindow.rootViewController.view.backgroundColor = [UIColor clearColor];
     [overlayWindow setWindowLevel:UIWindowLevelNormal];
     [overlayWindow setHidden:NO];
+    
+    overlayWindow.previousWindow.accessibilityElementsHidden = YES;
 }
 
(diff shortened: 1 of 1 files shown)


markdevocht:
@bvonheid Thank you for your contribution! I was just starting to investigate the same problem when I came accross your PR.

PR #8236, opened 2026-02-24
re-aligning the back button on android and iOS when using icon background logic

markdevocht:
With the new icon background logic added we need to re-align the backbutton in both iOS and Android.


adids1221 (approved):
Approved, great 🙏 🚀 🖌️

PR #8239, opened 2026-02-26
iOS add title customization to scroll edge appearance

markdevocht:
updating the missing pieces -
-mergeOption
-added some safety


Beat-YT:
Seems complete to me. 

I haven’t tested the merge option part yet but I don’t see at a glance any other issues. Not a lot of changes either.

PR #8245, opened 2026-03-11
iOS 26 bottom tab fix

markdevocht:


oferRounds:
@markdevocht thank you for your work here! it will help us a lot!

PR #8252, opened 2026-03-18
Navigation bar hierarchy fix

markdevocht:
In iOS 26.1 the hierarchy of the NavigationBar was changed so the back button could not be assigned the right ID for finding the object in tets.


yedidyak (approved):
Looks good after the fixes. One small non-blocking follow-up remains around clearing stale back-button testIDs when `setBackButtonTestID:` is called with `nil`, but I do not think that should hold up this PR.

PR #8258, opened 2026-03-23
fix

markdevocht:


GarryLaly:
@markdevocht please check this version is broken for me https://github.com/wix/react-native-navigation/issues/8260

PR #8262, opened 2026-03-30
Edge2edge android

markdevocht:
## Android Edge-to-Edge Support
Adds opt-in edge-to-edge display support for Android 15+ (API 35), where the system enforces edge-to-edge by default.
### Key changes
- **Theme-based opt-in** — Edge-to-edge is disabled by default. Apps opt in by setting `android:windowOptOutEdgeToEdgeEnforcement` to `false` in their theme. `NavigationActivity.enableEdgeToEdge()` reads this attribute and calls `EdgeToEdge.enable()` accordingly. The method is `protected` and overridable for custom behavior.
- **View-based system bar backgrounds** — Replaced deprecated `window.statusBarColor` / `window.navigationBarColor` APIs with view-based backgrounds in `SystemUiUtils`. A status bar background view is sourced from the DecorView, and a navigation bar background view is created and sized dynamically via `WindowInsetsCompat`. Both fall back to the deprecated window APIs when edge-to-edge is not active.
- **Correct inset handling for `SafeAreaView`** — `ComponentViewController.onApplyWindowInsets` now correctly passes navigation bar insets through to React Native's `SafeAreaView` when edge-to-edge is active, so content is properly inset from the system navigation bar. When edge-to-edge is off, only IME insets are forwarded (preserving existing behavior).
- **Bottom tabs edge-to-edge awareness** — `BottomTabsPresenter` and `BottomTabsContainer` account for navigation bar insets when drawing behind system bars.
- **Gesture vs 3-button detection** — `SystemUiUtils` detects the navigation mode (gesture vs 3-button) at runtime via tappable element insets, applying appropriate default opacity for the navigation bar background.
- **E2E snapshot updates** — Updated baseline screenshots for BottomTabs stylizing tests and overlay banner test to match the current rendering.

https://wix.atlassian.net/browse/WOAINFRA-3303


android/src/main/java/com/reactnativenavigation/utils/SystemUiUtils.kt
```diff
@@ -45,98 +57,165 @@ object SystemUiUtils {
...
+            FrameLayout.LayoutParams.MATCH_PARENT, 0, Gravity.BOTTOM
+        )
+        contentLayout.addView(view, params)
+        navBarBackgroundView = view
+
+        ViewCompat.setOnApplyWindowInsetsListener(view) { v, insets ->
+            val navBarHeight = insets.getInsets(WindowInsetsCompat.Type.navigationBars()).bottom
+            val tappableHeight = insets.getInsets(WindowInsetsCompat.Type.tappableElement()).bottom
+            val wasThreeButton = isThreeButtonNav
+            isEdgeToEdgeActive = navBarHeight > 0
```
yedidyak: `setupSystemBarBackgrounds()` is installed for every activity, but this listener flips `isEdgeToEdgeActive` to `true` whenever `navigationBars().bottom > 0`. That is also true on pre-35 devices and on API 35 themes that keep `windowOptOutEdgeToEdgeEnforcement=true`, so `EdgeToEdge.enable()` never has to run for the global flag to turn on. `ComponentViewController.onApplyWindowInsets()` then starts forwarding nav-bar insets into `SafeAreaView`, which changes layout for apps that never opted into edge-to-edge.

android/src/main/java/com/reactnativenavigation/viewcontrollers/bottomtabs/BottomTabsPresenter.kt
```diff
@@ -376,4 +389,10 @@ class BottomTabsPresenter(
         val margin = UiUtils.dpToPx(bottomTabsContainer.context, marginDp.toFloat()).roundToInt()
         return margin
     }
+
+    private fun syncNavigationBarColor(options: Options, tabsColor: Int) {
```
yedidyak, line 393: `syncNavigationBarColor()` only checks the raw merge payload. In merge flows that object often contains just the delta, so an explicitly configured navigation-bar color inherited from the current or default options still looks `unset` here. Updating only `bottomTabs.backgroundColor` will then recolor the nav bar to match the tabs even though the effective navigation-bar color was already defined higher up.

android/src/main/java/com/reactnativenavigation/utils/SystemUiUtils.kt
```diff
@@ -45,98 +57,165 @@ object SystemUiUtils {
...
+            isThreeButtonNav = tappableHeight > 0
+            if (isThreeButtonNav != wasThreeButton) {
```
yedidyak, line 109: When the listener detects a gesture/3-button mode transition, it unconditionally repaints the injected navigation-bar background with `getDefaultNavBarColor()`. That loses any explicit color previously applied through `navigationBar.backgroundColor` or `syncNavigationBarColor`, so simply switching navigation mode can revert a customized bar back to the default black/80%-black value.

android/src/main/java/com/reactnativenavigation/NavigationActivity.java
```diff
@@ -146,6 +153,25 @@ public void onReload() {
...
+     */
+    protected void enableEdgeToEdge() {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.VANILLA_ICE_CREAM) {
+            android.content.res.TypedArray a = getTheme().obtainStyledAttributes(
+                    new int[]{android.R.attr.windowOptOutEdgeToEdgeEnforcement});
+            boolean optOut = a.getBoolean(0, true);
+            a.recycle();
+            if (!optOut) {
+                EdgeToEdge.enable(this);
+                SystemUiUtils.activateEdgeToEdge();
```
yedidyak: The new internal edge-to-edge flag is only set in the default `NavigationActivity.enableEdgeToEdge()` implementation via `SystemUiUtils.activateEdgeToEdge()`. If an app overrides `enableEdgeToEdge()` and calls `EdgeToEdge.enable(this)` itself, the override never reaches that helper, so `ComponentViewController` still treats edge-to-edge as inactive and skips forwarding nav-bar insets to `SafeAreaView`, while nav-bar colors continue going through the non-edge-to-edge path. That is also the exact override pattern documented in the new guide, so custom adopters of this API will get partially broken edge-to-edge behavior unless they know to call an extra internal helper.

android/src/main/java/com/reactnativenavigation/viewcontrollers/bottomtabs/BottomTabsPresenter.kt
markdevocht: Fixed

android/src/main/java/com/reactnativenavigation/utils/SystemUiUtils.kt
markdevocht: Fixed

android/src/main/java/com/reactnativenavigation/utils/SystemUiUtils.kt
markdevocht: Fixed

android/src/main/java/com/reactnativenavigation/NavigationActivity.java
markdevocht: Fixed

yedidyak: The last change fixes the default `NavigationActivity` path, because the base implementation now calls `SystemUiUtils.activateEdgeToEdge()` after `EdgeToEdge.enable(this)`. But the custom override path is still problematic: `ComponentViewController` continues to gate bottom inset forwarding on `SystemUiUtils.isEdgeToEdgeActive()`, and the docs still show an override example that only calls `EdgeToEdge.enable(this)`. So apps following that documented override pattern can still enable Android edge-to-edge without enabling RNN's internal edge-to-edge state.

yedidyak (approved):
Re-reviewed after the latest edge-to-edge follow-up changes. The remaining override-path concern now looks addressed: custom overrides are routed through activateEdgeToEdge(), and the docs match that contract.

Issue #8198, opened 2026-01-10
Android app crash on start


OneAboveAll1964:
### What happened?

On some devices the app seems to crash immediately as shown by crashlytics, giving the error
com.reactnativenavigation.viewcontrollers.stack.StackAnimator....

Fatal Exception: java.lang.NullPointerException: Parameter specified as non-null is null: method com.reactnativenavigation.viewcontrollers.stack.StackAnimator.<init>, parameter context
       at com.reactnativenavigation.viewcontrollers.stack.StackAnimator.<init>(:2)
       at com.reactnativenavigation.viewcontrollers.stack.StackControllerBuilder.<init>(StackControllerBuilder.java:35)
       at com.reactnativenavigation.options.LayoutFactory.createStack(LayoutFactory.java:190)
       at com.reactnativenavigation.options.LayoutFactory.create(LayoutFactory.java:86)
       at com.reactnativenavigation.react.NavigationTurboModule.setRoot$lambda$0(NavigationTurboModule.kt:86)
       at com.reactnativenavigation.react.NavigationTurboModule.$r8$lambda$1CBKqa2vcSDrUXVo1M9sQBtPfiI()
       at com.reactnativenavigation.react.NavigationTurboModule$$ExternalSyntheticLambda4.run(D8$$SyntheticClass)
       at com.reactnativenavigation.react.NavigationTurboModule.handle$lambda$15(NavigationTurboModule.kt:287)
       at com.reactnativenavigation.react.NavigationTurboModule.$r8$lambda$ztKyjHGNL_lQJwL4DtqvybN_ZwE()
       at com.reactnativenavigation.react.NavigationTurboModule$$ExternalSyntheticLambda8.run(D8$$SyntheticClass)
       at android.os.Handler.handleCallback(Handler.java:958)
       at android.os.Handler.dispatchMessage(Handler.java:99)
       at android.os.Looper.loopOnce(Looper.java:205)
       at android.os.Looper.loop(Looper.java:294)
       at android.app.ActivityThread.main(ActivityThread.java:8500)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:640)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1026)
        

### What was the expected behaviour?

The app not to crash

### Was it tested on latest react-native-navigation?

[...]

PR #8267, opened 2026-04-09
fix crashes on android by adding null safety

OneAboveAll1964:
in response to this [issue](https://github.com/wix/react-native-navigation/issues/8198)


markdevocht:
@OneAboveAll1964 , I've created an internal PR: https://github.com/wix/react-native-navigation/pull/8269 with your changes and added some more safeties. Please have a look.

OneAboveAll1964:
@markdevocht thanks for taking the time, and the changes you did look legit, my main concern making these changes was, if it doesnt crash then what would happen? because then we would have to rely on customer reports rather than crashylitics

and just for reference the two crashes these changes aim to fix are not small by any means

Screenshot 2026-04-13 at 10 52 25 AM
Screenshot 2026-04-13 at 10 52 31 AM

last 90 days on crashlytics ^

markdevocht:
@OneAboveAll1964 so you're saying that I should take your PR as is? I have no problem with that.

OneAboveAll1964:
@markdevocht no not at all, i was just asking if any of our changes will have implications, ofc you know better and your changes surely cover some cases i didn't think of, i was referring to both our changes not just yours

and in fact i have just moved your changes to our patch as well

markdevocht (approved):

PR #8276, opened 2026-04-19
WOAA-286 endless loading spinner

markdevocht:
## Bug

When navigating between screens of the same component type (pop + push in quick succession), the new screen loads data into a shared store on mount, but the old screen's `componentWillUnmount` fires **after** the new screen has already mounted and appeared — clearing the store and leaving the new screen stuck in a loading/empty state.

## Root Cause

When a screen is popped on iOS, RNN relies on ARC deallocation of the `UIViewController` to tear down the React surface (via `RCTSurfaceHostingView.dealloc` → `[_surface stop]`). This deallocation happens asynchronously — after the pop animation completes and UIKit releases its internal references to the view controller.

If a new screen is pushed immediately after the pop (especially the same component type), the new screen's React surface mounts **before** the old one is destroyed. This means `componentDidMount` of the new screen fires before `componentWillUnmount` of the old screen. Any shared state that is loaded on mount and cleared on unmount gets corrupted — the old screen's late unmount wipes data the new screen just loaded.

Timeline observed (from the video):
- New screen **componentDidMount** — loads data into shared store
- New screen **componentDidAppear** — screen is visible
- Old screen **componentWillUnmount** (25ms later) — clears the shared store
- New screen is now stuck showing stale/empty state

## Fix

Modified `RNNStackController.popViewControllerAnimated:` to explicitly tear down the React surface during the pop, rather than waiting for ARC:

1. **Snapshot** the popped view controller's view so the pop animation still renders correctly
2. **Send `componentDidDisappear`** through the React view while it's still alive, so the JS component receives the event before unmount
3. **Destroy the React view** (`destroyReactView`) immediately after `[super popViewControllerAnimated:]` returns — this stops the Fabric surface synchronously, ensuring `componentWillUnmount` fires before any subsequent push creates a new surface

The existing `destroyReactView` call in `StackControllerDelegate.didShowViewController:` becomes a safe no-op since `reactView` is already nil.

## Test

Added an E2E test (`"pop and re-push same component should not have stale unmount"`) that reproduces the exact race condition:
- A screen sets a module-level shared variable on mount and clears it on unmount
- The test pushes the screen, then fires `pop` + `push` in the same JS tick
- After a delay, it verifies the shared variable wasn't cleared by a stale unmount

Without the fix the test fails (shows `"stale_unmount"`), with the fix it passes. All 156 existing iOS E2E tests continue to pass with no regressions.


ios/RNNStackController.mm
```diff
@@ -42,7 +42,26 @@ - (void)mergeChildOptions:(RNNNavigationOptions *)options child:(UIViewControlle
...
+        if (snapshot) {
+            snapshot.frame = topVC.view.bounds;
+            snapshot.autoresizingMask =
+                UIViewAutoresizingFlexibleWidth | UIViewAutoresizingFlexibleHeight;
+            [topVC.view addSubview:snapshot];
+        }
+    }
+
+    if ([topVC isKindOfClass:[RNNComponentViewController class]]) {
+        [[(RNNComponentViewController *)topVC reactView] componentDidDisappear];
```
yedidyak: Root pop still reaches this path, and popAnimated treats a nil return as a rejected or no-op pop. Sending componentDidDisappear before we know super actually removed a controller means a root pop now tells JS the screen disappeared even though it remains visible, which can trigger the same cleanup race this PR is fixing.

ios/RNNStackController.mm
```diff
@@ -42,7 +42,26 @@ - (void)mergeChildOptions:(RNNNavigationOptions *)options child:(UIViewControlle
...
+    }
+
+    UIViewController *poppedVC = [super popViewControllerAnimated:animated];
+    [poppedVC destroyReactView];
```
yedidyak: This only fixes the single-screen pop path. popTo and popToRoot still go through popToViewController: or popToRootViewControllerAnimated:, and StackControllerDelegate tears down only the previously presented controller. Additional popped screens still rely on delayed ARC teardown, so popTo* followed immediately by push can still replay a stale componentWillUnmount against the newly mounted screen.

markdevocht: Added overrides for popToViewController:animated: and popToRootViewControllerAnimated: that iterate all popped VCs and tear down their React views. The same teardownPoppedControllers: helper is shared across all three pop paths.

ios/RNNStackController.mm
markdevocht: componentDidDisappear and destroyReactView now only run on VCs that were actually popped (guarded by the return value of [super popViewControllerAnimated:]). Root pop returns nil → no teardown, no spurious events.

yedidyak (approved):
Approving per request. I still have one follow-up concern on the custom iOS pop-animation path, but treating it as a non-blocking risk for this review.

PR #8312, opened 2026-05-25
iOS 26 Navigation bar button alignment

markdevocht:


ios/TopBarAppearancePresenter.mm
```diff
@@ -1,16 +1,116 @@
...
+        return;
+    }
+
+    if (color) {
+        if (@available(iOS 26.0, *)) {
+            [appearance configureWithTransparentBackground];
+            appearance.backgroundEffect = nil;
+            appearance.shadowColor = nil;
+            appearance.backgroundColor = color;
+            appearance.backgroundImage = [UIImage imageWithSize:CGSizeMake(1, 1) color:color];
```
donatoaguirre24, line 99: After upgrading from version `8.7.3` to `8.8.7`, mounted iOS screen headers stopped reacting correctly to system light/dark appearance changes.

It's worth mentioning that **our app currently opts out of Liquid Glass** by setting `UIDesignRequiresCompatibility` to `true` in the `Info.plist`.

Expected behavior:
- Native top bars using dynamic `UIColor` values should update automatically when the system appearance changes.
- Already mounted screens should resolve the correct light/dark top bar color immediately after an appearance change.

Actual behavior:
- New screens show the correct top bar colors.
- Already mounted screens show the stale top bar color after an appearance change.
  - This affects the currently focused stack as well as inactive mounted bottom tab stacks.
  - The top bar color corrects itself on screen re-focus, for example by navigating to another tab and then returning to the original.

This appears related to the iOS top bar appearance changes from this PR.

In `TopBarAppearancePresenter.mm`, the iOS 26 path converts the configured `UIColor` into a 1x1 `UIImage`:

```objc
appearance.backgroundColor = color;
appearance.backgroundImage = [UIImage imageWithSize:CGSizeMake(1, 1) color:color];
```

If I'm not mistaken, this bakes the currently resolved light/dark value into a static image. Once the image is assigned to the navigation bar appearance, UIKit can no longer re-resolve the original dynamic UIColor on trait changes.

Patching the library to avoid generating the `backgroundImage` fixed the issue for us:

```objc
appearance.backgroundColor = color;
appearance.backgroundImage = nil;
```

I imagine a proper fix would be checking whether Liquid Glass is enabled to determine whether a `backgroundImage` should be generated, but wanted to check before opening an issue.

cc @markdevocht

PR #8317, opened 2026-05-27
fix(ios): separate top bar buttons on iOS 26 with zero-width spacers

### LINKED ISSUE #8203: IOS 26
### What happened?

I was not able to configure the bottonTab at IOS 26, not backgroundColor, not drawbehind, nothing works : (

I was able to control the style and behavior also at iOS 26

- [x] I have tested this issue on the latest react-native-navigation release and it still reproduces.

_No response_

React Native Navigation version: 8.7.0
React Native version: 0.83.1
Has Fabric (React Native's new rendering system) enabled: (yes/no)  YES
Node version: 
Device model: 
iOS version: 26

IsaacIsrael:
## Problem

On iOS 26 (Liquid Glass), when multiple `UIBarButtonItem`s are placed in the navigation bar (left or right side), they merge into a single visual platter. Users cannot distinguish individual buttons.

## Solution

After building the `barButtonItems` array in `RNNButtonsPresenter`, insert zero-width fixed-space items (`[UIBarButtonItem fixedSpaceItemOfWidth:0]`) between each bar button item. This causes the system to render each button as a separate element with its own platter.

The fix is gated behind `@available(iOS 26.0, *)` so earlier iOS versions are completely unaffected.

## Changes

- `ios/RNNButtonsPresenter.mm`: Insert spacers between bar button items on iOS 26+

Fixes #8203


IsaacIsrael:
hey @yedidyak  !
Can we merge this one?

PR #8319, opened 2026-06-01
[Fix] handle searchBar focus on mergeOptions for existing searchController

### LINKED ISSUE #8318: [Fix] searchBar `focus:` true ignored on mergeOptions when searchController already exists
## Bug

When calling `Navigation.mergeOptions()` with `topBar.searchBar.focus: true` on a screen that already has a search controller, the focus is silently ignored. The search bar only auto-focuses on **initial creation** (first `applyOptions`), never on subsequent `mergeOptions` calls.

### Current Behaviour

`setSearchBarWithOptions:` in `UIViewController+RNNOptions.mm` only handles `focus` inside the `if (!self.navigationItem.searchController)` branch (initial creation). The `else` branch (existing search controller) skips focus entirely.

### Expected Behaviour

`mergeOptions` with `focus: true` should refocus the search bar even when the search controller already exists. This is needed for re-focusing the search bar when the screen reappears (e.g., switching back to a bottom tab).

### Steps to Reproduce

1. Register a screen with `searchBar: { visible: true, focus: true }` in static options
2. Screen mounts — search bar is focused (correct)
3. Navigate away from the screen (e.g., switch tabs)
4. Navigate back and call `Navigation.mergeOptions(componentId, { topBar: { searchBar: { visible: true, focus: true } } })`
5. Search bar is **not** focused

### Proposed Fix

Add focus handling in the `else` branch of `setSearchBarWithOptions:`:

```objc
} else {
    // ... existing iOS 26 placement code ...

    if (focus) {
        dispatch_async(dispatch_get_main_queue(), ^{
          self.navigationItem.searchController.active = true;
          [self.navigationItem.searchController.searchBar becomeFirstResponder];
        });
    }
}
```

### Environment

- react-native-navigation: 8.8.2
- React Native: 0.85.3
- iOS 18+

IsaacIsrael:
## Summary

- When `mergeOptions` is called with `topBar.searchBar.focus: true` on a screen whose `searchController` already exists, the focus was silently ignored — the `becomeFirstResponder` logic only ran in the initial-creation branch
- Adds focus handling in the `else` branch so the search bar refocuses on existing search controllers (e.g., when switching back to a bottom tab)

Fixes #8318

## Test plan

1. Register a bottom-tab screen with `searchBar: { visible: true, focus: true }`
2. Mount the screen — verify search bar auto-focuses
3. Switch to a different tab, then switch back
4. Call `Navigation.mergeOptions(componentId, { topBar: { searchBar: { visible: true, focus: true } } })` on `componentDidAppear`
5. Verify the search bar re-focuses on return


yedidyak (approved):
Looks good. I added a focused iOS regression test that verifies mergeOptions refocuses an existing search controller.

IsaacIsrael:
Hey @yedidyak ,  can we merge it?

IsaacIsrael:
Hey @yedidyak, can we merge this one?

Issue #6456, opened 2020-08-05
Custom topbar title does not take the whole space and looks shifted (iOS + Android)


samcampisi:
### Issue Description

I've seen some issues related to the ```alignment``` value for the title in the topbar options, but I'm having trouble with this.

I'm trying to render a custom title component in a topbar with "alignment fill", but the custom component does not take the whole space.

Here's how it looks on both platforms: 

iOS:
<img width="400" alt="Screenshot 2020-08-05 at 13 53 28" src="https://user-images.githubusercontent.com/42766924/89409869-0b33e900-d723-11ea-8d03-c07865bb9aa9.png">

Android:
<img width="554" alt="Screenshot 2020-08-05 at 13 55 12" src="https://user-images.githubusercontent.com/42766924/89410027-49c9a380-d723-11ea-8039-85d64a904fb7.png">

### Steps to Reproduce / Code Snippets / Screenshots

I looked to see if I had some weird style but after reducing it to the minimum I still observe this.

Here are the topbar options:
```
topBar: {
        visible: true,
        title: {
          component: {
            name: 'app.CustomTopBar',
            aligment: 'fill',
          },
          aligment: 'fill',
        },
        backButton: {
          visible: false,
        },
},
```

And the view in the custom component (aka app.CustomTopBar)
```
<View style={{
    flex: 1,
    margin: 0,
    padding: 0,
    height: 80,
    flexDirection: 'row',
    justifyContent: 'space-between',
    backgroundColor: 'pink'
}}>
        <Text>A</Text>
        <Text>B</Text>
        <Text>C</Text>
</View>
```

Where does this left shifting come from? Is there anything I could do to make this occupy the whole space? 
Even after giving it width: '100%' or width the size of the screen this space is still there

---
### Environment
* React Native Navigation version: 6.11
* React Native version: 0.62.2
* Platform(s) (iOS, Android, or both?): Both
* Device info (Simulator/Device? OS version? Debug/Release?): iOS 13.6 in an iPhoneSE Simulator, Android 9 Xiaomi Mi A1

jinshin1013:
Hi @samcampisi, thanks for reporting the issue! It'd be great if you could fork the project and push a reproduction in the Playground app so that we could have more close look at the issue. Cheers!

guyca:
Hey @samcampisi 
Seems like there's a typo in the options declaration. It should be `alignment`.
Also, if you want to set a background component which will stretch across the entire TopBar, then use:
```js
[...]

PR #8325, opened 2026-06-06
android: fix custom top bar title Fill alignment

phuongwd:
## Summary

Closes a regression introduced in [PR #8182](https://github.com/wix/react-native-navigation/pull/8182): Fill-aligned custom title components on Android still render with a margin gap on both edges instead of spanning the entire space between the left and right button bars. Reproducible on upstream `master`.

Three layered fixes plus tests, in three reviewable commits:

### 1. `makeTitleAtMostWidthMeasureSpec` - drop the 32dp reservation for Fill

#8182 added the `isFill` parameter and switched the MeasureSpec mode from `AT_MOST` to `EXACTLY` for Fill, but the else branch still subtracts `2 * DEFAULT_LEFT_MARGIN_PX` (32dp). That reservation is the Material-style breathing room for left-aligned titles - wrong for Fill, whose intent is edge-to-edge.

### 2. `resolveHorizontalTitleBoundsLimit` + `TitleAndButtonsContainer.onLayout` - position Fill edge-to-edge

Even with #1, Fill titles were laid out via `resolveHorizontalTitleBoundsLimit`, which had no Fill awareness and applied the same 16dp reservation as left-aligned titles. Result: the React component is measured at the full available width but laid out in a slot 16dp narrower on each side. This commit adds `isFill: Boolean = false` to the helper (default keeps existing callers unchanged) and threads the flag from `onLayout`.

### 3. Tests - Robolectric coverage for the matrix that turned the regression on

`makeTitleAtMostWidthMeasureSpec` had zero direct unit tests, and the existing Fill assertion in `TitleAndButtonsContainerTest` codified the buggy left-aligned-with-margins position. This commit:

- Adds Robolectric tests for `makeTitleAtMostWidthMeasureSpec` covering Fill / Center / left-aligned with and without button bars, asserting on both `MeasureSpec.getMode()` and `MeasureSpec.getSize()`.
- Adds Robolectric tests for `resolveHorizontalTitleBoundsLimit(isFill = true)` covering no-buttons / with-buttons / RTL.
- Corrects the buggy Fill assertion in `TitleAndButtonsContainerTest.onLayout - should measure and layout children when alignment changes` to assert the actual Fill intent (`left = 0`, `right = UUT_WIDTH`).

## Related

- Closes the Fill manifestation reported in [#6456](https://github.com/wix/react-native-navigation/issues/6456) (Custom topbar title does not take the whole space).
- Follow-up to [#8182](https://github.com/wix/react-native-navigation/pull/8182), which added the `isFill` plumbing without the margin handling and without a Robolectric assertion on the measured width.

## Validation

Verified end-to-end on a physical Samsung Galaxy S21 (Android 15) by capturing matched before/after pairs against the same playground build. Repro builds: master `1afec76bae` vs this PR `462c6f53ca`. A custom `TopBarTitleTestScreen` mounts a 32×32 red marker as `leftButton.component` plus a long Fill-aligned title; the red square makes the leftover gap obvious.

**State 1 - Fill title with subtitle:**

| Before (master `1afec76bae`) | After (this PR) |
|---|---|
| state-1-with-subtitle | state-1-with-subtitle |

**State 2 - Fill title without subtitle (the original report from #6456):**

| Before (master `1afec76bae`) | After (this PR) |
|---|---|
| state-2-without-subtitle-BUG | state-2-without-subtitle-BUG |

In the "after" shots the title sits flush against the red marker, the ~32dp leftover gap is gone, and a few more characters of the long title fit before truncation - matching the Fill-aligned edge-to-edge intent.

New Robolectric tests added in this PR exercise the matrix that turned the regression on. Local `yarn test-unit-android` run pending.

## What is not yet here / open questions

- A matching playground assertion (Detox `toHaveRect`) on the Fill-aligned title's right edge - happy to add this if you would prefer it land in this PR.
- iOS parity. #8182 also touched `RNNReactTitleView.mm` (`_expectedHeight`). My device repro is Android-only; iOS Fill appears correct in adjacent testing but a maintainer eye on it would be welcome.

### DIFF
--- a/android/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/TitleAndButtonsContainer.kt
+++ b/android/src/main/java/com/reactnativenavigation/views/stack/topbar/titlebar/TitleAndButtonsContainer.kt
@@ -106,6 +106,7 @@ class TitleAndButtonsContainer(context: Context) : ViewGroup(context) {
     override fun onLayout(changed: Boolean, l: Int, t: Int, r: Int, b: Int) {
         val titleComponent = getTitleComponent()
         val isCenter = titleComponentAlignment == Alignment.Center
+        val isFill = titleComponentAlignment == Alignment.Fill
         val containerWidth = r - l
         val containerHeight = b - t
         val isRTL = isRTL()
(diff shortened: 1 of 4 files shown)


yedidyak:
Hi, it seems to shift some titles higher than they are previously. I'm not sure if that is a better implementation, but it does change the behaviour a bit.

